PR #6, opened 2016-06-24
StorageHelper in base utils.

jaeklim:
StorageHelper to access Android local storage such as shared preferences and internal storage.


mh8395648:
Thank you very much for activating my iPhone 6 Plus

PR #239, opened 2016-11-16
Fix logs printed multiple times

jaelim-ms:


sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/MobileCenter.java
```diff
@@ -354,15 +349,14 @@ private synchronized void startService(@NonNull MobileCenterService service) {
         service.onChannelReady(mApplication, mChannel);
         if (isInstanceEnabled())
             mApplication.registerActivityLifecycleCallbacks(service);
+        MobileCenterLog.debug(LOG_TAG, service.getClass().getSimpleName() + " service started.");
```
guperrot: Should be info log

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/MobileCenter.java
```diff
@@ -281,42 +281,42 @@ private synchronized boolean instanceConfigure(Application application, String a
...
+
+        MobileCenterLog.logAssert(LOG_TAG, "Mobile Center SDK configuration failed.");
+        return false;
     }
 
     @SafeVarargs
     private final synchronized void startServices(Class<? extends MobileCenterService>... services) {
         if (services == null) {
-            MobileCenterLog.logAssert(LOG_TAG, "Cannot start services, services array is null. Failed to start Mobile Center SDK");
+            MobileCenterLog.error(LOG_TAG, "Cannot start services, services array is null. Failed to start services.");
```
guperrot: should replace services by class names

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/MobileCenter.java
```diff
@@ -281,42 +281,42 @@ private synchronized boolean instanceConfigure(Application application, String a
...
             return;
         }
         if (mApplication == null) {
-            MobileCenterLog.logAssert(LOG_TAG, "Cannot start services, Mobile Center has not been configured. Failed to start Mobile Center SDK");
+            MobileCenterLog.error(LOG_TAG, "Cannot start services, Mobile Center has not been configured. Failed to start services.");
```
guperrot: should replace services by class names

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/MobileCenter.java
jaelim-ms: I agree but not this line. This is the case if `services == null`. I'll change the below log with class names.

guperrot: ah yes indeed

PR #246, opened 2016-11-21
Fix issues wrapper sdk exception manager

guperrot:
* Fix a file leak when data could not be read.
* Fix noisy logging when wrapper sdk not used.
* Fix unit tests.


sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/WrapperSdkExceptionManager.java
```diff
@@ -50,36 +51,35 @@ public static void saveWrapperExceptionData(byte[] data, UUID errorId) {
...
         if (errorId == null) {
             MobileCenterLog.error(Crashes.LOG_TAG, "Failed to delete wrapper exception data: null errorId");
             return;
         }
-
         File dataFile = getFile(errorId);
-        byte[] loadResult = loadWrapperExceptionData(errorId);
-        if (loadResult == null) {
-            MobileCenterLog.error(Crashes.LOG_TAG, "Failed to delete wrapper exception data: data not found");
-            return;
```
guperrot: the return statement was the leak @achocron

PR #248, opened 2016-11-22
Test delete logs from DatabaseAsync as not yet covered

guperrot:


guperrot:
More tests are coming in that class but issue is now fixed.

PR #272, opened 2016-12-07
Feature/call callback upon request failures

jaelim-ms:


sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
```diff
@@ -375,19 +389,21 @@ private synchronized void handleSendingFailure(@NonNull final GroupState groupSt
...
-
         /* Check group name is registered. */
         final GroupState groupState = mGroupStates.get(groupName);
         if (groupState == null) {
             MobileCenterLog.error(LOG_TAG, "Invalid group name:" + groupName);
             return;
         }
 
+        /* Check if disabled with discarding logs. */
+        if (mDiscardLogs) {
```
guperrot: Hum, that could break people expecting onBeforeSending being called before one of sent/failure. Should we call onBeforeSending then?

guperrot: Otherwise happy with the change.

guperrot: I think we should call `onBeforeSending` in that case to simplify callback users code.

jaelim-ms: That will make the callback flow consistent but the SDK already knows that it will fail anyway. Looks no better idea for this case, I'll just call it before `onSendingFailed`.

guperrot:
This does not solve the scenario where crash would be enqueued but not yet batched and analytics is failing in fatal error between the two events.

guperrot:
Will address the comment above in a new PR as this one already fixes 2 bugs without regression.

PR #276, opened 2016-12-09
More crash buttons

guperrot:


jaelim-ms:
Since you are working on Sasquatch, can we have separators in MainActivity by services with titles? It would be good to group list items by service something similar with Settings.

PR #278, opened 2016-12-09
Feature/add analytics callbacks

achocron:


sdk/mobile-center-analytics/src/androidTest/java/com/microsoft/azure/mobile/analytics/AnalyticsAndroidTest.java
```diff
@@ -0,0 +1,97 @@
...
+    public void testAnalyticsListener() {
+
+        AnalyticsListener analyticsListener = mock(AnalyticsListener.class);
+        Analytics.setListener(analyticsListener);
+        Channel channel = mock(Channel.class);
+        Analytics.getInstance().onChannelReady(sContext, channel);
+        Analytics.trackEvent("event");
+
+        /* First process: enqueue log but network is down... */
+        android.util.Log.i(TAG, "Process 1");
```
jaelim-ms: Just curious. Do you want to see this log in console while the test is running? If it is meaningless, just delete it.

sdk/mobile-center-analytics/src/main/java/com/microsoft/azure/mobile/analytics/Analytics.java
```diff
@@ -265,6 +281,43 @@ public synchronized void setInstanceEnabled(boolean enabled) {
...
     }
 
+    @Override
+    protected Channel.GroupListener getChannelListener() {
+        return new Channel.GroupListener() {
+            private static final int BEFORE_SENDING = 0;
+            private static final int SENDING_SUCCEEDED = 1;
+            private static final int SENDING_FAILED = 2;
+
+            private void callback(int type, Log log, Exception e) {
```
jaelim-ms: You don't need to follow this logic in Crashes. Simply call callbacks from each method and remove `callback` method.
The reason of doing this in Crashes is that there are duplicated code blocks in the listener and I wanted to eliminate that.

sdk/mobile-center-analytics/src/main/java/com/microsoft/azure/mobile/analytics/channel/AnalyticsListener.java
```diff
@@ -0,0 +1,29 @@
...
+import com.microsoft.azure.mobile.analytics.ingestion.models.EventLog;
+
+public interface AnalyticsListener {
+
+    /**
+     * Called right before sending an event log. The callback can be invoked multiple times based on the number of event logs.
+     *
+     * @param eventLog The event log that will be sent.
+     */
+    void onBeforeSending(EventLog eventLog);
```
jaelim-ms: Why is this limited to EventLog?

sdk/mobile-center-analytics/src/androidTest/java/com/microsoft/azure/mobile/analytics/AnalyticsAndroidTest.java
guperrot: It's copy paste from the crash test, should be removed indeed.

sdk/mobile-center-analytics/src/main/java/com/microsoft/azure/mobile/analytics/Analytics.java
```diff
@@ -121,6 +128,15 @@ public static void setEnabled(boolean enabled) {
     }
 
     /**
+     * Sets an analytics listener.
+     *
+     * @param listener The custom crashes listener.
+     */
+    public static void setListener(AnalyticsListener listener) {
+        getInstance().mAnalyticsListener = listener;
```
guperrot: Can you try changing to protected and use android bindings config to change it back to public?

sdk/mobile-center-analytics/src/main/java/com/microsoft/azure/mobile/analytics/channel/AnalyticsListener.java
guperrot: Could indeed use the generic type.

sdk/mobile-center-analytics/src/main/java/com/microsoft/azure/mobile/analytics/Analytics.java
```diff
@@ -132,7 +132,8 @@ public static void setEnabled(boolean enabled) {
      *
      * @param listener The custom crashes listener.
      */
-    public static void setListener(AnalyticsListener listener) {
+    @VisibleForTesting
+    protected static void setListener(AnalyticsListener listener) {
```
jaelim-ms: Didn't lint complain about this? default?

sdk/mobile-center-analytics/src/main/java/com/microsoft/azure/mobile/analytics/channel/AnalyticsListener.java
achocron: In iOS we use the specific type. Since it's for internal use I suppose that difference won't matter, though, correct?

achocron: So if we are going to just use the generic log type, there is probably no point in using AnalyticsListener at all

sdk/mobile-center-analytics/src/main/java/com/microsoft/azure/mobile/analytics/Analytics.java
```diff
@@ -121,6 +128,17 @@ public static void setEnabled(boolean enabled) {
...
 
     /**
+     * Sets an analytics listener.
+     *
+     * @param listener The custom crashes listener.
+     */
+    @SuppressWarnings("WeakerAccess")
+    @VisibleForTesting
+    protected static void setListener(AnalyticsListener listener) {
+        getInstance().mAnalyticsListener = listener;
```
guperrot: Please create setInstanceListener private method and move code there.

jaelim-ms:
Please run re-format before merge.

PR #282, opened 2016-12-12
Refactor synchronization logic to fix local coverage report

guperrot:
Jacoco does not like synchronized (object)


sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/utils/UUIDUtils.java
```diff
@@ -33,6 +33,7 @@ public UUID randomUUID() {
     UUIDUtils() {
     }
 
+
```
jaelim-ms: intended empty line?

guperrot: fixed

PR #283, opened 2016-12-13
Wait logs to be flushed on disk on crash

guperrot:
Posting the PR early before tests so that we can discuss the design.

### DIFF
--- a/sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
+++ b/sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
@@ -258,7 +258,7 @@ public class DefaultChannel implements Channel {
         try {
             mIngestion.close();
         } catch (IOException e) {
-            MobileCenterLog.error(LOG_TAG, "Failed to shutdown ingestion", e);
+            MobileCenterLog.error(LOG_TAG, "Failed to close ingestion", e);
         }
         for (GroupState groupState : mGroupStates.values()) {
             handleFailureCallback(groupState, deleteLogs);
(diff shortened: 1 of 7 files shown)


guperrot:
For some reason codecov did not update the comment but you can check directly on the green marks that it's now back to 100%

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
```diff
@@ -48,6 +48,12 @@
     static final int CLEAR_BATCH_SIZE = 100;
 
     /**
+     * Shutdown timeout in millis.
+     */
+    @VisibleForTesting
+    static final int SHUTDOWN_TIMEOUT = 5000;
```
jaelim-ms: Any reason of this number? Just curious, I also considered 5 sec for waiting.

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/persistence/DatabasePersistenceAsync.java
```diff
@@ -203,6 +209,27 @@ public void run() {
...
+        final Semaphore semaphore = new Semaphore(0);
+        mHandler.post(new Runnable() {
+
+            @Override
+            public void run() {
+                semaphore.release();
+                MobileCenterLog.debug(LOG_TAG, "Persistence tasks completed.");
+            }
+        });
+        if (!semaphore.tryAcquire(timeout, TimeUnit.MILLISECONDS)) {
```
jaelim-ms: It is non-blocking acquire, can you explain of tryAcquire instead of acquire?

PR #286, opened 2016-12-14
Run file I/O in background for Crashes service

jaelim-ms:


sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -142,6 +166,10 @@ private Crashes() {
         mCrashesListener = DEFAULT_ERROR_REPORTING_LISTENER;
         mUnprocessedErrorReports = new LinkedHashMap<>();
         mErrorReportCache = new LinkedHashMap<>();
+        HandlerThread thread = new HandlerThread(THREAD_NAME);
```
guperrot: Why using handler thread for one shot tasks?
This case is legit for ASyncTask as after we load what we need, there won't be any other use of the thread until application dies.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -88,6 +97,18 @@
...
+     * Handler for background thread loop.
+     */
+    @VisibleForTesting
+    final Handler mHandler;
+
+    /**
+     * Handler for main thread loop.
+     */
+    @VisibleForTesting
+    final Handler mMainHandler;
```
guperrot: Code analysis warning here.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -526,43 +585,50 @@ public void setWrapperSdkListener(WrapperSdkListener wrapperSdkListener) {
...
     @VisibleForTesting
-    private synchronized void handleUserConfirmation(@UserConfirmationDef int userConfirmation) {
+    private synchronized void handleUserConfirmation(@UserConfirmationDef final int userConfirmation) {
         if (mChannel == null) {
             MobileCenterLog.error(LOG_TAG, "Crashes service not initialized, discarding calls.");
             return;
         }
 
-        if (userConfirmation == DONT_SEND) {
+        mHandler.post(new Runnable() {
```
guperrot: Check if already in UI thread.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: It was exposed for unit test and now don't need it. Also removing the annotation.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: We are free from thread pool limitation with Handler. It is an SDK and trying not to consume resources as many as possible. Going with AsyncTask is not a problem since it is one shot tasks as you mentioned.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: Not a big deal but it is just using same handler if it is already on background thread.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -287,24 +287,19 @@ private synchronized boolean hasInstanceCrashedInLastSession() {
...
-                }
-                mLastCrashErrorReportListeners.add(listener);
-                break;
-            case PROCESSED:
-                listener.onSuccess(mLastSessionErrorReport);
-                break;
-            case PROCESSING_FAILED:
-                listener.onFailure();
-                break;
-            case NOT_FOUND:
```
jaelim-ms: That didn't help for 100% coverage. You can see some explanations in stackoverflow. Looks byte code has more cases.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -102,6 +103,11 @@
     private static final CrashesListener DEFAULT_ERROR_REPORTING_LISTENER = new DefaultCrashesListener();
 
     /**
+     * Timeout for getting crash for the last session.
+     */
+    private static final int GET_LAST_SESSION_CRASH_REPORT_TIMEOUT = 10000;
```
guperrot: We should wait forever for last session error report.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -267,12 +283,23 @@ public static boolean hasCrashedInLastSession() {
 
     /**
      * Provides information about any available crash report from the last session, if it crashed.
+     * This method is a synchronous call and blocks UI thread.
```
guperrot: Javadoc: not UI thread but calling thread. If you call it in background it won't block UI thread.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -267,12 +283,23 @@ public static boolean hasCrashedInLastSession() {
...
+     * Use {@link #getLastSessionCrashReportAsync(LastCrashErrorReportListener)} for asynchronous call.
+     *
+     * @return The crash report from the last session if one was set.
+     */
+    @Nullable
+    public static ErrorReport getLastSessionCrashReport() {
```
guperrot: Please add @WorkerThread

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -182,6 +197,7 @@ private Crashes() {
         thread.start();
         mHandler = new Handler(thread.getLooper());
         mLastSessionCrashProcessingStatus = PREPARE_PROCESSING;
+        mSemaphore = new Semaphore(0);
```
guperrot: This is wrong to use a shared semaphore if you have more than 1 caller. This is not the right synchronization mechanism to use in that case. Use a Condition or a CountDownLatch (I prefer CountDownLatch as it does not have finally block, finally is troublesome with code coverage).

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -155,7 +157,9 @@
     /**
      * Flag indicates whether Crashes found ErrorReport for the last session.
      */
-    private boolean mFoundLastSessionErrorReport;
+    private LastCrashProcessingStatus mFoundLastSessionErrorReport;
+
+    private List<LastCrashErrorReportListener> mLastCrashErrorReportListeners;
```
guperrot: Set

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -253,28 +258,44 @@ public static boolean hasCrashedInLastSession() {
...
+                if (mLastCrashErrorReportListeners == null) {
+                    mLastCrashErrorReportListeners = new ArrayList<>();
+                }
+                mLastCrashErrorReportListeners.add(listener);
+                break;
+            case PROCESSED:
+                listener.onSuccess(mLastSessionErrorReport);
+                break;
+            case PROCESSING_FAILED:
+                listener.onFailure();
```
guperrot: We should not have a failure case for getSessionErrorReport, either return result or return null like before. The sync version has to return either a result or null and never throw an exception, so why would the async version introduce that?

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -694,6 +745,37 @@ void saveErrorLog(ManagedErrorLog errorLog, File errorStorageDirectory, String f
...
+     */
+    private enum LastCrashProcessingStatus {
+        PREPARE_PROCESSING,
+        PROCESSING,
+        PROCESSED,
+        PROCESSING_FAILED,
+        NOT_FOUND
+    }
+
+    public interface LastCrashErrorReportListener {
```
guperrot: This is related to another comment on simplifying cases but we should not have this interface. We will have more data to return asynchronously like that in the near future such as installId for example.

We should add a generics callback in the core module with only 1 method onResult(T data) and data can be null (if not found), and should never fail on an exception.

Azme has that generic callback if you want to look at it.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -238,12 +294,61 @@ public static ErrorReport getLastSessionCrashReport() {
...
+                mLastSessionCrashProcessingStatus == PROCESSING) {
+            if (mLastCrashErrorReportListeners == null) {
+                mLastCrashErrorReportListeners = new ArrayList<>();
+            }
+            mLastCrashErrorReportListeners.add(listener);
+            MobileCenterLog.info(LOG_TAG, "Crashes for the last session have not been processed yet. The SDK will call listener when it completes process.");
+        } else if (mLastSessionCrashProcessingStatus == PROCESSED)
+            listener.onSuccess(mLastSessionErrorReport);
+        else if (mLastSessionCrashProcessingStatus == PROCESSING_FAILED)
+            listener.onFailure();
```
guperrot: We should not have a failure case for getSessionErrorReport, either return result or return null like before. The sync version has to return either a result or null and never throw an exception, so why would the async version introduce that?

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -613,6 +800,26 @@ void saveErrorLog(ManagedErrorLog errorLog, File errorStorageDirectory, String f
         MobileCenterLog.debug(Crashes.LOG_TAG, "Saved JSON content for ingestion into " + errorLogFile);
     }
 
+    public interface LastCrashErrorReportListener {
```
guperrot: This is related to another comment on simplifying cases but we should not have this interface. We will have more data to return asynchronously like that in the near future such as installId for example.

We should add a generics callback in the core module with only 1 method onResult(T data) and data can be null (if not found), and should never fail on an exception.

Azme has that generic callback if you want to look at it.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
guperrot: Make sure last session error report is loaded first, then process pending crashes.
Then quit looper?

guperrot:
one of the review section is marked as outdated because I did the review from an individual commit, will redo the comments.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: Introduced generic callback in mobile-center and replaced by that.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: Introduced generic callback in mobile-center and replaced by that.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: Added.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: Agreed.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: Yes, you are right. It will block forever if there are more than 1 callers. Changed the implementation by using CountDownLatch.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: It sounds like a PM question for requirement.
-> Got the same opinion. Removing timeout and just wait forever.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: We need to talk about this. Taking offline.

guperrot: Talked offline. Will be done in a separate PR, switching to ASyncTask might be more practical to handle cancelling on disabled and then this comment would be obsolete.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -219,16 +251,18 @@ public static void notifyUserConfirmation(@UserConfirmationDef int userConfirmat
...
      * @return {@code true} if a crash was recorded in the last session, otherwise {@code false}.
      */
     public static boolean hasCrashedInLastSession() {
-        return getLastSessionCrashReport() != null;
+        return getInstance().hasInstanceCrashedInLastSession();
     }
 
     /**
      * Provides information about any available crash report from the last session, if it crashed.
+     * This method is a synchronous call and blocks caller thread.
```
guperrot: Please add @WorkerThread in addition to that explanation.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -219,16 +251,18 @@ public static void notifyUserConfirmation(@UserConfirmationDef int userConfirmat
...
+     * Use {@link #getLastSessionCrashReportAsync(ResultCallback)} for asynchronous call.
```
guperrot: Can use @See also.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -238,12 +272,51 @@ public static ErrorReport getLastSessionCrashReport() {
...
+    /**
      * Implements {@link #getLastSessionCrashReport()} at instance level.
      */
     private synchronized ErrorReport getInstanceLastSessionCrashReport() {
+        if (mCountDownLatch != null) {
+            MobileCenterLog.debug(LOG_TAG, "Waiting for Crashes service to complete crash report for the last session.");
+            try {
+                mCountDownLatch.await();
+            } catch (InterruptedException e) {
+                MobileCenterLog.debug(LOG_TAG, "Could not get crash report for the last session.");
```
guperrot: `, e);` to log the reason why.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -238,12 +272,51 @@ public static ErrorReport getLastSessionCrashReport() {
...
     }
 
+    /**
+     * Implements {@link #getLastSessionCrashReportAsync(ResultCallback)} at instance level.
+     */
+    private synchronized void getInstanceLastSessionCrashReportAsync(ResultCallback<ErrorReport> callback) {
+        if (mCountDownLatch == null || mCountDownLatch.getCount() <= 0)
+            callback.onResult(mLastSessionErrorReport);
+        else {
+            if (mLastCrashErrorReportCallbacks == null) {
```
guperrot: Should simplify, in other places we don't lazy initialize listeners.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -238,12 +272,51 @@ public static ErrorReport getLastSessionCrashReport() {
...
+                mLastCrashErrorReportCallbacks = new ArrayList<>();
+            }
+            mLastCrashErrorReportCallbacks.add(callback);
+            MobileCenterLog.info(LOG_TAG, "Crashes for the last session have not been processed yet. The SDK will call listener when it completes process.");
```
guperrot: processing. (or finishes to process or equivalent)

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -408,53 +480,105 @@ private void initialize() {
...
+                    @Override
+                    public void run() {
+                        String logFileContents = StorageHelper.InternalStorage.read(logFile);
+                        if (logFileContents == null)
+                            MobileCenterLog.error(LOG_TAG, "Error reading last session error log.");
+                        else {
+                            try {
+                                ManagedErrorLog log = (ManagedErrorLog) mLogSerializer.deserializeLog(logFileContents);
+                                mLastSessionErrorReport = buildErrorReport(log);
+                                MobileCenterLog.debug(LOG_TAG, "Completed to process crash report for the last session.");
```
guperrot: "Completed to process" => I would suggest just "Processed"

guperrot:
Also would suggest using both versions of the get in Sasquatch, for example, use async, and inside the async callback call the sync getter to log so that both functions are used there.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: Done

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: Done

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: Done

jaelim-ms:
Your last feedback will be done in the next PR.

jaelim-ms:
I meant the example will be in the next PR. As we introduce a new API, we should show how to use in our test application. This will also fix lint issue. Anyway, it will come soon. 😄

guperrot:
Ah ok, misread the comment "next pr" ;)

PR #292, opened 2016-12-15
Fix linter suggestions

guperrot:


jaelim-ms:
Looks good to me but it would be good to take projectDependency stuffs out of main. Default should be jcenterDependency which has less features than the other.

guperrot:
This branch does not have any difference between the 2 deps in terms of features. You are probably referring to your pr @jaelim-ms

PR #295, opened 2017-01-05
Hide part of app secret in the log

jaelim-ms:


sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/ingestion/http/IngestionHttp.java
```diff
@@ -82,6 +82,12 @@
     private static final int READ_TIMEOUT = 20000;
 
     /**
+     * Regex pattern for application secret in header.
```
guperrot: Overkill to use regex like this but ok...

However appSecret can be something else than UUID soon (for codepush migration or something like that), that's why we removed the client side check earlier.

Your code will not obfuscate an appSecret that does not match UUID format. Why not using a simple substring based code that does not assume anything about format (except handling short length edge case)?

jaelim-ms: I understand you you are saying and I also concerned about the format that will eventually get changed. The reason of using regex was to have simple lines of code instead of several and just implemented based on current format of app secret (having dashes looks better than all asterisks). 😄
Instead of revisiting here when the format changes, I'll simply check the length and hide the first 3/4 of the string. This way is more flexible if the app secret length is shorter. If you have different opinion and just want to hide/show fixed length, let me know. That way also works.

guperrot: If it's too short, obfuscate everything, I believe that's how we did it for azme gcm on portal.

guperrot:
Related but not blocking: sasquatch set app secret feature in menu should not validate format.

jaelim-ms:
Created a task for the validation issue.

sdk/mobile-center/src/test/java/com/microsoft/azure/mobile/ingestion/http/IngestionHttpTest.java
```diff
@@ -103,8 +103,8 @@ public void success() throws Exception {
         IngestionHttp httpClient = new IngestionHttp(serializer);
         httpClient.setServerUrl("http://mock");
 
-        /* Test calling code. */
-        String appSecret = UUIDUtils.randomUUID().toString();
+        /* Test calling code. Use shorter but valid app secret. */
+        String appSecret = "SHORT";
```
guperrot: Constant ;)

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/ingestion/http/IngestionHttp.java
```diff
@@ -82,6 +83,11 @@
     private static final int READ_TIMEOUT = 20000;
 
     /**
+     * Maximum characters to be displayed in a log for application secret.
+     */
+    private static final int MAX_CHARACTERS_DISPLAY_FOR_APP_SECRET = 8;
```
guperrot: DISPLAYED?

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/ingestion/http/IngestionHttp.java
```diff
@@ -126,7 +132,15 @@ private static void doCall(String baseUrl, LogSerializer logSerializer, String a
             urlConnection.setRequestProperty(CONTENT_TYPE_KEY, CONTENT_TYPE_VALUE);
             urlConnection.setRequestProperty(APP_SECRET, appSecret);
             urlConnection.setRequestProperty(INSTALL_ID, installId.toString());
-            MobileCenterLog.verbose(LOG_TAG, "Headers: " + urlConnection.getRequestProperties());
+
+            /* Log headers. */
+            int hidingEndIndex = appSecret.length() - (appSecret.length() >= MAX_CHARACTERS_DISPLAY_FOR_APP_SECRET ? MAX_CHARACTERS_DISPLAY_FOR_APP_SECRET : 0);
```
guperrot: maybe surround this by a if getLogLevel <= Verbose?

sdk/mobile-center/src/test/java/com/microsoft/azure/mobile/ingestion/http/IngestionHttpTest.java
jaelim-ms: Just a value in test. Don't think this will be reused by anywhere.

guperrot: Yeah just realized it was a test class.

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/ingestion/http/IngestionHttp.java
jaelim-ms: Good point!

PR #297, opened 2017-01-11
Ensure crash callbacks are executed on main thread

guperrot:


sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/CrashesListener.java
```diff
@@ -1,5 +1,6 @@
 package com.microsoft.azure.mobile.crashes;
 
+import android.support.annotation.MainThread;
```
jaelim-ms: Please remove.

PR #299, opened 2017-01-11
Reimplement file I/O in worker thread with AsyncTask

jaelim-ms:


sdk/mobile-center-crashes/src/androidTest/java/com/microsoft/azure/mobile/crashes/CrashesAndroidTest.java
```diff
@@ -76,11 +77,25 @@ public void cleanup() {
...
         final Semaphore semaphore = new Semaphore(0);
 
-        /* Waiting background thread for initialize and processPendingErrors. */
-        Crashes.getInstance().mHandler.post(new Runnable() {
+        /* Waiting background thread for initialize. */
+        if (Crashes.getInstance().mCountDownLatchForLastSessionErrorReport != null)
+            Crashes.getInstance().mCountDownLatchForLastSessionErrorReport.await();
+
+        /* Waiting background thread for processPendingErrors. Wait for 2 second to make sure Crashes service starts processing pending errors. */
+        HandlerThread thread = new HandlerThread("TestThread");
```
guperrot: Why do you need a HanlderThread in unit test then? HandlerThread is for long lived looper usually tied to application lifetime, not for a single task.

sdk/mobile-center-crashes/src/androidTest/java/com/microsoft/azure/mobile/crashes/CrashesAndroidTest.java
```diff
@@ -76,11 +77,25 @@ public void cleanup() {
...
+        thread.start();
+        Handler handler = new Handler(thread.getLooper());
+        handler.post(new Runnable() {
 
             @Override
             public void run() {
+                try {
+                    TimeUnit.SECONDS.sleep(2);
+                    if (Crashes.getInstance().mCountDownLatchForProcessPendingErrors != null)
```
guperrot: a package private getter would be nicer, we did that so far.

sdk/mobile-center-crashes/src/androidTest/java/com/microsoft/azure/mobile/crashes/CrashesAndroidTest.java
jaelim-ms: The test should wait until it starts processing error report and I don't want the main thread to be blocked. I can move this to static member unless it is not a problem if the main thread is blocked. I can give a try for this one.

sdk/mobile-center-crashes/src/androidTest/java/com/microsoft/azure/mobile/crashes/CrashesAndroidTest.java
jaelim-ms: Actually the method is defined only for testing but I can use getters for count down latches.

sdk/mobile-center-crashes/src/androidTest/java/com/microsoft/azure/mobile/crashes/CrashesAndroidTest.java
guperrot: What I mean, is why HandlerThread? Seems like it's a silver bullet now, you could have used a single thread for example.

sdk/mobile-center-crashes/src/androidTest/java/com/microsoft/azure/mobile/crashes/CrashesAndroidTest.java
guperrot: What I mean is that for every member variable that we exposed so far in unit test it was through getters.

sdk/mobile-center-crashes/src/androidTest/java/com/microsoft/azure/mobile/crashes/CrashesAndroidTest.java
jaelim-ms: Oh! 😄

jaelim-ms:
Closing this PR since we have a customer issue that is caused by AsyncTask. Keep using Handler and implement cancel when it disables.

PR #301, opened 2017-01-12
Fix a crash while counting logs and disabling a module

guperrot:


sdk/mobile-center/src/test/java/com/microsoft/azure/mobile/channel/DefaultChannelTest.java
```diff
@@ -1077,4 +1081,114 @@ public boolean matches(Object argument) {
...
+                new Thread() {
+
+                    @Override
+                    public void run() {
+
+                        /* Wait for add/remove calls to be done. */
+                        try {
+                            latch.await();
+                        } catch (InterruptedException e) {
+                            throw new RuntimeException(e);
```
jaelim-ms: What if it throws a runtime exception, does the test fail?
It is running on another thread so I don't think it will fail.

guperrot: It's just that I prefer the unit test to crash so we can investigate than having a silent exception and trying to understand why the test breaks (in the future).

PR #302, opened 2017-01-13
Feature/cancel background task when disable

jaelim-ms:


sdk/mobile-center-crashes/src/androidTest/java/com/microsoft/azure/mobile/crashes/CrashesAndroidTest.java
```diff
@@ -119,21 +93,11 @@ public void run() {
...
-                    throw new AssertionFailedError();
-                }
-            }
-        });
         Crashes.getInstance().onChannelReady(sContext, channel);
         assertNotNull(Crashes.getLastSessionCrashReport());
 
         /* Try to get last session crash after Crashes service completed processing. */
-        waitForCrashesHandlerTasksToComplete();
+//        waitForCrashesHandlerTasksToComplete();
```
guperrot: remove then

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -535,12 +551,18 @@ public void run() {
...
-                        for (Iterator<ResultCallback<ErrorReport>> iterator = mLastCrashErrorReportCallbacks.iterator(); iterator.hasNext(); ) {
-                            ResultCallback<ErrorReport> callback = iterator.next();
-                            iterator.remove();
-                            callback.onResult(mLastSessionErrorReport);
-                        }
+                        HandlerUtils.runOnUiThread(new Runnable() {
+
+                            @Override
+                            public void run() {
+                                /* Call callbacks for getInstanceLastSessionCrashReport(ResultCallback) . */
```
guperrot: new line should go before comment

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -698,31 +721,35 @@ public void run() {
...
+                        Map.Entry<UUID, ErrorLogReport> unprocessedEntry = unprocessedIterator.next();
+                        ErrorLogReport errorLogReport = unprocessedEntry.getValue();
+                        ErrorAttachment attachment = mCrashesListener.getErrorAttachment(errorLogReport.report);
+                        if (attachment == null)
+                            MobileCenterLog.debug(LOG_TAG, "CrashesListener.getErrorAttachment returned null, no additional information will be attached to log: " + errorLogReport.log.getId().toString());
+                        else
+                            errorLogReport.log.setErrorAttachment(attachment);
+                        mChannel.enqueue(errorLogReport.log, ERROR_GROUP);
+
+						/* Clean up an error log file and map entry. */
```
guperrot: tab instead of whitespaces apparently

sdk/mobile-center-crashes/src/androidTest/java/com/microsoft/azure/mobile/crashes/CrashesAndroidTest.java
jaelim-ms: Good catch. Forgot to remove the line after testing.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: Done

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: This is very weird. My AS configuration should use spaces instead of tabs. It is not possible!!

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -698,31 +722,35 @@ public void run() {
...
+
+                        /* Clean up an error log file and map entry. */
+                        unprocessedIterator.remove();
+                        ErrorLogHelper.removeStoredErrorLogFile(unprocessedEntry.getKey());
+                    }
                 }
+
+                /* Processed crash report for the last session. */
+                if (isInstanceEnabled())
+                    mHandler.getLooper().quit();
```
guperrot: This can't work if the SDK is re-enabled later, I don't see code that re-creates a new HandlerThread the next time the SDK is enabled. The code that creates the HandlerThread should be moved to the init while enabled method I guess.

jaelim-ms: Nope. Basically all file related operations should be just one time job so it doesn't have any scenarios that require this handler by disable/enable.

guperrot: If the SDK is disabled while processing pending reports, it must resume when enabled again.

jaelim-ms: We are deleting all logs when the service gets disabled. There is nothing we need to take care of.

guperrot: But we still try to post runnables to Handler no if we reenable?

jaelim-ms: No. It goes through initialization steps but it won't post any runnables to that handler because the service already deleted all pending files when it disabled.

guperrot: Ok, but later when we move listFiles and delete calls to worker thread too we will need to do this modification as well.

PR #303, opened 2017-01-14
Fix race conditions in channel

guperrot:


sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
```diff
@@ -502,6 +535,20 @@ public void onFailure(Exception e) {
         });
     }
 
+    private synchronized void checkLogsAfterPut(GroupState groupState, int currentState) {
+        if (checkStateDidNotChange(groupState, currentState)) {
+            groupState.mPendingLogCount++;
+            MobileCenterLog.debug(LOG_TAG, "enqueue(" + groupState.mName + ") pendingLogCount=" + groupState.mPendingLogCount);
+
+            /* Increment counters and schedule ingestion if we are not disabled. */
+            if (!mEnabled) {
```
jaelim-ms: Can you remove `!` and swap `if` and `else` body?

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
```diff
@@ -339,93 +369,103 @@ private synchronized void triggerIngestion(final @NonNull String groupName) {
 
         /* Get a batch from Persistence. */
         final List<Log> batch = new ArrayList<>(groupState.mMaxLogsPerBatch);
+        final int currentState = mStateChangeCounter;
```
jaelim-ms: Can you rename the local variable? Getting confused about  **State**, better to have **Counter**

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
```diff
@@ -275,33 +297,41 @@ private void suspend(boolean deleteLogs, Exception exception) {
...
-                    handleFailureCallback(groupState);
-                } else {
-                    mPersistence.deleteLogs(groupState.mName);
-                }
+                deleteLogsOnSuspended(groupState, currentState, logs);
             }
         });
     }
 
+    private synchronized void deleteLogsOnSuspended(GroupState groupState, int currentState, List<Log> logs) {
```
jaelim-ms: This method is a recursion. Once the Channel starts deleting logs on suspend, it should keep deleting even though the state gets changed in the middle. This is more like a PM question but we need to decide whether Channel keeps calling failure callbacks upon state change or not.

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
```diff
@@ -275,33 +297,41 @@ private void suspend(boolean deleteLogs, Exception exception) {
...
+            if (logs.size() > 0 && groupState.mListener != null) {
+                for (Log log : logs) {
+                    groupState.mListener.onBeforeSending(log);
+                    groupState.mListener.onFailure(log, new CancellationException());
+                }
+            }
+            if (logs.size() >= CLEAR_BATCH_SIZE && groupState.mListener != null) {
+                deleteLogsOnSuspended(groupState);
+            } else {
+                mPersistence.deleteLogs(groupState.mName);
```
jaelim-ms: If state gets changed, it will stop deleting logs for previous calls. Seems not right. Just `checkStateDidNotChange` out of the method is enough I think.

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
```diff
@@ -110,6 +110,12 @@
     private Device mDevice;
 
     /**
+     * State checker. If this counter changes during a call to persistence, we have to ignore the result in the callback.
+     * Cancelling a database call would be unreliable, and if it's too fast you could still have the callback being called.
+     */
+    private int mStateChangeCounter;
```
jaelim-ms: This variable can only be changed in synchronized method as of now so it is not a problem at all. However I recommend to use AtomicReference just in case to prevent any mistakes from future implementations.
Just my opinion but what do you think?

guperrot: They say in Javadoc that its not a general replacement for int, and it has overhead and we still need synchronized for general thread safety (even before this patch the code was meant to be thread safe).

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
guperrot: I don't see the relation with the patch.

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
guperrot: When state change, we should stop what we are doing at the fist chance we get.

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
guperrot: Renaming the member variable instead.

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
guperrot: Sure.

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
guperrot: I will rename both to make it more explicit.

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
jaelim-ms: We have to make sure that the logs should be gone at the time it is disabled.
It will send logs after re-enable that were created before it was disabled. I expect all the logs will be discarded after it is disabled. When it is enabled again, it should be in a clean states without any pending logs.

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
jaelim-ms: That is right but I think it is not in this case.

sdk/mobile-center/src/test/java/com/microsoft/azure/mobile/channel/AbstractDefaultChannelTest.java
```diff
@@ -0,0 +1,119 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.powermock.api.mockito.PowerMockito.mockStatic;
+import static org.powermock.api.mockito.PowerMockito.whenNew;
+
+@SuppressWarnings("WeakerAccess")
+@PrepareForTest({DefaultChannel.class, IdHelper.class, DeviceInfoHelper.class, DatabasePersistenceAsync.class, MobileCenterLog.class})
+public class AbstractDefaultChannelTest {
+
+    static final String TEST_GROUP = "group_test";
```
jaelim-ms: `protected` for all default?

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/channel/DefaultChannel.java
guperrot: Settled offline with team, it's better to accidentally send logs that were disabled than accidentally call back old logs with a failure callback after it's already re-enabled. Either way it's a weird corner case but we had to choose between the two.

sdk/mobile-center/src/test/java/com/microsoft/azure/mobile/channel/AbstractDefaultChannelTest.java
guperrot: Code analysis would complain since they are in same package.

jaelim-ms: It won't complain (I tried). We can change that later if we have any sub-classes in other packages.

PR #324, opened 2017-02-15
Updates are now inactive if installed from Store

guperrot:


sdk/mobile-center-updates/src/main/java/com/microsoft/azure/mobile/updates/InstallerUtils.java
```diff
@@ -0,0 +1,54 @@
...
+     * Installer package names that are not app stores.
+     */
+    private static final Set<String> LOCAL_STORES = new HashSet<>();
+
+    /**
+     * Used to cache the result of {@link #isInstalledFromAppStore(String, Context)}, null until first call.
+     */
+    private static Boolean sInstalledFromAppStore;
+
+    /** Populate local stores. */
```
jaelim-ms: Expand the doc unless it is `/* ... */`

sdk/mobile-center-updates/src/main/java/com/microsoft/azure/mobile/updates/InstallerUtils.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.microsoft.azure.mobile.utils.MobileCenterLog;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Installer utils.
+ */
+class InstallerUtils {
```
jaelim-ms: Does it work for Amazon App Store? It usually downloads APKs from store and install so it can be considered as local install.

sdk/mobile-center-updates/src/main/java/com/microsoft/azure/mobile/updates/Updates.java
```diff
@@ -374,6 +374,13 @@ private synchronized void cancelPreviousTasks() {
     private synchronized void resumeUpdateWorkflow() {
         if (mForegroundActivity != null && !mWorkflowCompleted && isInstanceEnabled()) {
 
+            /* Don't go any further if the app was installed from an app store. */
```
jaelim-ms: This can be check whether the app is in background or not.

sdk/mobile-center-updates/src/main/java/com/microsoft/azure/mobile/updates/InstallerUtils.java
guperrot: It's not a doc, removed the extra * and kept 1 line.

sdk/mobile-center-updates/src/main/java/com/microsoft/azure/mobile/updates/Updates.java
guperrot: That does not change anything here.

sdk/mobile-center-updates/src/main/java/com/microsoft/azure/mobile/updates/InstallerUtils.java
guperrot: Amazon store also returns an installer just like Google Play does.

guperrot: I checked the installer on amazon: `com.amazon.venezia` so the code works as expected.

jaelim-ms: Good

sdk/mobile-center-updates/src/main/java/com/microsoft/azure/mobile/updates/Updates.java
jaelim-ms: I meant you can move this to the top of the method (above the if statement).
Basically we already know that the application cannot be updated at the time Updates service is initialized. You can check this early.

sdk/mobile-center-updates/src/main/java/com/microsoft/azure/mobile/updates/InstallerUtils.java
jaelim-ms: OK

sdk/mobile-center-updates/src/main/java/com/microsoft/azure/mobile/updates/Updates.java
guperrot: That would not simplify the code, today we start checking only when app starts and then we have a set of conditions. If I start moving that, I have also to move the isInstanceEnabled() check etc.... I would keep it as is.

PR #326, opened 2017-02-16
Store token if received before start by initializing storage

guperrot:


guperrot:
@jaelim-ms closing this as merging this pr would cause strict mode policy to complain. And fixing that would result in a much complex code (with a new async task and state checks), without resolving the original edge case of process that can crash at any time.

PR #329, opened 2017-02-22
Fix download checks

guperrot:


guperrot:
@jaelim-ms think about the load diff button or you might miss modifications on 2 big files ;)

sdk/mobile-center-updates/src/main/java/com/microsoft/azure/mobile/updates/Updates.java
```diff
@@ -160,9 +165,14 @@
...
-     * Current task to process download completion.
+     * Current task to check download state and act on it.
      */
-    private ProcessDownloadCompletionTask mProcessDownloadCompletionTask;
+    private CheckDownloadTask mCheckDownloadTask;
+
+    /**
+     * Remember if we checked download since our own process restarted.
+     */
+    private boolean mCheckedDownload;
```
guperrot: If the process restarts, you don't know if last time the process crashes while processing the download manager intent. So I need a in memory flag to do that check once per process.

guperrot:
The timestamp is commented line 450.

If the download was completed in background, we display a notification.

Since we can't monitor the install of the app we don't know when the app resumes (or even restarts for that matter) if the application has been updated. If not updated and we know the only thing we did was generating a notification, we were showing install U.I. on resume.

But since we were not checking for an app update, we could show the install U.I. after upgrade, this patch fixes that corner case.

PR #331, opened 2017-02-22
Add push module

MatkovIvan:
Added push module to SDK.
- Send push token in push installation log
- Added module settings in sasquatch app UI


jaelim-ms:
Please fix the broken build first. Sasquatch application has two flavors in build. One should be fine but the other shouldn't because it has dependencies with libraries that are in jcenter. You are introducing a new feature but it is not in jcenter yet. I recommend you to remove all Sasquatch implementation at the moment.

guperrot:
@MatkovIvan Or you can use reflection too, there is an example in the updates branch to enable updates only in the project build flavour.

guperrot:
Please run `./gradlew build coverageReport` locally to debug why the build fails. When it succeeds, use the coverage report in the `sdk/build` local folder to debug what tests are missing.

sdk/mobile-center-push/build.gradle
```diff
@@ -0,0 +1,10 @@
+//
+// :sdk:push
+//
+project.description = 'This package contains functionalities to recive push notifications for your application.'
+evaluationDependsOn(':sdk')
+
+dependencies {
+    compile project(':sdk:mobile-center')
+    compile 'com.google.android.gms:play-services-gcm:10.2.0'
```
guperrot: GCM is deprecated, we have to use Firebase SDK now.

sdk/mobile-center-push/src/main/java/com/microsoft/azure/mobile/push/Push.java
```diff
@@ -0,0 +1,256 @@
...
+     */
+    private static final String PREFERENCE_PREFIX = SERVICE_NAME + ".";
+
+    /**
+     * Preference key to store push token.
+     */
+    @VisibleForTesting
+    static final String PREFERENCE_KEY_PUSH_TOKEN = PREFERENCE_PREFIX + "push_token";
+
+
```
guperrot: Extra blank line that should be removed here.

sdk/mobile-center-push/src/main/java/com/microsoft/azure/mobile/push/Push.java
```diff
@@ -0,0 +1,256 @@
...
+    /**
+     * Shared instance.
+     */
+    @SuppressLint("StaticFieldLeak")
+    private static Push sInstance = null;
```
guperrot: In our coding conventions we don't initiliaze like that when its null, 0 or false.

sdk/mobile-center-push/src/main/java/com/microsoft/azure/mobile/push/Push.java
```diff
@@ -0,0 +1,256 @@
...
+     * Enqueue a push installation log.
+     *
+     * @param pushToken the push token value
+     */
+    private void enqueuePushInstallationLog(@NonNull String pushToken) {
+        if (isInactive())
+            return;
+        if (mPushTokenSent)
+            return;
+
```
guperrot: Either add a comment before each block or remove blank line.

sdk/mobile-center-push/src/main/java/com/microsoft/azure/mobile/push/Push.java
```diff
@@ -0,0 +1,256 @@
...
+        if (mPushToken != null) {
+            enqueuePushInstallationLog(mPushToken);
+        } else if (mSenderId != null) {
+            final PushTokenTask pushTokenTask = new PushTokenTask();
+            try {
+                pushTokenTask.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR);
+            } catch (final RejectedExecutionException e) {
+                MobileCenterLog.error(LOG_TAG, "THREAD_POOL_EXECUTOR saturated", e);
```
guperrot: Please add a TODO comment here, so that we use ASyncTaskUtils when its gets merged in the future.

sdk/mobile-center-push/src/main/java/com/microsoft/azure/mobile/push/Push.java
```diff
@@ -0,0 +1,256 @@
...
+    private synchronized void setInstanceSenderId(String senderId) {
+        mSenderId = senderId;
+    }
+
+    @VisibleForTesting
+    class PushTokenTask extends AsyncTask<Void, Void, String> {
+
+        @Override
+        protected String doInBackground(Void[] params) {
+            InstanceID instanceID = InstanceID.getInstance(mContext);
```
guperrot: GCM being deprecated, this code has to use Firebase SDK now.

sdk/mobile-center-push/src/test/java/com/microsoft/azure/mobile/push/PushTest.java
```diff
@@ -0,0 +1,161 @@
...
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.powermock.api.mockito.PowerMockito.mockStatic;
+import static org.powermock.api.mockito.PowerMockito.verifyStatic;
+import static org.powermock.api.mockito.PowerMockito.when;
+import static org.powermock.api.mockito.PowerMockito.whenNew;
+
+@SuppressWarnings("unused")
+@RunWith(PowerMockRunner.class)
```
guperrot: Ok that's the reason why code coverage fails to report percentage for tests originating from this class: if you use `@RunWith(PowerMockRunner.class)`, it will not cover classes that are in `@PrepareForTest`.

The solution for that is to remove `@RunWith(PowerMockRunner.class)` and instead use:

```
    @Rule
    public PowerMockRule mPowerMockRule = new PowerMockRule();
```

The rule is slower and requires the vm arg `-noverify` to test from studio so that's why we don't use it everywhere, but here you need the rule.

sdk/mobile-center-push/src/test/java/com/microsoft/azure/mobile/push/PushTest.java
```diff
@@ -0,0 +1,161 @@
...
+    public void handlePushToken() {
+        String testToken = "TEST";
+        Push push = Push.getInstance();
+        Channel channel = mock(Channel.class);
+        push.onChannelReady(mock(Context.class), channel);
+        push.handlePushToken(testToken);
+        verifyStatic(times(1));
+        StorageHelper.PreferencesStorage.putString(eq(PREFERENCE_KEY_PUSH_TOKEN), eq(testToken));
+
+        // For check enqueue only once
```
guperrot: We use // only to disable code temporarily.
Real comments are /*

sdk/mobile-center-push/build.gradle
```diff
@@ -0,0 +1,10 @@
+//
```
jaelim-ms: Remove the first 3 lines.

MatkovIvan: @jaelim-ms I used other modules as example. In analytics and crashes it is same.  So, why here must be different?

guperrot: We cleaned that in updates actually. Reason is that old Android Studio was showing build.gradle as editor title so it was confusing. Now it's showing the module name in editor tab title so we don't need this comment anymore.

apps/sasquatch/src/main/java/com/microsoft/azure/mobile/sasquatch/activities/SettingsActivity.java
```diff
@@ -80,7 +80,7 @@ public boolean isEnabled() {
                 @Override
                 public void setEnabled(boolean enabled) {
                     Crashes.setEnabled(enabled);
-                    crashesEnabledPreference.setChecked(Analytics.isEnabled());
+                    crashesEnabledPreference.setChecked(Crashes.isEnabled());
```
guperrot: Good catch :D Will backport to develop ;)

jaelim-ms: @guperrot fixed this on `develop` branch. Revert this change, merging develop will take care of this change.

guperrot: Just merging develop is enough, no need to revert I think.

jaelim-ms: The same change will have different commit IDs. If it doesn't matter, I don't care. 😄

guperrot: I think it does not matter, and git annotate will likely show the one from develop after merge.

sdk/mobile-center-push/src/main/java/com/microsoft/azure/mobile/push/Push.java
```diff
@@ -0,0 +1,195 @@
...
+     * Handle push token update success.
+     *
+     * @param token the push token value
+     */
+    @VisibleForTesting
+    synchronized void onTokenRefresh(@NonNull String token) {
+        mPushToken = token;
+        MobileCenterLog.debug(LOG_TAG, "Push token: " + mPushToken);
+        PreferencesStorage.putString(PREFERENCE_KEY_PUSH_TOKEN, mPushToken);
+        enqueuePushInstallationLog(mPushToken);
```
guperrot: If the token is refreshed but we already sent one during process lifetime we will skip sending which is incorrect behavior. It's not unusual that Google can replace the token with a "canonical" version so we should be ready to accept a new token any time. I would replace the boolean check by a string comparison with previous token value.

sdk/mobile-center-push/src/main/java/com/microsoft/azure/mobile/push/Push.java
```diff
@@ -0,0 +1,195 @@
...
+        super.onChannelReady(context, channel);
+        updatePushToken();
+    }
+
+    @Override
+    public synchronized void setInstanceEnabled(boolean enabled) {
+        super.setInstanceEnabled(enabled);
+        if (enabled) {
+            updatePushToken();
+        }
```
guperrot: The rule for modules when disabled is to reset any state, delete persisted state also.
Expected behavior when we disable and re-enable is to get token and send token again.

sdk/mobile-center-push/src/main/java/com/microsoft/azure/mobile/push/Push.java
```diff
@@ -0,0 +1,189 @@
...
+     * @param enabled current state.
+     */
+    private synchronized void applyEnabledState(boolean enabled) {
+        if (enabled && mChannel != null) {
+            String token = FirebaseInstanceId.getInstance().getToken();
+            if (token != null) {
+                onTokenRefresh(token);
+            }
+        } else {
+            /* Reset module state if disabled */
```
guperrot: Missing blank line before comment

guperrot:
Just waiting on missing code coverage now.

PR #334, opened 2017-02-23
Make in app updates inactive in debug

guperrot:


sdk/mobile-center-updates/src/test/java/com/microsoft/azure/mobile/updates/UpdatesBeforeDownloadTest.java
```diff
@@ -64,14 +64,13 @@ public ServiceCall answer(InvocationOnMock invocation) throws Throwable {
...
         when(releaseDetails.getId()).thenReturn("someId");
         when(ReleaseDetails.parse(anyString())).thenReturn(releaseDetails);
-        Context context = mock(Context.class);
-        when(context.getPackageName()).thenReturn("com.contoso");
-        PackageManager packageManager = mock(PackageManager.class);
-        when(context.getPackageManager()).thenReturn(packageManager);
-        when(packageManager.getPackageInfo("com.contoso", 0)).thenThrow(new PackageManager.NameNotFoundException());
+
+        /* Override mock answer by calling it once then configuring new answer. */
+        mPackageManager.getPackageInfo("com.contoso", 0);
```
jaelim-ms: Did you try without this line? I don't think you need this line.. Not sure.

guperrot: There is an answer configured in AbstractUpdateTests.setUp, when you configure two when rules, the firs when needs to be called once, the second when clause will be used for further answers.

jaelim-ms: I thought it will replace the mock even though it hasn't been called. Hmm. I should be careful when I need to do it.

guperrot: Actually you're right, chaining calls is not the same as splitting when clause on multiple lines.

PR #335, opened 2017-02-23
Fix locking logic to make sure we cover all state checks

guperrot:


guperrot:
There was a random code coverage miss because not locking properly in some tests.

PR #339, opened 2017-02-23
Feature/hide app secret

jaelim-ms:


sdk/mobile-center-updates/src/main/java/com/microsoft/azure/mobile/updates/Updates.java
```diff
@@ -602,7 +606,33 @@ private synchronized void getLatestReleaseDetails(@NonNull String updateToken) {
         Map<String, String> headers = new HashMap<>();
         headers.put(HEADER_API_TOKEN, updateToken);
         final Object releaseCallId = mCheckReleaseCallId = new Object();
-        mCheckReleaseApiCall = httpClient.callAsync(url, METHOD_GET, headers, null, new ServiceCallback() {
+        mCheckReleaseApiCall = httpClient.callAsync(url, METHOD_GET, headers, new HttpClient.CallTemplate() {
+
+            @Override
+            public String buildRequestBody() throws JSONException {
+
+                /* This method is getting called. */
```
guperrot: Actually this method is not called for a get.
I would either remove this comment or change it to something like "No request body as its a get call"

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/http/HttpUtils.java
```diff
@@ -65,4 +71,18 @@ public static boolean isRecoverableError(Throwable t) {
         }
         return false;
     }
+
+    public static String hideSecret(String secret) {
+
+        /* Cannot hide null or empty string. */
+        if (secret == null || secret.length() <= 0) {
```
guperrot: length cannot be negative. And you can use `.isEmpty()`

sdk/mobile-center-updates/src/main/java/com/microsoft/azure/mobile/updates/Updates.java
jaelim-ms: Oh..  Where is my **NOT**????

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/http/HttpUtils.java
jaelim-ms: Using `isEmpty()` instead.

PR #345, opened 2017-03-03
Upgrade to Android Studio 2.3

guperrot:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Tue Aug 16 11:54:01 PDT 2016
+#Thu Mar 02 17:25:18 PST 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-all.zip
```
jaredsburrows: If you are upgrading gradle, can you upgrade to `3.4` which is the latest. 

Then run `gradlew wrapper`.

Then change `bin` to `all`.

versions.gradle
```diff
@@ -7,5 +7,5 @@ ext {
     targetSdkVersion = 25
     compileSdkVersion = 25
     buildToolsVersion = '25.0.2'
-    supportLibVersion = '25.1.0'
+    supportLibVersion = '25.1.1'
```
jaredsburrows: Can you upgrade to `25.2.0`?

gradle/wrapper/gradle-wrapper.properties
guperrot: The version used here is related to Android Studio upgrade (it was done automatically by Android Studio), I'd prefer making sure we use the gradle version that we use is the recommended one for Android to avoid running into unexpected behavior.

jaredsburrows: Its just the min version for 2.3.0. 

It works fine: https://github.com/jaredsburrows/android-gif-example/blob/master/gradle/wrapper/gradle-wrapper.properties#L6.

`3.3` is the min for `2.3.0`, see here: https://developer.android.com/studio/releases/gradle-plugin.html#updating-gradle.

jaredsburrows:
LGTM!

versions.gradle
jaredsburrows: Thanks!

gradle/wrapper/gradle-wrapper.properties
guperrot: Ok ok, did the change via Android Studio and thus just change to 3.4 within IDE, no need to run the command you mentioned and changing to bin.

jaredsburrows:
Yeah. But running Gradle wrapper updates the .Jar, .Sh, and .Bat files. No
big deal though.

guperrot:
Trying that `gradle wrapper` thing.

PR #348, opened 2017-03-04
Add all project dependencies to source when build generates javadoc

jaelim-ms:


jaelim-ms:
Fixed the issue. Moving dependencies to classpath.

PR #352, opened 2017-03-08
Don't show update dialog for incompatible release (api level)

guperrot:


sdk/mobile-center-distribute/src/androidTest/java/com/microsoft/azure/mobile/distribute/ReleaseDetailsTest.java
```diff
@@ -136,16 +148,63 @@ public void nullReleaseNotes() throws JSONException {
...
+                "version: '14'," +
+                "short_version: '2.1.5'," +
+                "release_notes: 'Fix a critical bug, this text was entered in Mobile Center portal.'," +
+                "download_url: 'http://download.thinkbroadband.com/1GB.zip'" +
+                "}";
+        ReleaseDetails.parse(json);
+    }
+
+    @Test
+    public void acceptApiLevelAsString() throws JSONException {
```
jaelim-ms: I can see swagger that its type is string. Pretty sad that it is not integer. :cry:

PR #361, opened 2017-03-17
Keep token if error is JSON "no_releases_for_user"

guperrot:


sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/ErrorDetails.java
```diff
@@ -0,0 +1,48 @@
+package com.microsoft.azure.mobile.distribute;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Release details JSON schema.
+ */
+class ErrorDetails {
```
jaelim-ms: There is a `message` in payload. It would be better to parse the message and log it.

guperrot: If the code is not one we know, the full payload is displayed already in the error. Why would we parse it in that case?

jaelim-ms: I forgot to think that there is an app id which can be confusing.

PR #367, opened 2017-03-21
Improve release cache and expiration

guperrot:


sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -419,10 +420,17 @@ private synchronized void resumeDistributeWorkflow() {
             int downloadState = getStoredDownloadState();
             if (mReleaseDetails == null && downloadState != DOWNLOAD_STATE_COMPLETED) {
                 mReleaseDetails = DistributeUtils.loadCachedReleaseDetails();
+
+                /* If cached release is optional and we have network, we should not reuse it. */
+                if (mReleaseDetails != null && !mReleaseDetails.isMandatoryUpdate() &&
+                        NetworkStateHelper.getSharedInstance(mContext).isNetworkConnected() &&
```
jaelim-ms: Network can be disconnected after receiving new release details. You already have a new state which is `DOWNLOAD_STATE_AVAILEABLE`, I don't think you need to consider network connection at this time.

guperrot: No this check is to avoid showing an old update dialog for optional update if I can get a more recent dialog. Remember we don't replace dialogs and wait for next time.

jaelim-ms: I remember what we decide to do for this case. What I'm saying is that why do you care about network connection at this time?

guperrot: See previous reply

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeMandatoryDownloadTest.java
```diff
@@ -359,4 +360,103 @@ private void verifyWithInvalidOrMissingCachedJson() throws Exception {
...
+        verifyStatic(times(2));
+        PreferencesStorage.putInt(PREFERENCE_KEY_DOWNLOAD_STATE, DOWNLOAD_STATE_INSTALLING);
+        Distribute.getInstance().onActivityPaused(mActivity);
+
+        /* Change what the next detected release will be: a more recent mandatory one. */
+        ReleaseDetails releaseDetails = mock(ReleaseDetails.class);
+        when(releaseDetails.getId()).thenReturn(5);
+        when(releaseDetails.getVersion()).thenReturn(8);
+        when(releaseDetails.getDownloadUrl()).thenReturn(mDownloadUrl);
+        when(releaseDetails.isMandatoryUpdate()).thenReturn(false);
```
jaelim-ms: Parameterized ??

guperrot: No too difficult to use that jUnit feature in those after download tests, already tried and it was a mess ;)

PR #368, opened 2017-03-21
Error attachment implemented as separate log

bmourat:
Error attachment implemented as ErrorAttachmentLog class which is send separatley from error log


apps/sasquatch/src/main/java/com/microsoft/azure/mobile/sasquatch/activities/MainActivity.java
```diff
@@ -142,11 +146,19 @@ public void onClick(DialogInterface dialog, int which) {
                 return true;
             }
 
-            /* TODO (getErrorAttachment): Re-enable error attachment when the feature becomes available. */
-//            @Override
-//            public ErrorAttachment getErrorAttachment(ErrorReport report) {
-//                return ErrorAttachments.attachment("This is a text attachment.", "This is a binary attachment.".getBytes(), "binary.txt", "text/plain");
-//            }
+            @Override
+            public Iterable<ErrorAttachmentLog> getErrorAttachment(ErrorReport report) {
```
guperrot: Add a s to method name as we now have ability to attach several attachments.

apps/sasquatch/src/main/java/com/microsoft/azure/mobile/sasquatch/activities/MainActivity.java
```diff
@@ -142,11 +146,19 @@ public void onClick(DialogInterface dialog, int which) {
...
+                ErrorAttachmentLog textLog = ErrorAttachments.attachmentWithText("This is a text attachment.", "text.txt");
+                ErrorAttachmentLog binaryLog = ErrorAttachments.attachmentWithBinary("This is a binary attachment.".getBytes(), "binary.txt");
+
+                textLog.setId(UUID.randomUUID());
+                binaryLog.setId(UUID.randomUUID());
+
+                textLog.setErrorId(UUID.fromString(report.getId()));
+                binaryLog.setErrorId(UUID.fromString(report.getId()));
+
```
guperrot: Our coding conventions is that if we put a blank line it has to have a comment.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/CrashesListener.java
```diff
@@ -31,15 +31,14 @@
     boolean shouldAwaitUserConfirmation();
 
     /**
-     * Called to get additional information to be attached to a crash report before sending.
-     * Attachment is an optional so this method can also return <code>null</code>.
+     * Called to get additional information to be send as separate ErrorAttachmentLog logs
```
guperrot: send => sent

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/ErrorAttachments.java
```diff
@@ -1,19 +1,20 @@
...
 /**
  * Error attachment utilities.
  */
-@SuppressWarnings("WeakerAccess")
-/* TODO (getErrorAttachment): Re-enable error attachment in javadoc when the feature becomes available. Add @ before link. */
-final class ErrorAttachments {
+public final class ErrorAttachments {
+
+    public static final String CONTENT_TYPE_TEXT_PLAIN  = "text/plain";
+    public static final String CONTENT_TYPE_BINARY      = "binary";
```
guperrot: Blank lines between fields. Comments are also recommended. Also we use default formatter of Android Studio, does not seem to be the case here.

guperrot: binary is not a Mime type, binary can be anything like an image or video or whatever, user will specify the type for all binaries.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -735,14 +739,17 @@ public void run() {
...
-//                            errorLogReport.log.setErrorAttachment(attachment);
                         mChannel.enqueue(errorLogReport.log, ERROR_GROUP);
 
+                        Iterable<ErrorAttachmentLog> attachments = mCrashesListener.getErrorAttachment(errorLogReport.report);
+                        if (attachments == null) {
+                            MobileCenterLog.debug(LOG_TAG, "CrashesListener.getErrorAttachment returned null, no additional information will be attached to log: " + errorLogReport.log.getId().toString());
+                        }
+                        else {
+                            for (ErrorAttachmentLog attachment : attachments) {
+                                mChannel.enqueue(attachment, ERROR_GROUP);
```
guperrot: Since the static util method can return null, we can expect to have null references in the iterable.

And the static method could be bypassed, I really think we should move the validation call to this place (validation itself can be moved to another class if you want to minimize code file length increase).

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/ErrorAttachments.java
```diff
@@ -22,52 +23,55 @@
...
-            attachment.setBinaryAttachment(binaryAttachment);
-        } else if (fileName != null || contentType != null) {
-            MobileCenterLog.warn(LOG_TAG, "Binary attachment file name and content ignored as data is null");
+        if (fileName == null) {
+            MobileCenterLog.warn(LOG_TAG, "Null file name passed to attachment method, returning null");
+            return null;
+        }
+
+        String content = text;
+        if(contentType.equals(CONTENT_TYPE_BINARY)){
```
guperrot: See my comment about binary, the test should be `!contentType.startsWith("text/")`.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
guperrot: Need to generate the id field, and set error_id to the id of the related crash here.

PR #379, opened 2017-03-24
Fix 2 scenarios where progress dialog can be stuck at 100%

guperrot:
* If receiving completion callback and updating progress at the same time.
* If application exits while processing completion callback.


guperrot:
Yes it's behaving as before the patch. Basically we keep processing v1, put v2 in cache, and if you don't install v1, next time it will suggest v2 (to avoid replacing dialogs in the middle of the workflow). This is covered by unit tests also.

PR #396, opened 2017-04-14
Add unit test for Distribute customization

jaelim-ms:


jaelim-ms:
This code still has *Ignore* which will be removed separately.

guperrot:
For validating the PR we need to do the reflection stuff (or build flavour specific sources) in sasquatch.
That's why bitrise failed.

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeCustomizationTest.java
```diff
@@ -0,0 +1,367 @@
...
+
+        /* Verify POSTPONE has been processed. */
+        verify(distribute).completeWorkflow();
+
+        /* Call handleUserUpdateAction. */
+        distribute.handleUserUpdateAction(UserUpdateAction.IGNORE);
+
+        /* Verify IGNORE has been processed. */
+        verifyStatic();
+        StorageHelper.PreferencesStorage.putInt(PREFERENCE_KEY_IGNORED_RELEASE_ID, details.getId());
```
guperrot: There is a bug if that works, after completeWorkflow or after selecting 1 of the user command, we cannot send another user action for the same workflow.

jaelim-ms:
Oops. I forgot about that. :cry:

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeCustomizationTest.java
guperrot: (already processed postpone)

guperrot:
I also have code that prevents sending user command if the update dialog was not customized, need a test for that.

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeCustomizationTest.java
jaelim-ms: Yeah.. That makes sense. The previous workflow is completed, it won't allow to work on same release. I'll fix that.

guperrot: Please add a comment this is very confusing, would have been better to have different test methods and also checking that further commands don't do anything after the first one.

jaelim-ms: Instead of adding a comment, I split into multiple tests. Should be good now.

jaelim-ms: And I added a test for your second thing. We will eventually remove IGNORE from distribute listener so I didn't write extra for IGNORE.

PR #398, opened 2017-04-18
Feature/distribute listener

guperrot:


sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/UpdateAction.java
```diff
@@ -0,0 +1,28 @@
...
+@IntDef({
+        UpdateAction.DOWNLOAD,
+        UpdateAction.POSTPONE
+})
+public @interface UpdateAction {
+
+    /**
+     * Action to trigger the download of the release.
+     */
+    int DOWNLOAD = DialogInterface.BUTTON_POSITIVE;
```
jaelim-ms: We decided to use UPDATE.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DistributeListener.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Listener for the Distribute allowing customization.
+ */
+@SuppressWarnings("unused")
+public interface DistributeListener {
+
+    /**
+     * Called whenever a new release is available to download.
+     * <p>
+     * If user does not action the release (either postpone or download), this callback
```
jaelim-ms: It is up to you but this is not `update`?

PR #400, opened 2017-04-19
enableFirebaseAnalytics method added to Push module 

bmourat:
Firebase Analytics is disabled by default, and only after user calls enableFirebaseAnalytics it is enabled

### DIFF
--- a/apps/sasquatch/src/main/java/com/microsoft/azure/mobile/sasquatch/activities/SettingsActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/azure/mobile/sasquatch/activities/SettingsActivity.java
@@ -359,7 +359,7 @@ public class SettingsActivity extends AppCompatActivity {
             editor.apply();
         }
 
-        private boolean isFirebaseEnabled(){
+        private boolean isFirebaseEnabled() {
             return MainActivity.sSharedPreferences.getString(FIREBASE_ENABLED_KEY, null) != null;
         }
 
--- a/apps/sasquatch/src/main/java/com/microsoft/azure/mobile/sasquatch/activities/SettingsActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/azure/mobile/sasquatch/activities/SettingsActivity.java
@@ -31,10 +31,9 @@ import java.util.UUID;
 
 import static com.microsoft.azure.mobile.sasquatch.activities.MainActivity.APP_SECRET_KEY;
 import static com.microsoft.azure.mobile.sasquatch.activities.MainActivity.LOG_URL_KEY;
+import static com.microsoft.azure.mobile.sasquatch.activities.MainActivity.FIREBASE_ENABLED_KEY;
 
 public class SettingsActivity extends AppCompatActivity {
-
-    private static final String FIREBASE_ENABLED_KEY = "firebaseEnabled";
 
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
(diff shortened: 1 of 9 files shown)


sdk/mobile-center-push/src/main/java/com/microsoft/azure/mobile/push/Push.java
```diff
@@ -111,6 +117,29 @@ public static void setEnabled(boolean enabled) {
...
+    /**
+     * Enable or disable firebase analytics collection.
+     *
+     * @param context the context to retrieve FirebaseAnalytics instance.
+     * @param enabled <code>true</code> to enable, <code>false</code> to disable.
+     */
+    @SuppressWarnings("MissingPermission")
+    private static void setFirebaseAnalyticsEnabled(@NonNull Context context, boolean enabled) {
+        FirebaseAnalyticsUtils.setEnabled(context, enabled);
+        PreferencesStorage.putBoolean(PREFERENCE_KEY_ANALYTICS_ENABLED, enabled);
```
guperrot: Don't use storage, just a boolean field, the decision to enable firebase can be made at each run and people will call the method before start (usually right before start).

Sorry if the spec was not clear we should have clarified that it's not like a setEnabled thing like in modules.

guperrot: To be clear:

* In every case call setAnalyticsCollectionEnabled(true) upon calling Push.enableFirebaseAnaltyics.
* If the method is called before start, store temp boolean to remember not to disable on start.

apps/sasquatch/src/main/java/com/microsoft/azure/mobile/sasquatch/activities/SettingsActivity.java
```diff
@@ -137,6 +141,31 @@ public boolean isEnabled() {
                         }
                     }
                 });
+
+                final Method enableFirebaseAnalytics = push.getMethod("enableFirebaseAnalytics", Context.class);
+                initCheckBoxSetting(R.string.mobile_center_push_firebase_state_key, isFirebaseEnabled(), R.string.mobile_center_push_firebase_summary_enabled, R.string.mobile_center_push_firebase_summary_disabled, new HasEnabled() {
+
+                    @Override
+                    public void setEnabled(boolean enabled) {
```
jaelim-ms: What do you want to do in this method? You are ignoring `enabled` value, is it intended?

bmourat: Yes, it was intended, since user can only enable firebase analytics we don't need this value. After app restart it will be again in disabled state.

guperrot: Will suggest a patch on your fork to have Sasquatch more consistent with this setting.

guperrot: @bmourat this is what I had in mind when creating the task: https://github.com/bmourat/mobile-center-sdk-android/pull/2.

This will be better for our internal demos ;)

guperrot: Looks good otherwise, will merge after you merge https://github.com/bmourat/mobile-center-sdk-android/pull/2

PR #402, opened 2017-04-20
Update distribute to match design specifications

guperrot:


sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -785,10 +777,34 @@ else if (Build.VERSION.SDK_INT >= releaseDetails.getMinApiLevel()) {
...
+     * @return false if release optional AND user has clicked ask me in day since less than 24 hours ago, true otherwise.
+     */
+    private boolean canUpdateNow(ReleaseDetails releaseDetails) {
+        if (releaseDetails.isMandatoryUpdate()) {
+            MobileCenterLog.debug(LOG_TAG, "Release is mandatory, ignoring any postpone action.");
+            return true;
+        }
+        long postponedTime = PreferencesStorage.getLong(PREFERENCE_KEY_POSTPONE_TIME, 0);
+        long postponedUntil = postponedTime + POSTPONE_TIME_THRESHOLD;
+        if (System.currentTimeMillis() < postponedUntil) {
```
jaelim-ms: Can you also add a condition if current time is greater than `postponedTime - POSTPONE_TIME_THRESHOLD`?
There is a scenario that if device time was wrong e.g. 1/1/2020 8:00 AM UTC and it is postponed, it will never prompt the dialog again once the time is back to current unless it is a mandatory update.
In iOS, we are handling this case since this doesn't make any sense. What do you think?

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -839,41 +855,39 @@ private synchronized void showUpdateDialog() {
...
-                    completeWorkflow(releaseDetails);
-                }
-            });
-            setOnCancelListener(dialogBuilder, releaseDetails);
         }
         mUpdateDialog = dialogBuilder.create();
         showAndRememberDialogActivity(mUpdateDialog);
     }
 
+    private String getAppName() {
```
jaelim-ms: We have to make sure what app name should be displayed. Application name from Context (or an app itself) or the one in ReleaseDetails.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -839,41 +855,39 @@ private synchronized void showUpdateDialog() {
...
-            dialogBuilder.setMessage(R.string.mobile_center_distribute_update_dialog_message);
-        else
-            dialogBuilder.setMessage(releaseNotes);
+        String message;
+        if (releaseDetails.isMandatoryUpdate()) {
+            message = mContext.getString(R.string.mobile_center_distribute_update_dialog_message_mandatory);
+        } else {
+            message = mContext.getString(R.string.mobile_center_distribute_update_dialog_message_optional);
+        }
+        message = String.format(message, getAppName(), mReleaseDetails.getShortVersion(), mReleaseDetails.getVersion());
```
jaelim-ms: You can make `getInstallReadyMessage` method taking a resource ID as a parameter and then use it.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
jaelim-ms: If you do so, you also need to rename the method.

jaelim-ms:
One more question here; where is `View release notes` button?

guperrot:
@jaelim-ms https://github.com/Microsoft/mobile-center-sdk-android/pull/402#issue-222930701

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
guperrot: @jaelim-ms why not just checking current time > postponeTime? The minus seems unnecessary.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
guperrot: will do

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
guperrot: I am asking designers.

guperrot: So using app name from bundle as better for localization. Designers will talk to distribute team to improve portal first time install experience to match app.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
jaelim-ms: I got confused what I did in iOS. I don't do the minus thing. 😄 
We agreed it offline, please check what you said.

PR #415, opened 2017-04-27
Set setCustomProperties API to protected

MatkovIvan:


MatkovIvan:
Oops, I missed this because of jcenter flavor

PR #433, opened 2017-05-11
Always use base64 for attachments

guperrot:


sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/ingestion/models/ErrorAttachmentLog.java
```diff
@@ -46,19 +44,14 @@
...
-    private String contentType;
-
-    /**
      * File name.
      */
     private String fileName;
 
     /**
-     * Data (plain text or base64 string for binary data).
+     * Data encoded as base64.
```
MatkovIvan: Data not encoded as base64 here. You are store data as is and encode only for json serialize.

versions.gradle
```diff
@@ -2,7 +2,7 @@
 
 ext {
     versionCode = 19
-    versionName = '0.8.1'
+    versionName = '0.9.0'
```
jaelim-ms: I think 0.8.2 should be enough

guperrot: No, error attachments is a feature

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/ingestion/models/ErrorAttachmentLog.java
guperrot: This is just the description we copy from swagger though.

guperrot: adapted comment

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/ingestion/models/ErrorAttachmentLog.java
```diff
@@ -68,52 +74,25 @@
...
-
-    /**
-     * Build an error attachment log with text and binary suitable for using in {link CrashesListener#getErrorAttachments(ErrorReport)}.
-     *
      * @param data        binary data.
-     * @param text        text to attach to the attachment log.
      * @param fileName    file name to use in error attachment log.
      * @param contentType binary data MIME type.
-     * @return ErrorAttachmentLog attachment or null if text and data are null.
+     * @return ErrorAttachmentLog attachment or null if null data is passed.
```
jaelim-ms: `if data is null.` instead?

guperrot: oops, its never null actually, old comment

PR #450, opened 2017-05-30
Fix a crash when calling AnyService.isEnabled() before start

guperrot:
This returns false and prints an error now.

### DIFF
--- a/sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/MobileCenter.java
+++ b/sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/MobileCenter.java
@@ -308,7 +308,7 @@ public class MobileCenter {
         queueCustomProperties(properties);
     }
 
-     /**
+    /**
      * {@link #isConfigured()} implementation at instance level.
      */
     private synchronized boolean isInstanceConfigured() {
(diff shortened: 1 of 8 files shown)


guperrot:
I don't think we have other bugs like this.

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/MobileCenter.java
```diff
@@ -491,8 +491,10 @@ private synchronized void setInstanceEnabled(boolean enabled) {
             mUncaughtExceptionHandler.unregister();
         }
 
-        /* Update state. */
```
jaelim-ms: Why do you need this change?

guperrot: Because service checks isInstanceEnabled which an API that depends on core.isEnabled

jaelim-ms: Had a conversation offline. `setInstanceEnabled` is looking at the storage to determine the core is disabled or enabled while core state is changing. We do need this.

PR #452, opened 2017-05-31
Don't decorate start service log with session

guperrot:


guperrot:
UWP has this check too, what about iOS @jaelim-ms ?

PR #453, opened 2017-06-01
Print warning to log if there more than 2 attachments per crash report

erychagov:


sdk/mobile-center-crashes/src/test/java/com/microsoft/azure/mobile/crashes/CrashesTest.java
```diff
@@ -1170,4 +1179,41 @@ public void saveWrapperSdkErrorLogIOException() throws IOException, JSONExceptio
...
+        when(ErrorLogHelper.getStoredErrorLogFiles()).thenReturn(new File[]{mock(File.class)});
+
+        when(StorageHelper.InternalStorage.read(any(File.class))).thenReturn("");
+        when(StorageHelper.InternalStorage.readObject(any(File.class))).thenReturn(mock(Throwable.class));
+
+        Crashes crashes = Crashes.getInstance();
+        crashes.setInstanceListener(listener);
+        crashes.setLogSerializer(logSerializer);
+
+        Crashes spy = spy(crashes);
```
guperrot: You can't spy on this as getInstance will return the original object and not the spy.
You can instead use `whenNew` to mock the new ErrorReport that gets created in Crashes.
There is plenty of `whenNew` examples in our project for reference.

PR #459, opened 2017-06-10
Fix strict mode in session tracker

guperrot:


sdk/mobile-center-analytics/src/main/java/com/microsoft/azure/mobile/analytics/Analytics.java
```diff
@@ -419,26 +442,34 @@ protected synchronized void applyEnabledState(boolean enabled) {
...
+                queuePage(name, properties);
             }
         });
     }
 
     /**
+     * Enqueue page log now.
+     */
+    @WorkerThread
+    private void queuePage(String name, Map<String, String> properties) {
```
jaelim-ms: You probably want to do the same thing like `queueEvent`?

guperrot: It's used in only 1 place for queueEvent, do you still want the change?

jaelim-ms: Hmm.. Let's leave it.

PR #460, opened 2017-06-11
Auto fragment tracker

AliMajed10:
You know some apps are fragment base or have some tabs that implement with Fragment, So Auto-activity-tracking is not enough. 
But auto-Fragment tracking makes many unimportant logs, so it semi-auto-fragment-tracker is more efficient.That means fragments should be added to Analytic as a list (or fragment) to track them.

````
String[] FragmentNames = 
                new String []{"BookFragment.class.getName()","ListFragment.class.getName()"};
Analytics.addToAutoSubPagesTrackList(getSupportFragmentManager(),FragmentNames);

MobileCenter.start(
                getApplication(),
                {secret code} ,
                Analytics.class ,
                Crashes.class);

````


guperrot:
Hi, the backend does not support page logs yet, there is no ETA for that.
The API was removed from public surface.

AliMajed10:
Thanks for the review.
I know backend does not support page tracking But it exists in SDK So I thought maybe, it is ok to add this close and useful part too. Although, backend doesn't support it now. So I did it like what you did in page tracking and it is disabled (Although I should be made `addToAutoSubPagesTrackList()` private)  But if it is not a useful feature to add to the backend, yes, it doesn't need to add to the SDK too.

guperrot:
We will look at this in the future if and when pages become a real thing.

AliMajed10:
Thanks
@guperrot

guperrot:
Will reopen later when the topic becomes current. We won't forget.

PR #463, opened 2017-06-15
Fix android 8 update notification

guperrot:


guperrot:
The new text is the only reason not to merge yet, you can still review.

PR #471, opened 2017-06-22
Change Sasquatch icon for project dependency build

jaelim-ms:
Approve if you like, otherwise feel free to close. 😄


guperrot:
@jaelim-ms I can approve that one, but would be better to use the real pink Mobile Center logo.

PR #477, opened 2017-07-03
Use PackageManager.canRequestPackageInstalls() on Android O

MatkovIvan:
> Starting from O, apps should use canRequestPackageInstalls()

https://developer.android.com/reference/android/provider/Settings.Secure.html#INSTALL_NON_MARKET_APPS

### DIFF
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -29,6 +29,9 @@ subprojects {
 
         defaultConfig {
             consumerProguardFiles 'proguard-rules.pro'
+
+            buildConfigField 'int', "MIN_SDK_VERSION", "${ext.minSdkVersion}"
+            buildConfigField 'int', "TARGET_SDK_VERSION", "${ext.targetSdkVersion}"
         }
 
         buildTypes {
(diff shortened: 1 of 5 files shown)


MatkovIvan:
@guperrot 
I guess pass min and target sdk versions to build config via gradle is better than hardcode these in test loops. It's allow us don't miss possible issues on update target version. Anyway let me know what you think about it.

sdk/build.gradle
```diff
@@ -29,6 +29,9 @@ subprojects {
 
         defaultConfig {
             consumerProguardFiles 'proguard-rules.pro'
+
+            buildConfigField 'int', "MIN_SDK_VERSION", "${ext.minSdkVersion}"
+            buildConfigField 'int', "TARGET_SDK_VERSION", "${ext.targetSdkVersion}"
```
guperrot: I did not understand your comment about the need for this.

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeWarnUnknownSourcesTest.java
```diff
@@ -276,6 +284,48 @@ public void clickSettingsGoBackWithoutEnabling() throws Exception {
...
+    public PowerMockRule mPowerMockRule = new PowerMockRule();
+    @Test
+    @PrepareForTest(Build.class)
+    @SuppressLint("InlinedApi")
+    public void clickSettingsOnAndroidO() throws Exception {
+
+        /* Save original SDK_INT. */
+        final int SDK_INT = Build.VERSION.SDK_INT;
+
+        /* Since Whitebox.setInternalState don't work here, do it manually. */
```
guperrot: That's why we use TestUtils in other places in distribution already.

guperrot:
I am okay with https://github.com/Microsoft/mobile-center-sdk-android/pull/477#issuecomment-312697377

guperrot:
Tested on real devices on all api ranges.

PR #494, opened 2017-08-07
Implement public in-app updates based on initial spec

guperrot:


sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -641,21 +649,28 @@ else if (mUnknownSourcesDialog != null) {
                 return;
             }
 
-            /* Check if we have previous stored the update token. */
+            /*
+             * Check if we have previously stored the redirection parameters,
+             * note that distribution group is missing in previous SDK releases.
```
jaelim-ms: I don't think we need the 2nd line unless the version is specified. It is saying previous SDK but we don't actually know what the "previous" is.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -710,39 +725,53 @@ synchronized void completeWorkflow() {
...
+            mBeforeStartUpdateToken = updateToken;
+            mBeforeStartDistributionGroupId = distributionGroupId;
         } else if (requestId.equals(PreferencesStorage.getString(PREFERENCE_KEY_REQUEST_ID))) {
-            String encryptedToken = CryptoUtils.getInstance(mContext).encrypt(updateToken);
-            PreferencesStorage.putString(PREFERENCE_KEY_UPDATE_TOKEN, encryptedToken);
+            if (updateToken != null) {
+                String encryptedToken = CryptoUtils.getInstance(mContext).encrypt(updateToken);
+                PreferencesStorage.putString(PREFERENCE_KEY_UPDATE_TOKEN, encryptedToken);
+            } else {
+                PreferencesStorage.remove(PREFERENCE_KEY_UPDATE_TOKEN);
```
jaelim-ms: Do you need to remove the key? If we keep it here, we can get a private download for update even though the current app is from public. Is this based on spec?

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeBeforeApiSuccessTest.java
```diff
@@ -315,6 +357,139 @@ public boolean matches(Object argument) {
...
+        when(UUIDUtils.randomUUID()).thenReturn(requestId);
+        when(PreferencesStorage.getString(PREFERENCE_KEY_REQUEST_ID)).thenReturn(requestId.toString());
+
+        /* Start and resume: open browser. */
+        start();
+        Distribute.getInstance().onActivityResumed(mActivity);
+        verifyStatic();
+        String url = DistributeConstants.DEFAULT_INSTALL_URL;
+        url += String.format(UPDATE_SETUP_PATH_FORMAT, "a");
+        url += "?" + PARAMETER_RELEASE_HASH + "=" + TEST_HASH;
```
jaelim-ms: You can add line 389 to 394 in String.format. Or, maybe StringBuffer (or single line so that compiler can utilize it).

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
guperrot: This is just to make what is in storage consistent as a public group should not have a token.
In reality we open browser only once but this is better for future proofness to have this in place if we start refreshing info from browser in future.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
guperrot: Previous as in 0.11.0.
In the future the or will be replaced by an and.
Again writing code that way makes it more simpler and is future proof.

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeBeforeApiSuccessTest.java
guperrot: This would not make the test code more readable I believe. Also StringBuilder, if you look at generated byte code of += it uses StringBuilder, += is syntaxic sugar.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
jaelim-ms: I agree with your last sentence but what I concerned was that is just saying "previous". Please track it and replace to a specific version after release.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
jaelim-ms: What I meant was this scenario.
1. An app has been installed from private distribution group.
2. The app opened and completed all install processes.
3. A newer version has been installed not from in-app-update but from install URL. -> This is still possible.
4. After opening the app, update token will be wiped out. -> We can prevent this case without the line and the app will still be able to get newer versions from private distribution group.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -135,14 +137,19 @@
...
     /**
-     * In memory token if we receive deep link intent before onStart.
+     * In memory request identifier if we receive deep link intent before onStart.
      */
-    private String mBeforeStartUpdateToken;
+    private String mBeforeStartRequestId;
 
     /**
-     * In memory request identifier if we receive deep link intent before onStart.
+     * In memory distribution group if we receive deep link intent before onStart.
```
jaelim-ms: Let's say distribution group id (or identifier).

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
guperrot: The browser is opened only once so your scenario cannot happen @jaelim-ms 

The line is just for storage consistency in case for some reason there was a token we make sure we are not confused between public and private groups.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
guperrot: I can add a TODO was we don't know the version yet.

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeBeforeApiSuccessTest.java
jaelim-ms: What I meant at the last was to do like below
```
String url = DistributeConstants.DEFAULT_INSTALL_URL +
    String.format(UPDATE_SETUP_PATH_FORMAT, "a") +
    "?" + PARAMETER_RELEASE_HASH + "=" + TEST_HASH +
    "&" + PARAMETER_REDIRECT_ID + "=" + mContext.getPackageName() +
    "&" + PARAMETER_REQUEST_ID + "=" + requestId.toString() +
    "&" + PARAMETER_PLATFORM + "=" + PARAMETER_PLATFORM_VALUE;
```
This is unit test so not a big deal.

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeBeforeApiSuccessTest.java
```diff
@@ -315,6 +357,139 @@ public boolean matches(Object argument) {
...
+
+            @Override
+            public boolean matches(Object argument) {
+                return argument.toString().startsWith(DistributeConstants.DEFAULT_API_URL);
+            }
+        }), anyString(), eq(headers), any(HttpClient.CallTemplate.class), any(ServiceCallback.class));
+
+        /* Call is still in progress. If we restart app, nothing happens we still wait. */
+        restartResumeLauncher(mActivity);
+        Distribute.getInstance().onActivityPaused(mActivity);
```
jaelim-ms: restartResumeLauncher will call the line 455 - 459. Isn't it duplicate code block?

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeHttpTest.java
```diff
@@ -79,6 +79,34 @@ public void onBeforeCalling() throws Exception {
     }
 
     @Test
+    public void onBeforeCallingWithoutToken() throws Exception {
```
jaelim-ms: You already tested this in the above method. Isn't it a duplicate test?
Look at line 60, it doesn't have token in header.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
jaelim-ms: Let's do that. This will be released by me or after 3 weeks and we can easily forget about it.

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeBeforeApiSuccessTest.java
guperrot: I still think string.format of a new pattern just for test will actually make this code less maintainable and readable.
As for StringBuilder I already replied for that one.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
jaelim-ms: This feedback shouldn't be an approval blocker but I need more test one this one. I approve the PR but will bring this back if I find something.
If users go to the mobile center install UI and try to download it, the scenario can be a valid one.

guperrot: It's not valid as of now as if we have something in storage we never open the browser again unless there is a fatal error which is another story...

Even if it was valid, to switch from private to public you would need to delete the token as keeping the token would make the SDK check for a private release.

PR #500, opened 2017-08-28
Add set HTTP request method

MatkovIvan:


sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/http/DefaultHttpClient.java
```diff
@@ -158,7 +159,7 @@ private static String doHttpCall(String urlString, String method, Map<String, St
             MobileCenterLog.verbose(LOG_TAG, "HTTP response status=" + status + " payload=" + response);
 
             /* Generate exception on failure. */
-            if (status != 200)
+            if (status != HttpURLConnection.HTTP_OK)
```
MatkovIvan: Should we use `status >= 200 && status < 300` here?

jaelim-ms: Good point. We should change that.

sdk/mobile-center/src/main/java/com/microsoft/azure/mobile/http/DefaultHttpClient.java
```diff
@@ -154,8 +154,10 @@ private static String doHttpCall(String urlString, String method, Map<String, St
             MobileCenterLog.verbose(LOG_TAG, "HTTP response status=" + status + " payload=" + response);
 
             /* Generate exception on failure. */
-            if (status != HttpURLConnection.HTTP_OK)
+            boolean successful = status >= 200 && status < 300;
+            if (!successful) {
```
jaelim-ms: Don't need to define the local variable. 😄 
Please also add unit tests.

MatkovIvan: Unit tests already here

PR #502, opened 2017-08-28
start v0.12.0

dhei:


versions.gradle
```diff
@@ -2,7 +2,7 @@
 
 ext {
     versionCode = 25
```
jaelim-ms: You also need to increase this one.

PR #512, opened 2017-09-14
Rum

guperrot:
Early look, no test yet


apps/sasquatch/src/main/java/com/microsoft/azure/mobile/sasquatch/activities/MainActivity.java
```diff
@@ -101,7 +102,9 @@ protected void onCreate(Bundle savedInstanceState) {
         }
 
         /* Start Mobile center. */
-        MobileCenter.start(getApplication(), sSharedPreferences.getString(APP_SECRET_KEY, getString(R.string.app_secret)), Analytics.class, Crashes.class, Distribute.class, Push.class);
+        RealUserMeasurements.setRumKey("d3632912f5214cce9b1351b826f9a0fb");
```
MatkovIvan: Move it to resources as well as other keys

sdk/mobile-center-rum/proguard-rules.pro
```diff
@@ -0,0 +1,17 @@
+# Add project specific ProGuard rules here.
+# By default, the flags in this file are appended to flags specified
+# in C:\Users\ivan.matkov\AppData\Local\Android\Sdk/tools/proguard/proguard-android.txt
```
MatkovIvan: I guess we should remove our local paths even from comments.
Maybe it's better to use the `proguard` file, as in `analytics` module?

sdk/mobile-center-rum/src/main/java/com/microsoft/azure/mobile/rum/RealUserMeasurements.java
```diff
@@ -0,0 +1,475 @@
...
+
+    /**
+     * Additional headers.
+     */
+    private static final Map<String, String> HEADERS = Collections.emptyMap();
+
+    /**
+     * Shared instance.
+     */
+    @SuppressLint("StaticFieldLeak")
```
MatkovIvan: Maybe store context as `WeakReference<Context>` instead?

guperrot: Doc claims storing app context has issue with instant run but never saw any issue in reality. Passing everywhere to a weak reference could lead to null references if it ever invalidates to another context and we would need to manage that. I would prefer to have proofs that the app context gets updated for real before doing anything.

sdk/mobile-center-rum/proguard-rules.pro
guperrot: cleanup should be done on develop then for the other ones. I can clean this one here.

sdk/mobile-center-rum/src/main/java/com/microsoft/azure/mobile/rum/RealUserMeasurements.java
```diff
@@ -0,0 +1,516 @@
...
+    private static String rumUniqueId() {
+        return UUIDUtils.randomUUID().toString().replace("-", "");
+    }
+
+    /**
+     * Implements {@link #setRumKey(String)} at instance level.
+     */
+    private synchronized void setInstanceRumKey(String rumKey) {
+        if (rumKey == null) {
+            MobileCenterLog.error(LOG_TAG, "Rum key is invalid.");
```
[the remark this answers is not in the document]
guperrot: I think I did it because of the trim and trim does not work on null.

PR #513, opened 2017-09-14
Cleanup proguard files

guperrot:


guperrot:
From https://github.com/Microsoft/mobile-center-sdk-android/pull/512#discussion_r138819604

PR #526, opened 2017-10-10
Wrapper SDK changes for Crashes

guperrot:
So that React Native can use JS listener for everything like other SDKs.

### DIFF
--- a/sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
+++ b/sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
@@ -32,6 +32,9 @@ import org.json.JSONException;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
@@ -145,6 +148,11 @@ public class Crashes extends AbstractMobileCenterService {
      * Flag to remember whether we already saved uncaught exception or not.
      */
     private boolean mSavedUncaughtException;
+
+    /**
+     * Automatic processing flag (automatic is the default).
+     */
+    private boolean mAutomaticProcessing = true;
 
     /**
      * Init.
@@ -511,19 +519,8 @@ public class Crashes extends AbstractMobileCenterService {
         }
     }
 
-    private boolean shouldStopProcessingPendingErrors() {
-        if (!isInstanceEnabled()) {
-            MobileCenterLog.info(LOG_TAG, "Crashes service is disabled while processing errors. Cancel processing all pending errors.");
-            return true;
-        }
-        return false;
-    }
-
     private void processPendingErrors() {
         for (File logFile : ErrorLogHelper.getStoredErrorLogFiles()) {
-            if (shouldStopProcessingPendingErrors()) {
-                return;
-            }
             MobileCenterLog.debug(LOG_TAG, "Process pending error file: " + logFile);
             String logfileContents = StorageHelper.InternalStorage.read(logFile);
             if (logfileContents != null) {
@@ -533,8 +530,10 @@ public class Crashes extends AbstractMobileCenterService {
                     ErrorReport report = buildErrorReport(log);
                     if (report == null) {
                         removeAllStoredErrorLogFiles(id);
-                    } else if (mCrashesListener.shouldProcess(report)) {
-                        MobileCenterLog.debug(LOG_TAG, "CrashesListener.shouldProcess returned true, continue processing log: " + id.toString());
+                    } else if (!mAutomaticProcessing || mCrashesListener.shouldProcess(report)) {
+                        if (!mAutomaticProcessing) {
+                            MobileCenterLog.debug(LOG_TAG, "CrashesListener.shouldProcess returned true, continue processing log: " + id.toString());
+                        }
                         mUnprocessedErrorReports.put(id, mErrorReportCache.get(id));
                     } else {
                         MobileCenterLog.debug(LOG_TAG, "CrashesListener.shouldProcess returned false, clean up and ignore log: " + id.toString());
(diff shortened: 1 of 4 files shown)


guperrot:
@achocron I agree with the second point but not with the first one.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -546,33 +545,54 @@ private void processPendingErrors() {
...
+            /* Proceed to check if user confirmation is needed. */
+            sendCrashReportsOrAwaitUserConfirmation();
         }
 
-        processUserConfirmation();
     }
 
-    private void processUserConfirmation() {
+    /**
+     * Check user confirmation or send logs now.
```
jaelim-ms: Can you add `@return` here? I don't know the purpose of the return value. Is it only for testing?

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -546,33 +545,54 @@ private void processPendingErrors() {
...
+                        handleUserConfirmation(SEND);
                     } else {
-                        MobileCenterLog.debug(LOG_TAG, "The flag for user confirmation is set to ALWAYS_SEND, continue sending logs");
+                        MobileCenterLog.debug(LOG_TAG, "CrashesListener.shouldAwaitUserConfirmation returned true, wait sending logs.");
                     }
-                    handleUserConfirmation(SEND);
                 }
             }
         });
+        return alwaysSend;
```
jaelim-ms: Why are you returning `alwaysSend`? If it returns `true` when it processes crashes, there is one case that is returning `false` when `mCrashesListener.shouldAwaitUserConfirmation()` returns `false`.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
```diff
@@ -650,23 +672,30 @@ public void run() {
...
+
+                        /* Send report. */
                         Map.Entry<UUID, ErrorLogReport> unprocessedEntry = unprocessedIterator.next();
                         ErrorLogReport errorLogReport = unprocessedEntry.getValue();
                         mChannel.enqueue(errorLogReport.log, ERROR_GROUP);
 
-                        Iterable<ErrorAttachmentLog> attachments = mCrashesListener.getErrorAttachments(errorLogReport.report);
-                        handleErrorAttachmentLogs(attachments, errorLogReport);
+                        /* Get attachments from callback in automatic processing. */
+                        if (mAutomaticProcessing) {
```
jaelim-ms: Do you need to check here? Shouldn't it process error attachments?

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
guperrot: I don't put Javadoc for private and package private. The JavaDoc can be found in WrapperSdkExceptionManager.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
guperrot: We need to return that so that React Native knows that value is persisted into storage, see: https://github.com/Microsoft/mobile-center-sdk-react-native/blob/9de1aa7d0edb00fd2a2ee3bfc320c2b3971a9d39/mobile-center-crashes/Crashes.js#L89

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
guperrot: No it works differently in React Native: https://github.com/Microsoft/mobile-center-sdk-react-native/blob/9de1aa7d0edb00fd2a2ee3bfc320c2b3971a9d39/mobile-center-crashes/Crashes.js#L104

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: This is why we need java doc for this method.
https://github.com/Microsoft/mobile-center-sdk-android/pull/526#discussion_r144171645
Can you just say here that the method returns an alwaysSend flag?

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: This is why we need java doc for this method even this is a private method.
https://github.com/Microsoft/mobile-center-sdk-android/pull/526#discussion_r144171645
Can you just say here that the method returns an alwaysSend flag?

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: I see.

sdk/mobile-center-crashes/src/main/java/com/microsoft/azure/mobile/crashes/Crashes.java
jaelim-ms: Is `mAutomaticProcess` only for RN or it is also used by Xamarin?

guperrot: Xamarin uses automatic processing and is not impacted by these changes.

guperrot: However we expect that Unity and Cordova would use it in the future if and when we will support crashes for those.

guperrot:
Will test on Xamarin before merging

PR #527, opened 2017-10-10
Feature/localization

jwallra:
Adding resource files and folders.

Still to do in a later PR - make this process a part of the loc job


jwallra:
@guperrot I'll fix that when I fix the script.  It will deposit it in the correct location.  I'll also fix the proliferation of branches.

PR #528, opened 2017-10-17
Only show in-app update failure message once per release

max-mironov:
The aim of this PR is to bring the similar functionality as was currently created at iOS SDK: https://github.com/Microsoft/mobile-center-sdk-ios/pull/773.

Any feedback and remarks are much appreciated.

### DIFF
--- a/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/BrowserUtils.java
+++ b/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/BrowserUtils.java
@@ -12,6 +12,8 @@ import android.support.annotation.VisibleForTesting;
 
 import com.microsoft.azure.mobile.utils.MobileCenterLog;
 
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.List;
 
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.LOG_TAG;
@@ -100,5 +102,25 @@ class BrowserUtils {
                 activity.startActivity(intent);
             }
         }
+    }
+
+    /**
+     * Append a query parameter to an existing URI.
+     *
+     * @param uri         initial uri.
+     * @param appendQuery parameter to append.
+     */
+    static String appendUri(@NonNull String uri, @NonNull String appendQuery) throws URISyntaxException {
+        URI oldUri = new URI(uri);
+        String newQuery = oldUri.getQuery();
+        if (newQuery == null) {
+            newQuery = appendQuery;
+        } else {
+            newQuery += "&" + appendQuery;
+        }
+        URI newUri = new URI(oldUri.getScheme(), oldUri.getAuthority(),
+                oldUri.getPath(), newQuery, oldUri.getFragment());
+
+        return newUri.toString();
     }
 }
--- a/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/BrowserUtils.java
+++ b/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/BrowserUtils.java
@@ -109,6 +109,8 @@ class BrowserUtils {
      *
      * @param uri         initial uri.
      * @param appendQuery parameter to append.
+     *
+     * @return uri string with appended query item.
      */
     static String appendUri(@NonNull String uri, @NonNull String appendQuery) throws URISyntaxException {
         URI oldUri = new URI(uri);
@@ -120,7 +122,6 @@ class BrowserUtils {
         }
         URI newUri = new URI(oldUri.getScheme(), oldUri.getAuthority(),
                 oldUri.getPath(), newQuery, oldUri.getFragment());
-
         return newUri.toString();
     }
 }
--- a/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DeepLinkActivity.java
+++ b/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DeepLinkActivity.java
@@ -9,6 +9,7 @@ import com.microsoft.azure.mobile.utils.MobileCenterLog;
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.EXTRA_DISTRIBUTION_GROUP_ID;
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.EXTRA_REQUEST_ID;
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.EXTRA_UPDATE_TOKEN;
+import static com.microsoft.azure.mobile.distribute.DistributeConstants.EXTRA_UPDATE_SETUP_FAILED;
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.LOG_TAG;
 
 /**
@@ -25,14 +26,19 @@ public class DeepLinkActivity extends Activity {
         String requestId = intent.getStringExtra(EXTRA_REQUEST_ID);
         String distributionGroupId = intent.getStringExtra(EXTRA_DISTRIBUTION_GROUP_ID);
         String updateToken = intent.getStringExtra(EXTRA_UPDATE_TOKEN);
+        String updateSetupFailed = intent.getStringExtra(EXTRA_UPDATE_SETUP_FAILED);
         MobileCenterLog.debug(LOG_TAG, getLocalClassName() + ".getIntent()=" + intent);
         MobileCenterLog.debug(LOG_TAG, "Intent requestId=" + requestId);
         MobileCenterLog.debug(LOG_TAG, "Intent distributionGroupId=" + distributionGroupId);
         MobileCenterLog.debug(LOG_TAG, "Intent updateToken passed=" + (updateToken != null));
+        MobileCenterLog.debug(LOG_TAG, "Intent updateSetupFailed passed=" + (updateSetupFailed != null));
 
         /* Store redirection parameters if both required values were passed. */
         if (requestId != null && distributionGroupId != null) {
             Distribute.getInstance().storeRedirectionParameters(requestId, distributionGroupId, updateToken);
+        } else if (updateSetupFailed != null){
+            // Otherwise just store error message to show update failure dialog in future.
+            Distribute.getInstance().storeUpdateSetupFailedParameter(updateSetupFailed);
         }
 
         /*
--- a/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DeepLinkActivity.java
+++ b/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DeepLinkActivity.java
@@ -36,8 +36,9 @@ public class DeepLinkActivity extends Activity {
         /* Store redirection parameters if both required values were passed. */
         if (requestId != null && distributionGroupId != null) {
             Distribute.getInstance().storeRedirectionParameters(requestId, distributionGroupId, updateToken);
-        } else if (updateSetupFailed != null){
-            // Otherwise just store error message to show update failure dialog in future.
+        } else if (updateSetupFailed != null) {
+
+            /* Otherwise just store error message to show update failure dialog in future. */
             Distribute.getInstance().storeUpdateSetupFailedParameter(updateSetupFailed);
         }
 
--- a/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DeepLinkActivity.java
+++ b/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DeepLinkActivity.java
@@ -36,10 +36,10 @@ public class DeepLinkActivity extends Activity {
         /* Store redirection parameters if both required values were passed. */
         if (requestId != null && distributionGroupId != null) {
             Distribute.getInstance().storeRedirectionParameters(requestId, distributionGroupId, updateToken);
-        } else if (updateSetupFailed != null) {
+        } else if (requestId != null && updateSetupFailed != null) {
 
             /* Otherwise just store error message to show update failure dialog in future. */
-            Distribute.getInstance().storeUpdateSetupFailedParameter(updateSetupFailed);
+            Distribute.getInstance().storeUpdateSetupFailedParameter(requestId, updateSetupFailed);
         }
 
         /*
--- a/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
+++ b/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
@@ -1129,7 +1129,7 @@ public class Distribute extends AbstractMobileCenterService {
     /**
      * Store failed package info hash in preferences.
      */
-    private synchronized void storeUpdateSetupFailedPackageHash(Dialog dialog) {
+    private synchronized void storeUpdateSetupFailedPackageHash(DialogInterface dialog) {
         if (mUpdateSetupFailedDialog == dialog) {
             PreferencesStorage.putString(PREFERENCE_KEY_UPDATE_SETUP_FAILED_PACKAGE_HASH_KEY, DistributeUtils.computeReleaseHash(mPackageInfo));
         } else {
@@ -1140,9 +1140,8 @@ public class Distribute extends AbstractMobileCenterService {
     /**
      * Redirect user to install url page in browser and clear current failed package hash from preferences.
      */
-    private synchronized void handleUpdateFailedDialogReinstallAction(Dialog dialog) {
+    private synchronized void handleUpdateFailedDialogReinstallAction(DialogInterface dialog) {
         if (mUpdateSetupFailedDialog == dialog) {
-
             /* Add a flag to the install url to indicate that the update setup failed, to show a help page. */
             String url = mInstallUrl;
             try {
@@ -1235,14 +1234,14 @@ public class Distribute extends AbstractMobileCenterService {
 
                 @Override
                 public void onClick(DialogInterface dialog, int which) {
-                    storeUpdateSetupFailedPackageHash(mUpdateSetupFailedDialog);
+                    storeUpdateSetupFailedPackageHash(dialog);
                 }
             });
             dialogBuilder.setNegativeButton(R.string.mobile_center_distribute_update_failed_dialog_reinstall, new DialogInterface.OnClickListener() {
 
                 @Override
                 public void onClick(DialogInterface dialog, int which) {
-                    handleUpdateFailedDialogReinstallAction(mUpdateSetupFailedDialog);
+                    handleUpdateFailedDialogReinstallAction(dialog);
                 }
             });
             mUpdateSetupFailedDialog = dialogBuilder.create();
--- a/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
+++ b/sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
@@ -76,6 +76,7 @@ import static com.microsoft.azure.mobile.distribute.DistributeConstants.LOG_TAG;
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.MEBIBYTE_IN_BYTES;
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.NOTIFICATION_CHANNEL_ID;
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.POSTPONE_TIME_THRESHOLD;
+import static com.microsoft.azure.mobile.distribute.DistributeConstants.EXTRA_UPDATE_SETUP_FAILED;
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.PREFERENCE_KEY_DISTRIBUTION_GROUP_ID;
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.PREFERENCE_KEY_DOWNLOAD_ID;
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.PREFERENCE_KEY_DOWNLOAD_STATE;
@@ -84,6 +85,8 @@ import static com.microsoft.azure.mobile.distribute.DistributeConstants.PREFEREN
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.PREFERENCE_KEY_RELEASE_DETAILS;
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.PREFERENCE_KEY_REQUEST_ID;
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.PREFERENCE_KEY_UPDATE_TOKEN;
+import static com.microsoft.azure.mobile.distribute.DistributeConstants.PREFERENCE_KEY_UPDATE_SETUP_FAILED_PACKAGE_HASH_KEY;
+import static com.microsoft.azure.mobile.distribute.DistributeConstants.PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY;
 import static com.microsoft.azure.mobile.distribute.DistributeConstants.SERVICE_NAME;
 import static com.microsoft.azure.mobile.distribute.DistributeUtils.computeReleaseHash;
 import static com.microsoft.azure.mobile.distribute.DistributeUtils.getStoredDownloadState;
@@ -190,6 +193,11 @@ public class Distribute extends AbstractMobileCenterService {
      * Mandatory download completed in app notification.
      */
     private Dialog mCompletedDownloadDialog;
+
+    /**
+     * Update setup failed dialog.
+     */
+    private Dialog mUpdateSetupFailedDialog;
 
     /**
      * Last activity that did show a dialog.
@@ -501,6 +509,7 @@ public class Distribute extends AbstractMobileCenterService {
         mUnknownSourcesDialog = null;
         mProgressDialog = null;
         mCompletedDownloadDialog = null;
+        mUpdateSetupFailedDialog = null;
         mLastActivityWithDialog.clear();
         mUsingDefaultUpdateDialog = null;
         mReleaseDetails = null;
@@ -542,6 +551,21 @@ public class Distribute extends AbstractMobileCenterService {
                 MobileCenterLog.info(LOG_TAG, "Not checking in app updates as installed from a store.");
                 mWorkflowCompleted = true;
                 return;
+            }
+
+            // If failed to enable in-app updates on the same app build before, don't go any further.
+            // Only if the app build is different (different package hash), try enabling in-app updates again.
+            String releaseHash = DistributeUtils.computeReleaseHash(this.mPackageInfo);
+            String updateSetupFailedPackageHash = PreferencesStorage.getString(PREFERENCE_KEY_UPDATE_SETUP_FAILED_PACKAGE_HASH_KEY);
+            if (updateSetupFailedPackageHash != null) {
+                if (releaseHash.equals(updateSetupFailedPackageHash)) {
+                    MobileCenterLog.info(LOG_TAG, "Skipping in-app updates setup, because it already failed on this release before.");
+                    return;
+                } else {
+                    MobileCenterLog.info(LOG_TAG, "Re-attempting in-app updates setup and cleaning up failure info from storage.");
+                    PreferencesStorage.remove(PREFERENCE_KEY_UPDATE_SETUP_FAILED_PACKAGE_HASH_KEY);
+                    PreferencesStorage.remove(PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY);
+                }
             }
 
             /* If we received the redirection parameters before Mobile Center was started/enabled, process them now. */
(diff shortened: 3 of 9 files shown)


sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DeepLinkActivity.java
```diff
@@ -25,14 +26,19 @@ public void onCreate(Bundle savedInstanceState) {
...
         MobileCenterLog.debug(LOG_TAG, getLocalClassName() + ".getIntent()=" + intent);
         MobileCenterLog.debug(LOG_TAG, "Intent requestId=" + requestId);
         MobileCenterLog.debug(LOG_TAG, "Intent distributionGroupId=" + distributionGroupId);
         MobileCenterLog.debug(LOG_TAG, "Intent updateToken passed=" + (updateToken != null));
+        MobileCenterLog.debug(LOG_TAG, "Intent updateSetupFailed passed=" + (updateSetupFailed != null));
 
         /* Store redirection parameters if both required values were passed. */
         if (requestId != null && distributionGroupId != null) {
             Distribute.getInstance().storeRedirectionParameters(requestId, distributionGroupId, updateToken);
+        } else if (updateSetupFailed != null){
```
max-mironov: Not sure if this is the only true condition for `updateSetupFailed` case. Is it possible that `distributionGroupId ` could be also `true` in this case? Please take a look at this.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -1130,6 +1173,51 @@ public void onClick(DialogInterface dialog, int which) {
...
+                    PreferencesStorage.putString(PREFERENCE_KEY_UPDATE_SETUP_FAILED_PACKAGE_HASH_KEY, DistributeUtils.computeReleaseHash(mPackageInfo));
+                }
+            });
+            dialogBuilder.setNegativeButton(R.string.mobile_center_distribute_update_failed_dialog_reinstall, new DialogInterface.OnClickListener() {
+
+                @Override
+                public void onClick(DialogInterface dialog, int which) {
+                    String url = mInstallUrl;
+                    // Add a flag to the install url to indicate that the update setup failed, to show a help page
+                    url += "?" + EXTRA_UPDATE_SETUP_FAILED + "=" + "true";
```
jaelim-ms: Can you define another constant for this one? `EXTRA_UPDATE_SETUP_FAILED` is the one for extra in intent.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -1130,6 +1173,51 @@ public void onClick(DialogInterface dialog, int which) {
...
+                public void onClick(DialogInterface dialog, int which) {
+                    PreferencesStorage.putString(PREFERENCE_KEY_UPDATE_SETUP_FAILED_PACKAGE_HASH_KEY, DistributeUtils.computeReleaseHash(mPackageInfo));
+                }
+            });
+            dialogBuilder.setNegativeButton(R.string.mobile_center_distribute_update_failed_dialog_reinstall, new DialogInterface.OnClickListener() {
+
+                @Override
+                public void onClick(DialogInterface dialog, int which) {
+                    String url = mInstallUrl;
+                    // Add a flag to the install url to indicate that the update setup failed, to show a help page
```
jaelim-ms: Add empty line before comment.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -728,6 +763,14 @@ synchronized void completeWorkflow() {
...
 
     /*
+     * Store update update setup failure message used later to show in setup failure dialog for user.
+     */
+    synchronized void storeUpdateSetupFailedParameter(@NonNull String updateSetupFailed) {
+        MobileCenterLog.debug(LOG_TAG, "Stored update setup failed parameter.");
+        PreferencesStorage.putString(PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY, updateSetupFailed);
+    }
+
+    /*
```
jaelim-ms: `/*` -> `/**`

jaelim-ms: Can you also change the line 765?

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DistributeConstants.java
```diff
@@ -201,6 +216,16 @@
...
 
+    /**
+     * Preference key for update setup failure package hash.
+     */
+    static final String PREFERENCE_KEY_UPDATE_SETUP_FAILED_PACKAGE_HASH_KEY = PREFERENCE_PREFIX + "update_setup_failed_package_hash";
+
+    /**
+     * Preference key for update setup failure error message.
+     */
+    static final String PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY = PREFERENCE_PREFIX + "update_setup_failed_message";
```
jaelim-ms: Do we really need this one? I saw the same implementation in iOS but it will be removed soon.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DeepLinkActivity.java
```diff
@@ -25,14 +26,19 @@ public void onCreate(Bundle savedInstanceState) {
...
+            // Otherwise just store error message to show update failure dialog in future.
```
jaelim-ms: There is a code style rule that
1. you need to use `/* */` instead of `//` for explanation of the code.
2. add empty line before comment.
3. etc...

max-mironov: Got it @jaelim-ms, thanks for feedback! I'll address this and other comments tomorrow and also will work on adding tests!

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DistributeConstants.java
max-mironov: I believe we need this to store/retrieve error message. What could be the alternative approach for storing the message? Should we use instance variable for this?

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DeepLinkActivity.java
MatkovIvan: Missing space before bracket

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -1130,6 +1173,51 @@ public void onClick(DialogInterface dialog, int which) {
     }
 
     /**
+     * Show update setupp failed dialog.
+     */
+    @UiThread
+    private synchronized void showUpdateSetupFailedDialog(final String errorMessage) {
+        /* Check if we need to replace dialog. */
+        if (!shouldRefreshDialog(mUpdateSetupFailedDialog)){
```
MatkovIvan: Missing space before bracket

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
MatkovIvan: Missing dot at end

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -1130,6 +1173,51 @@ public void onClick(DialogInterface dialog, int which) {
...
+            dialogBuilder.setNegativeButton(R.string.mobile_center_distribute_update_failed_dialog_reinstall, new DialogInterface.OnClickListener() {
+
+                @Override
+                public void onClick(DialogInterface dialog, int which) {
+                    String url = mInstallUrl;
+                    // Add a flag to the install url to indicate that the update setup failed, to show a help page
+                    url += "?" + EXTRA_UPDATE_SETUP_FAILED + "=" + "true";
+                    BrowserUtils.openBrowser(url, mForegroundActivity);
+
+                    // Clear the update setup failure info from storage, to re-attempt setup on reinstall
```
MatkovIvan: Missing dot at end

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -1130,6 +1173,51 @@ public void onClick(DialogInterface dialog, int which) {
...
+                    PreferencesStorage.remove(PREFERENCE_KEY_UPDATE_SETUP_FAILED_PACKAGE_HASH_KEY);
+                }
+            });
+            mUpdateSetupFailedDialog = dialogBuilder.create();
+            showAndRememberDialogActivity(mUpdateSetupFailedDialog);
+
+            // Don't show this dialog again
```
MatkovIvan: Missing dot at end

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
guperrot: and comment is always preceding any code block, you reversed the 2 lines here.

sdk/mobile-center-distribute/src/main/res/values/mobile_center_distribute.xml
```diff
@@ -15,4 +15,7 @@
     <string name="mobile_center_distribute_install_ready_message">%1$s %2$s (%3$d) has been downloaded and is ready to be installed.</string>
     <string name="mobile_center_distribute_install">Install</string>
     <string name="mobile_center_distribute_notification_category">App updates</string>
-</resources>
\ No newline at end of file
+    <string name="mobile_center_distribute_update_failed_dialog_title">In-app updates disabled</string>
```
guperrot: Were those texts and the UI reviewed by a PM? Were designers in loop?

pniko: @guperrot everyone is aware of the changes we are making to the SDK's. The experience should be close to exactly the same on both platforms.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
JessicaYeh: I don't know if it's 100% safe to assume that the URL we are building off of (`mInstallUrl`) will always not have other query parameters, so to be safe we should add some logic that adds either the `?` or `&` depending on whether there are already query parameters or not in `url`

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DistributeConstants.java
JessicaYeh: Yup, I just removed it on the iOS one, I saved it but never actually used it. The reason I didn't need to use the saved one was because I called the function to show the error dialog in the same scope as the function that parses the query parameters in the redirect URL, so it still had access to that message in a variable.

sdk/mobile-center-distribute/src/main/res/values/mobile_center_distribute.xml
JessicaYeh: yes, these strings are the same ones used in the iOS PR (https://github.com/Microsoft/mobile-center-sdk-ios/pull/773/files) which has been looked at by designers already

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DeepLinkActivity.java
JessicaYeh: this should be fine, if the in-app updates setup fails, there won't be any query parameters except for updateSetupFailed and requestId

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -643,6 +667,16 @@ else if (mUnknownSourcesDialog != null) {
                 }
             }
 
+            // If the in-app updates setup failed, and user ignores the failure, store the error
+            // message and also store the package hash that the failure occurred on. The setup
+            // will only be re-attempted the next time the app gets updated (and package hash changes).
+            String updateSetupFailedMessage = PreferencesStorage.getString(PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY);
+            if (updateSetupFailedMessage != null){
+                MobileCenterLog.debug(LOG_TAG, "In-app updates setup failure detected.");
+                showUpdateSetupFailedDialog(updateSetupFailedMessage);
```
JessicaYeh: should this block be moved into `DeepLinkActivity.java` into the area where `storeUpdateSetupFailedParameter` is being called, and then we no longer need to store the update setup failure message?

max-mironov: @JessicaYeh 
Initially I supposed to do so but the problem with calling this from `DeepLinkActivity` is that `mForegroundActivity` is `null` at that moment, so we can't build dialog at that time. 
That is also the reason I decided to store `errorMessage` with `PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY` in order to be able to retrieve it later to show error dialog with this message. 
Im open to any suggestions if you see how we can handle this otherwise.

@jaelim-ms  Could you also please take a look at this and advise?

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
max-mironov: That make sense, thanks for pointing at this, I'll fix it!

guperrot:
@max-mironov please fix the codecov tests now, you can click on the details to see what is missing or use local html report.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/BrowserUtils.java
```diff
@@ -101,4 +103,24 @@ static void openBrowser(@NonNull String url, @NonNull Activity activity) {
...
+        URI oldUri = new URI(uri);
+        String newQuery = oldUri.getQuery();
+        if (newQuery == null) {
+            newQuery = appendQuery;
+        } else {
+            newQuery += "&" + appendQuery;
+        }
+        URI newUri = new URI(oldUri.getScheme(), oldUri.getAuthority(),
+                oldUri.getPath(), newQuery, oldUri.getFragment());
+
```
guperrot: remove blank line if no comment

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/BrowserUtils.java
```diff
@@ -101,4 +103,24 @@ static void openBrowser(@NonNull String url, @NonNull Activity activity) {
             }
         }
     }
+
+    /**
+     * Append a query parameter to an existing URI.
+     *
+     * @param uri         initial uri.
+     * @param appendQuery parameter to append.
+     */
```
guperrot: @return parameter in Javadoc comments

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -544,6 +555,21 @@ private synchronized void resumeDistributeWorkflow() {
                 return;
             }
 
+            // If failed to enable in-app updates on the same app build before, don't go any further.
```
guperrot: Use /* comments

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -643,6 +669,16 @@ else if (mUnknownSourcesDialog != null) {
                 }
             }
 
+            // If the in-app updates setup failed, and user ignores the failure, store the error
```
guperrot: Use /* comment

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -544,6 +555,21 @@ private synchronized void resumeDistributeWorkflow() {
...
+            // Only if the app build is different (different package hash), try enabling in-app updates again.
+            String releaseHash = DistributeUtils.computeReleaseHash(this.mPackageInfo);
+            String updateSetupFailedPackageHash = PreferencesStorage.getString(PREFERENCE_KEY_UPDATE_SETUP_FAILED_PACKAGE_HASH_KEY);
+            if (updateSetupFailedPackageHash != null) {
+                if (releaseHash.equals(updateSetupFailedPackageHash)) {
+                    MobileCenterLog.info(LOG_TAG, "Skipping in-app updates setup, because it already failed on this release before.");
+                    return;
+                } else {
+                    MobileCenterLog.info(LOG_TAG, "Re-attempting in-app updates setup and cleaning up failure info from storage.");
+                    PreferencesStorage.remove(PREFERENCE_KEY_UPDATE_SETUP_FAILED_PACKAGE_HASH_KEY);
```
guperrot: When SDK is disabled you need to remove those new keys as well (in the method where everything is cleared).

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DeepLinkActivity.java
```diff
@@ -25,14 +26,20 @@ public void onCreate(Bundle savedInstanceState) {
...
         MobileCenterLog.debug(LOG_TAG, "Intent updateToken passed=" + (updateToken != null));
+        MobileCenterLog.debug(LOG_TAG, "Intent updateSetupFailed passed=" + (updateSetupFailed != null));
 
         /* Store redirection parameters if both required values were passed. */
         if (requestId != null && distributionGroupId != null) {
             Distribute.getInstance().storeRedirectionParameters(requestId, distributionGroupId, updateToken);
+        } else if (updateSetupFailed != null) {
+
+            /* Otherwise just store error message to show update failure dialog in future. */
+            Distribute.getInstance().storeUpdateSetupFailedParameter(updateSetupFailed);
```
guperrot: You need to check requestId like in `storeRedirectionParameters`

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -1130,6 +1175,57 @@ public void onClick(DialogInterface dialog, int which) {
...
+     */
+    @UiThread
+    private synchronized void showUpdateSetupFailedDialog(final String errorMessage) {
+
+        /* Check if we need to replace dialog. */
+        if (!shouldRefreshDialog(mUpdateSetupFailedDialog)) {
+            return;
+        }
+        MobileCenterLog.debug(LOG_TAG, "Show update setup failed dialog.");
+
```
guperrot: Either add a comment or remove blank line

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -1130,6 +1175,57 @@ public void onClick(DialogInterface dialog, int which) {
...
+        if (mForegroundActivity != null) {
+            AlertDialog.Builder dialogBuilder = new AlertDialog.Builder(mForegroundActivity);
+            dialogBuilder.setCancelable(false);
+            dialogBuilder.setTitle(R.string.mobile_center_distribute_update_failed_dialog_title);
+            dialogBuilder.setMessage(errorMessage);
+            dialogBuilder.setPositiveButton(R.string.mobile_center_distribute_update_failed_dialog_ignore, new DialogInterface.OnClickListener() {
+
+                @Override
+                public void onClick(DialogInterface dialog, int which) {
+                    PreferencesStorage.putString(PREFERENCE_KEY_UPDATE_SETUP_FAILED_PACKAGE_HASH_KEY, DistributeUtils.computeReleaseHash(mPackageInfo));
```
guperrot: State can change between a dialog is shown and the click or cancel, please follow the same code pattern with state management as in every other dialog.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -1130,6 +1175,57 @@ public void onClick(DialogInterface dialog, int which) {
...
+                }
+            });
+            dialogBuilder.setNegativeButton(R.string.mobile_center_distribute_update_failed_dialog_reinstall, new DialogInterface.OnClickListener() {
+
+                @Override
+                public void onClick(DialogInterface dialog, int which) {
+
+                    /* Add a flag to the install url to indicate that the update setup failed, to show a help page. */
```
guperrot: Same comment about state change

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DistributeConstants.java
guperrot: I don't think this comment applies to the Android implementation, you have to persist the message to avoid losing it after kill/restart.

sdk/mobile-center-distribute/src/main/res/values/mobile_center_distribute.xml
guperrot: Thanks

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DistributeConstants.java
jaelim-ms: Do we care of losing the message after kill/restart in this case? This step is very beginning and we can simply start over the process. It is not kinda handling download or update.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -1119,6 +1127,39 @@ private void viewReleaseNotes(ReleaseDetails releaseDetails) {
     }
 
     /**
+     * Store failed package info hash in preferences.
+     */
+    private synchronized void storeUpdateSetupFailedPackageHash() {
+        if (mPackageInfo != null && mUpdateSetupFailedDialog != null) {
```
guperrot: If mPackageInfo is null, that means the SDK is not started, that does not mean the SDK is disabled.
Please look at logic around `mBeforeStartUpdateToken` that addresses this scenario.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -1119,6 +1127,39 @@ private void viewReleaseNotes(ReleaseDetails releaseDetails) {
...
+        } else {
+            showDisabledToast();
+        }
+    }
+
+    /**
+     * Redirect user to install url page in browser and clear current failed package hash from preferences.
+     */
+    private synchronized void handleUpdateFailedDialogReinstallAction() {
+        if (mUpdateSetupFailedDialog != null) {
```
guperrot: prefer passing dialog as parameter and test against that instead of null like in other places

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -1119,6 +1127,39 @@ private void viewReleaseNotes(ReleaseDetails releaseDetails) {
     }
 
     /**
+     * Store failed package info hash in preferences.
+     */
+    private synchronized void storeUpdateSetupFailedPackageHash() {
```
guperrot: pass dialog as parameter and check against that instead of null

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/DistributeConstants.java
guperrot: We care about not opening browser twice, that would defeat the purpose I think.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
max-mironov: Got it, thanks for pointing

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
```diff
@@ -1140,9 +1140,8 @@ private synchronized void storeUpdateSetupFailedPackageHash(Dialog dialog) {
     /**
      * Redirect user to install url page in browser and clear current failed package hash from preferences.
      */
-    private synchronized void handleUpdateFailedDialogReinstallAction(Dialog dialog) {
+    private synchronized void handleUpdateFailedDialogReinstallAction(DialogInterface dialog) {
         if (mUpdateSetupFailedDialog == dialog) {
-
```
MatkovIvan: Revert this

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeBeforeApiSuccessTest.java
```diff
@@ -23,6 +24,7 @@
 
 import org.json.JSONException;
 import org.junit.Test;
+import org.junit.runner.RunWith;
```
MatkovIvan: and this

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeBeforeApiSuccessTest.java
```diff
@@ -32,6 +34,7 @@
 import org.mockito.verification.VerificationMode;
 import org.powermock.api.mockito.PowerMockito;
 import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
```
MatkovIvan: and this

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
max-mironov: yep, thanks!

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
guperrot: This comment is still active

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeBeforeApiSuccessTest.java
```diff
@@ -251,6 +320,108 @@ public void accept(Void aVoid) {
...
+        when(BrowserUtils.appendUri(anyString(), anyString())).thenThrow(new URISyntaxException("ex", "message"));
+
+        /* Click. */
+        clickListener.getValue().onClick(mDialog, DialogInterface.BUTTON_NEGATIVE);
+        verifyStatic();
+        PreferencesStorage.remove(PREFERENCE_KEY_UPDATE_SETUP_FAILED_PACKAGE_HASH_KEY);
+    }
+
+    @Test
+    @PrepareForTest(DistributeUtils.class)
```
guperrot: Are we using the real implementation of DistributeUtils in other methods of this file? If no we could move it the PrepareForTest to class level.

sdk/mobile-center-distribute/src/main/java/com/microsoft/azure/mobile/distribute/Distribute.java
guperrot: To be more accurate in `applyEnabledState(false)` : remove the new keys as well in the else.

sdk/mobile-center-distribute/src/test/java/com/microsoft/azure/mobile/distribute/DistributeBeforeApiSuccessTest.java
max-mironov: I like this idea, thanks for pointing!

PR #534, opened 2017-10-25
Fix project for Android Studio 3.0

guperrot:


sdk/build.gradle
```diff
@@ -6,7 +6,7 @@ allprojects {
     apply plugin: 'jacoco'
 
     jacoco {
-        toolVersion '0.7.9'
+        toolVersion '0.7.4.201502262128'
```
MatkovIvan: Why are you using exactly this version?

guperrot: Because it does not work anymore otherwise "Unable to read execution file" which is a binary version problem for reports.

I then executed ./gradlew androidDependencies to find out which version Andoid is using for coverage debug on the real device and it was that version.

jaelim-ms: If this fixes the issue, we can downgrade to the lower version. We can take it separately and fix it later.

versions.gradle
```diff
@@ -6,6 +6,5 @@ ext {
     minSdkVersion = 15
     targetSdkVersion = 26
     compileSdkVersion = 26
-    buildToolsVersion = '26.0.1'
```
jaelim-ms: Just curious, what's the change for this one? We don't need to specify build tools version anymore?

guperrot: Yes, it even triggers a linter if version mismatches with what android uses now.

PR #549, opened 2017-11-18
Feature/hosted script

guperrot:


scripts/publish-github-release.sh
```diff
@@ -73,25 +73,25 @@ else
 fi
 
 # 4. Copy binaries from :sdk
```
jaelim-ms: We can probably improve this step later.
I think we don't need to move this to that specific directory anymore, there is a copy task in VSTS.

scripts/publish-storage-account.sh
```diff
@@ -11,14 +13,16 @@ FILES="$(find sdk -name $BINARY_FILE_FILTER)"
...
-cd $BITRISE_DEPLOY_DIR
+cd $BUILD_ARTIFACTSTAGINGDIRECTORY
 mkdir $ARCHIVE
 cp $BINARY_FILE_FILTER $ARCHIVE
 zip -r $ZIP_FILE $ARCHIVE
+cd -
 
 # Upload file
-$1azure storage blob upload $ZIP_FILE sdk
+AZURE_STORAGE_ACCESS_KEY=$AZURE_STORAGE_ACCESS_KEY \
```
jaelim-ms: What is this line?

scripts/run-code-coverage.sh
```diff
@@ -0,0 +1,21 @@
+# We can't run emulator as a daemon
```
jaelim-ms: It is really bad that VSTS doesn't have an option for emulator. It was in Android Build task but it is deprecated. 😞

scripts/publish-storage-account.sh
guperrot: expose environment variable just during that call

scripts/publish-github-release.sh
guperrot: Yeah it actually works already without it, but this is nice to have when running locally

scripts/publish-storage-account.sh
jaelim-ms: Oh.. OK. Probably I need this in Apple SDK build. Currently it exports the environment variables.

PR #553, opened 2017-11-23
Fix/avoid crash big attachment

guperrot:


sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -195,13 +201,37 @@ public String getLogs(@NonNull String group, @IntRange(from = 0) int limit, @Non
             ContentValues values = iterator.next();
             Long dbIdentifier = values.getAsLong(DatabaseManager.PRIMARY_KEY);
 
+            /*
+             * When we can't even read the identifier (in this case ContentValues is most likely empty).
+             * That probably means it contained a record larger than 5MB and we hit the cursor limit.
+             * Get rid of first non pending log.
+             */
+            if (dbIdentifier == null) {
+                AppCenterLog.error(LOG_TAG, "Empty database record, probably entry larger than 5MB, need to delete as it's now corrupted");
```
MatkovIvan: This happens with 2.75 MB file for me, so I'm not sure about this value

guperrot: We can change the log message.

jaelim-ms: "entry larger than .." -> "an entry is larger than .."

PR #555, opened 2017-11-28
Fix some lint warnings

guperrot:


sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/crypto/CryptoAesHandler.java
```diff
@@ -24,7 +23,7 @@
 
     @Override
     public String getAlgorithm() {
-        return CIPHER_AES + "/" + AES_KEY_SIZE;
+        return CryptoConstants.CIPHER_AES + "/" + AES_KEY_SIZE;
```
jaelim-ms: Any reasons?

guperrot: Yes because the required version annotation at class level was not applied to static import

guperrot: and thus triggering the lint version check error

jaelim-ms: OK

PR #558, opened 2017-11-30
Add text for Distribute UI

jaelim-ms:
The localized strings are not from TouchDown. We need to localize again.

### DIFF
--- a/sdk/appcenter-distribute/src/main/res/values-cs/appcenter_distribute.xml
+++ b/sdk/appcenter-distribute/src/main/res/values-cs/appcenter_distribute.xml
@@ -18,4 +18,5 @@
   <string name="appcenter_distribute_update_failed_dialog_title">Neaktivní aktualizace v aplikacích</string>
   <string name="appcenter_distribute_update_failed_dialog_ignore">Ignorovat</string>
   <string name="appcenter_distribute_update_failed_dialog_reinstall">Znovu nainstalujte aplikaci</string>
+  <string name="appcenter_distribute_update_failed_dialog_message">Tato verze byla buď strana-naložený nebo stáhnout pomocí prohlížeče v soukromém režimu.</string>
 </resources>
--- a/sdk/appcenter-distribute/src/main/res/values-cs/appcenter_distribute.xml
+++ b/sdk/appcenter-distribute/src/main/res/values-cs/appcenter_distribute.xml
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-  <string name="appcenter_distribute_update_dialog_title">K dispozici je aktualizace aplikace.</string>
+  <string name="appcenter_distribute_update_dialog_title">K dispozici je aktualizace aplikace</string>
   <string name="appcenter_distribute_update_dialog_message_optional">%1$s %2$s (%3$d) je k dispozici ke stažení a instalaci.</string>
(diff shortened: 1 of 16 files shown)


sdk/appcenter-distribute/src/main/res/values-cs/appcenter_distribute.xml
```diff
@@ -1,21 +1,22 @@
...
   <string name="appcenter_distribute_update_dialog_view_release_notes">Zobrazit zprávu k vydání verze</string>
-  <string name="appcenter_distribute_dialog_actioned_on_disabled_toast">Distribuce byla zakázána.</string>
+  <string name="appcenter_distribute_dialog_actioned_on_disabled_toast">Distribuce byla zakázána</string>
   <string name="appcenter_distribute_unknown_sources_dialog_message">Z bezpečnostních důvodů je vaše zařízení nastaveno tak, aby blokovalo instalaci aplikací získaných z neznámých zdrojů.</string>
   <string name="appcenter_distribute_unknown_sources_dialog_settings">Nastavení</string>
-  <string name="appcenter_distribute_downloading_mandatory_update">Stahuje se aktualizace.</string>
+  <string name="appcenter_distribute_downloading_mandatory_update">Stahuje se aktualizace</string>
   <string name="appcenter_distribute_download_progress_number_format">%1$d MB z %2$d MB</string>
-  <string name="appcenter_distribute_install_ready_title">Aktualizace aplikace se stáhla.</string>
+  <string name="appcenter_distribute_install_ready_title">Aktualizace aplikace se stáhla</string>
```
guperrot: I don't understand why/how you localized if not from touch down, that should be a separate step.

jaelim-ms: As mentioned, it should be done after this one. We need to run TouchDown for this string.

guperrot: That does not answer my question ;) Why did you localize yourself then?

jaelim-ms: Well, just in case if we forgot getting localized text when we release newer version.

guperrot: But how can we make sure these translations are suitable? Where do they come from? Shouldn't we always commit english only and then ask TD in a separate step?

sdk/appcenter-distribute/src/main/res/values-es/appcenter_distribute.xml
```diff
@@ -12,10 +12,11 @@
   <string name="appcenter_distribute_downloading_mandatory_update">Descargando la actualización</string>
   <string name="appcenter_distribute_download_progress_number_format">%1$d MB de %2$d MB</string>
   <string name="appcenter_distribute_install_ready_title">Actualización de la aplicación descargada</string>
-  <string name="appcenter_distribute_install_ready_message">%1$s %2$s (%3$d) se ha descargado y está listo para instalarlo</string>
+  <string name="appcenter_distribute_install_ready_message">%1$s %2$s (%3$d) se ha descargado y está listo para instalarlo.</string>
```
guperrot: this is used for a system notification, notification messages usually end with dots, I just checked my phone with a system notification.

sdk/appcenter-distribute/src/main/res/values-cs/appcenter_distribute.xml
jaelim-ms: As told offline, I had extra time to do this. 😄 Will use English for future changes.

PR #568, opened 2017-12-06
Fix some warnings, remove checkstyle, update android build tools

guperrot:
We never ever used checkstyle and it uses deprecated gradle field
making a huge warning on logs.

Also reformat some files and cosmetic changes.


guperrot:
More fixes in later prs.

PR #571, opened 2017-12-06
Breakpad NDK Crash Support

pniko:
This PR adds support for crashes generated from JNI functions in NDK/breakpad projects. This enables the SDK to capture native crashes from JNI functions and send the binary dumps back to App Center for processing.

Changes:

-Added google_breakpad folder with dependencies so project can be built without any additional configuration steps.
-Updated gradle configuration to include building breakpad and Sasquatch sample c++ crash handler library
-Added 3 NDK crash scenarios - divide by zero, stack overflow, out of memory
-Added sample minidump crash example to Sasquatch inside of main.cpp file which outputs dumps to standard /error/breakpad folder.
-Added logic to pick up all minidumps created under breakpad and send them to the App Center API as an attachment to a crash for backend symbolication

Any questions - ask me and @JessicaYeh


pniko:
@guperrot Going to move the breakpad dependency to a sub-module to simplify this PR down to the basics then we'll check coverage and get those issues addressed today.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
```diff
@@ -38,10 +39,16 @@
 import com.microsoft.appcenter.utils.AppCenterLog;
 import com.microsoft.appcenter.utils.async.AppCenterConsumer;
 
+import java.io.File;
 import java.util.UUID;
 
 public class MainActivity extends AppCompatActivity {
 
+    // Used to load the 'native-lib' library on application startup.
```
JessicaYeh: this comment needs to be updated (was copied from sample)

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -555,6 +560,33 @@ private void processPendingErrors() {
...
+
+                    /* Create our Breakpad dump attachment */
+                    ErrorAttachmentLog breakpadAttachment = ErrorAttachmentLog.attachmentWithBinary(logfileContents, "minidump.dmp", "application/octet-stream");
+                    List<ErrorAttachmentLog> list = new LinkedList<>();
+                    list.add(breakpadAttachment);
+
+                    /* Attach dump to NDK managed exception */
+                    ManagedErrorLog errorLog = ErrorLogHelper.createErrorLog(mContext, Thread.currentThread(), new NativeException(), Thread.getAllStackTraces(), mInitializeTimestamp, true);
+                    if(errorLog != null) {
+                        errorLog.getException().setWrapperSdkName("appcenter.ndk");
```
JessicaYeh: i think there was a bad merge, i added a line here and it got removed. should also have

```
errorLog.getDevice().setWrapperSdkName("appcenter.ndk");
```

pniko: @JessicaYeh I accidentally removed this when refactoring for the unit tests. Put this back in.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelper.java
```diff
@@ -156,10 +181,23 @@ public boolean accept(File dir, String filename) {
                 return filename.endsWith(ERROR_LOG_FILE_EXTENSION);
             }
         });
+        return files != null && files.length > 0 ? files : new File[0];
```
MatkovIvan: If we already have a zero-length array, it can be returned as is

sdk/appcenter/src/test/java/com/microsoft/appcenter/StorageHelperTest.java
```diff
@@ -37,12 +38,13 @@
...
 
+@RunWith(PowerMockRunner.class)
 @SuppressWarnings("unused")
 @PrepareForTest({StorageHelper.InternalStorage.class, AppCenterLog.class, TextUtils.class})
 public class StorageHelperTest {
 
-    @Rule
-    public PowerMockRule rule = new PowerMockRule();
+    //@Rule
+    //public PowerMockRule rule = new PowerMockRule();
```
MatkovIvan: Revert this

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
```diff
@@ -121,6 +129,17 @@ protected void onCreate(Bundle savedInstanceState) {
         /* Start App Center. */
         AppCenter.start(getApplication(), sSharedPreferences.getString(APP_SECRET_KEY, getString(R.string.app_secret)), Analytics.class, Crashes.class, Distribute.class, Push.class);
 
+        /* Attach NDK Crash Handler (if available) after SDK is initialized */
```
MatkovIvan: Add dots at the end of all comments.

apps/sasquatch/build.gradle
```diff
@@ -40,6 +49,11 @@ android {
             matchingFallbacks = ['release']
         }
     }
+    externalNativeBuild {
```
guperrot: blank line

sdk/appcenter-crashes/src/androidTest/java/com/microsoft/appcenter/crashes/CrashesAndroidTest.java
```diff
@@ -58,6 +59,14 @@
 
     private Channel mChannel;
 
+    /* Filter out the breakpad folder. */
+    FilenameFilter breakpadFilter = new FilenameFilter() {
```
guperrot: If it can be private, it has to.
Also member variables in Android are following `mName` convention except for inner classes sometimes.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -555,6 +560,34 @@ private void processPendingErrors() {
 
     }
 
+    private void processPendingBreakpadErrors() {
+        post(new Runnable() {
+
+            @Override
+            public void run() {
+                for (File logFile : ErrorLogHelper.getStoredBreakpadLogFiles()) {
```
guperrot: This completely bypasses the listener system and people will not have any callbacks and ability to send a text attachment or monitor sending events. We need to integrate this in the same process as other crashes, just a different source of crash but keep all events and behaviors.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -555,6 +560,34 @@ private void processPendingErrors() {
...
+
+                    /* Create our Breakpad dump attachment. */
+                    ErrorAttachmentLog breakpadAttachment = ErrorAttachmentLog.attachmentWithBinary(logfileContents, "minidump.dmp", "application/octet-stream");
+                    List<ErrorAttachmentLog> list = new LinkedList<>();
+                    list.add(breakpadAttachment);
+
+                    /* Attach dump to NDK managed exception. */
+                    ManagedErrorLog errorLog = ErrorLogHelper.createErrorLog(mContext, Thread.currentThread(), new NativeException(), Thread.getAllStackTraces(), mInitializeTimestamp, true);
+                    if(errorLog != null) {
+                        errorLog.getException().setWrapperSdkName("appcenter.ndk");
```
guperrot: make a constant for that

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/model/NativeException.java
```diff
@@ -0,0 +1,10 @@
+package com.microsoft.appcenter.crashes.model;
+
+public class NativeException extends RuntimeException {
```
guperrot: does not seem native exception to me, can you explain?

sdk/appcenter/src/main/java/com/microsoft/appcenter/Constants.java
```diff
@@ -24,6 +24,10 @@
      * Maximum number of requests being sent for the group.
      */
     static final int DEFAULT_TRIGGER_MAX_PARALLEL_REQUESTS = 3;
+    /**
```
guperrot: blank line missing before an after

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
guperrot: Also I don't see modification to make hasCrashedInLastSesion/getLastSessionCrashReport work with native crash.

sdk/appcenter/src/main/java/com/microsoft/appcenter/Constants.java
pniko: I followed the pattern you guys established in this file. Why don't any of the constants above or below have a blank line before and after?

sdk/appcenter-crashes/src/androidTest/java/com/microsoft/appcenter/crashes/CrashesAndroidTest.java
pniko: Done.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/model/NativeException.java
pniko: It only represents the NDK exception, but isn't an actual one. This is only used after a breakpad minidump has been generated.

sdk/appcenter/src/main/java/com/microsoft/appcenter/Constants.java
guperrot: sometimes android studio will collapse blank lines by mistake when reordering, I guess it was missed during a prior review

pniko: Updated this along with the other constants in this file to have lines above and below.

apps/sasquatch/build.gradle
pniko: Done

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
pniko: Done

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
pniko: I believe events and getting hasCrashedInLastSession were stretch goals outside the scope of this initial PR.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
```diff
@@ -121,6 +129,16 @@ protected void onCreate(Bundle savedInstanceState) {
         /* Start App Center. */
         AppCenter.start(getApplication(), sSharedPreferences.getString(APP_SECRET_KEY, getString(R.string.app_secret)), Analytics.class, Crashes.class, Distribute.class, Push.class);
 
+        /* Attach NDK Crash Handler (if available) after SDK is initialized. */
```
guperrot: Is that going to be the documented steps? If yes we need to make a method for that to simplify.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
guperrot: Bu how will this look documentation wise?

sdk/appcenter/src/main/java/com/microsoft/appcenter/Constants.java
```diff
@@ -16,18 +16,32 @@
...
+
+    /**
+     * Directory under the FILES_PATH containing breakpad minidumps.
+     */
+    public static final String BREAKPAD_DIRECTORY = "breakpad";
+
+    /**
+     * Directory under the FILES_PATH containing breakpad minidumps.
+     */
+    public static final String BREAKPAD_WRAPPER_NAME = "appcenter.ndk";
```
guperrot: I saw this constant used in crashes as well. Better use only 1 constant (this one). It could be renamed WRAPPER_SDK_NAME_NDK also.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
achocron: @pniko Could you just create the managed error log and then save it in the folder with the others before the normal processing takes place? You could even send the error attachments in the preprocessing steps since they're different logs and you'll already have the error id

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
pniko: This should only provide them with an example. I don't believe there is going to be any gain from refactoring this into a function when it's really just 1 line of code for most users. This will be explained in documentation so users know, but it's one of the first steps in the breakpad document that most developers will have to read in order to figure out how to incorporate breakpad in their make process if they haven't already for HockeyApp.

guperrot: I still don't understand why customers would access classes not designed for public audience like ErrorLogHelper, we don't want to expose that class.

Every access to ErrorLogHelper must be wrapped into a public static method of Crashes.

Like Crashes.setUpNativeCrashes. Customers must not have to use our internals.

pniko: I agree that would be much cleaner. Moved this to a Crashes function and updated the MainActivity to reference that instead.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -499,9 +528,50 @@ private void initialize() {
...
+                modelException.setType("minidump");
+                modelException.setWrapperSdkName(Constants.WRAPPER_SDK_NAME_NDK);
+                modelException.setStackTrace(dest.getPath());
+                ManagedErrorLog errorLog = new ManagedErrorLog();
+                errorLog.setException(modelException);
+                errorLog.setTimestamp(new Date(logFile.lastModified()));
+                errorLog.setFatal(true);
+                errorLog.setId(UUID.randomUUID());
+
+                /* TODO The following properties are wrong. */
```
jwallra: Is this something that needs to be fixed before merging, then?

guperrot: Not really

apps/sasquatch/src/main/cpp/main.cpp
```diff
@@ -0,0 +1,91 @@
...
+        return succeeded;
+    }
+
+    /**
+     * Registers breakpad as the exception handler for NDK code.
+     *
+     * @param path returned from Crashes.getBreakpadDirectory()
+     */
+    void Java_com_microsoft_appcenter_sasquatch_activities_MainActivity_setupNativeCrashesListener(JNIEnv *env, jobject, jstring path) {
+        const char* dump_path = (char *)env->GetStringUTFChars(path, NULL);
```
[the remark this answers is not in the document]
guperrot: it's not formatted yet, I tested Android Studio has a reformat action and needs to be run on this file.

I was waiting on @pniko patch on OOM crashes that currently don't crash to avoid conflicts (hence the blocked task on our board).

apps/sasquatch/src/main/cpp/main.cpp
```diff
@@ -0,0 +1,91 @@
...
+    jint JNI_OnLoad(JavaVM *vm, void * /*reserved*/) {
+
+        __android_log_print(ANDROID_LOG_INFO, "breakpad", "JNI onLoad...");
+
+        return JNI_VERSION_1_4;
+    }
+
+    void Java_com_microsoft_appcenter_sasquatch_activities_CrashActivity_nativeDivideByZeroCrash(JNIEnv* env, jobject obj) {
+        volatile int *a = reinterpret_cast<volatile int *>(NULL);
+        *a = 1;
```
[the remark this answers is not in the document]
guperrot: @pniko it looks like it's another type of crash indeed.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
```diff
@@ -28,7 +28,6 @@
     private boolean mCrashSuperPauseNotCalled;
 
     private boolean mCrashSuperDestroyNotCalled;
-
```
jaelim-ms: Let's bring this empty back with later change

PR #575, opened 2017-12-07
Allow saving Java throwable along with wrapper crash

guperrot:


guperrot:
CI failed because it failed to download a jar file.

PR #581, opened 2017-12-11
Invalidate session id when analytics is disabled

achocron:


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -453,6 +454,7 @@ protected synchronized void applyEnabledState(boolean enabled) {
 
         /* Release resources if disabled and enabled before with resources. */
         else {
+            SessionIdContext.getInstance().invalidateSessionId();
```
guperrot: Please move to clearSessions and maybe remove plural or even rename that one.
Please also clean code that adds the session history.

PR #585, opened 2017-12-13
Fix large attachments

guperrot:


sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/Persistence.java
```diff
@@ -32,7 +32,7 @@
      * @param log   The log to be placed in the storage.
      * @throws PersistenceException Exception will be thrown if Persistence cannot write a log to the storage.
```
achocron: What is the return value? can you add a returns tag?

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -55,6 +58,22 @@
      */
     private static final ContentValues SCHEMA = getContentValues("", "");
 
+    /**
+     * Size limit for a database row log payload.
+     * A separate file is used if payload is larger.
+     */
+    private static final int PAYLOAD_MAX_SIZE = (int) (1.9 * 1024 * 1024);
```
achocron: I know it's possible, but is it efficient to have 2 megabyte rows in the database?

achocron: Also can you add to the comment that the size is in megabytes?

guperrot: I don't see any reason to lower the limit, plus Apple implementation will keep using SQLite for any size. Will update the comment.

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/Persistence.java
guperrot: Needed for new unit tests. Will add Javadoc comment.

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
achocron: I was just curious because if we had rows that contained data in excess of 2 megabytes, certain queries could take much longer. but I don't know if that's true or affects the sdk.

guperrot: Will also fix size check based on bytes and not String chars.

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -146,10 +205,13 @@ public void deleteLogs(@NonNull String group, @NonNull String id) {
         AppCenterLog.debug(LOG_TAG, "The IDs for deleting log(s) is/are:");
 
         List<Long> dbIdentifiers = mPendingDbIdentifiersGroups.remove(group + id);
+        File directory = getLargePayloadGroupDirectory(group);
         if (dbIdentifiers != null) {
             for (Long dbIdentifier : dbIdentifiers) {
                 AppCenterLog.debug(LOG_TAG, "\t" + dbIdentifier);
                 mDatabaseStorage.delete(dbIdentifier);
+                //noinspection ResultOfMethodCallIgnored we are not checking SQLite delete either.
+                getLargePayloadFile(directory, dbIdentifier).delete();
```
achocron: Could you make the identifier deleted after the file? If the identifier is deleted but something goes wrong while or before deleting the large file, there will be a large file leak.

guperrot: Was thinking about that this morning :D Will do.

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/Persistence.java
```diff
@@ -30,9 +30,10 @@
      *
      * @param group The group of the storage for the log.
      * @param log   The log to be placed in the storage.
+     * @return log identifier from persistence after saving.
```
jaelim-ms: `l` -> `L`og identifier

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -109,6 +133,9 @@ public void onError(String operation, RuntimeException e) {
                         AppCenterLog.error(LOG_TAG, "Cannot complete an operation (" + operation + ")", e);
                     }
                 });
+        mLargePayloadDirectory = new File(Constants.FILES_PATH + PAYLOAD_LARGE_DIRECTORY);
+        //noinspection ResultOfMethodCallIgnored we handle errors at read/write time for each file.
```
jaelim-ms: Didn't we have a line break before the line?

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -126,18 +153,56 @@ private static ContentValues getContentValues(@Nullable String group, @Nullable
...
+                directory.mkdir();
+                File payloadFile = getLargePayloadFile(directory, databaseId);
+                StorageHelper.InternalStorage.write(payloadFile, payload);
+                AppCenterLog.debug(LOG_TAG, "Payload written to " + payloadFile);
+            }
+            return databaseId;
         } catch (JSONException e) {
             throw new PersistenceException("Cannot convert to JSON string", e);
+        } catch (IOException e) {
+            throw new PersistenceException("Cannot save large payload in a file", e);
```
jaelim-ms: Shouldn't we delete the database entry if it fails?

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -161,6 +227,16 @@ public void deleteLogs(String group) {
         /* Log. */
         AppCenterLog.debug(LOG_TAG, "Deleting all logs from the Persistence database for " + group);
 
+        /* Delete large payload files */
```
jaelim-ms: Not deleting the directory? Is it intentional?

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -55,6 +58,22 @@
      */
     private static final ContentValues SCHEMA = getContentValues("", "");
 
+    /**
+     * Size limit (in bytes) for a database row log payload.
+     * A separate file is used if payload is larger.
+     */
+    private static final int PAYLOAD_MAX_SIZE = (int) (1.9 * 1024 * 1024);
```
jaelim-ms: Can we use `com.android.internal.R.integer.config_cursorWindowSize`?

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -228,7 +306,19 @@ public String getLogs(@NonNull String group, @IntRange(from = 0) int limit, @Non
...
+                    String logPayload;
+                    String databasePayload = values.getAsString(COLUMN_LOG);
+                    if (databasePayload == null) {
+                        File file = getLargePayloadFile(largePayloadGroupDirectory, dbIdentifier);
+                        AppCenterLog.debug(LOG_TAG, "Read payload file " + file);
+                        logPayload = StorageHelper.InternalStorage.read(file);
+                    } else {
+                        logPayload = databasePayload;
+                    }
+                    if (logPayload == null) {
```
jaelim-ms: Move this if statement in above if statement (after line 314). We don't need to even check this if payload is not big.

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
guperrot: It will be deleted at read record time, simpler that way.

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
guperrot: We could but it didn't feel useful as it will be re-created when enabling again. Could add it if you think it's useful.

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
guperrot: This is to make sure that if `StorageHelper.InternalStorage.read(file);` has an error, it will be managed. `StorageHelper.InternalStorage.read(file);` does not throw exceptions on error, it returns null.

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
guperrot: We should never use anything that has `android.internal` it could break

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
jaelim-ms: It was just a question. I wondered if you missed it.

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
jaelim-ms: I meant if `databasePayload != null` then you don't need this if statement.

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
guperrot: adding

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
guperrot: adding

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
jaelim-ms: I agree, it is much simpler. But if the maximum number of logs are already stored, we will lose valid logs because of this.

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
guperrot: ok

PR #589, opened 2017-12-20
Disable services if AppCenter is disabled 

MatkovIvan:
In a case when `AppCenter` is disabled, but some service is enabled, `applyEnabledState(true)` was called.
It's causing some issues, for example, rum sending network requests while `AppCenter` is disabled.


guperrot:
I guess that's an edge case where you remove a module from start line and restore it later.

MatkovIvan:
Yeah, but remove step isn't necessary.
So, the repro steps:
- Disable `AppCenter`. It Disables core and all current modules.
- Add a new module for next run. The new module will be enabled by default.

guperrot:
@MatkovIvan cannot reproduce via the edited comment, but original one is right. Anyway that does not change the fix.

PR #590, opened 2017-12-20
Sending start services log after enable

MatkovIvan:


guperrot:
It is actually intended to send start service log even when disabled. How does iOS or UWP behave?

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -115,6 +115,11 @@
      */
     private Set<AppCenterService> mServices;
 
+    /**
+     * Started services for which the log isn't sent yet.
+     */
+    private List<String> mStartedServicesNamesToLog;
```
guperrot: You can use Collection

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -655,6 +672,12 @@ private void setInstanceEnabled(boolean enabled) {
             StorageHelper.PreferencesStorage.putBoolean(PrefStorageConstants.KEY_ENABLED, true);
         }
 
+        /* Send started services. */
+        if (mStartedServicesNamesToLog != null && switchToEnabled) {
```
guperrot: Looks like 1 unit test is missing for code coverage

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -566,12 +571,24 @@ private void finishStartServices(Iterable<AppCenterService> services) {
...
+        sendStartServiceLog(serviceNames);
+    }
 
-        /* Queue start service log. */
+    /**
+     * Queue start service log.
+     *
+     * @param serviceNames the services to send.
+     */
+    private void sendStartServiceLog(List<String> serviceNames) {
```
guperrot: `@WorkerThread`

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
MatkovIvan: `StartServiceLog` required `List` which in turn is required for `JSONUtils`. So, this change may affect some other code. `List` looks good for me here, let me know if you will nor change your mind.

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
MatkovIvan: Not for this line, I think

PR #596, opened 2018-01-04
treat 429 status code as unrecoverable error

dhei:


sdk/appcenter/src/test/java/com/microsoft/appcenter/http/HttpClientRetryerTest.java
```diff
@@ -98,7 +98,7 @@ public ServiceCall answer(InvocationOnMock invocationOnMock) throws Throwable {
 
     @Test
     public void retryOnceThenFail() {
-        final HttpException expectedException = new HttpException(403);
+        final HttpException expectedException = new HttpException(429);
```
guperrot: please roll back those changes in test, this is not the right change to make (see CI failure)

dhei: @guperrot yes, but the change to `exhaustRetries()` unit test is still required.

PR #613, opened 2018-01-31
use tester app to enable in-app updates if it’s installed

JessicaYeh:
Apps installed using the native tester app currently aren't able to successfully enable in-app updates, because the cookies stored on install aren't there. This is part of the work to fix that.

When enabling in-app updates, the Android SDK should check if an app with the ms-actesterapp:// scheme (belongs to the App Center tester app) is installed. If it is, then it should try to enable in-app updates by redirecting to that scheme instead of the update-setup page on the install portal. If for some reason enabling in-app updates fails, then the SDK should fall back to attempting to enable in-app updates using the update-setup page. This could happen if someone has the native tester app installed, but for some reason they still install an app from the install portal. The user will see two attempts to enable in-app updates in this case, but hopefully this is rare. We will track telemetry for how often this happens.

### DIFF
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -839,7 +839,7 @@ public class Distribute extends AbstractAppCenterService {
     }
 
     /**
-     * Store a flag for failure to enable updates from the tester apps, to later reattempt using the
+     * Store a flag for failure to enable updates from the tester apps, to later reattempt using the browser update setup.
      */
     synchronized void storeTesterAppUpdateSetupFailedParameter(@NonNull String requestId, @NonNull String updateSetupFailed) {
         if (requestId.equals(PreferencesStorage.getString(PREFERENCE_KEY_REQUEST_ID))) {
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -406,7 +406,6 @@ public class Distribute extends AbstractAppCenterService {
             AppCenterLog.info(LOG_TAG, "Launcher activity restarted.");
             if (getStoredDownloadState() == DOWNLOAD_STATE_COMPLETED) {
                 mWorkflowCompleted = false;
-                mTesterAppOpenedOrAborted = false;
                 mBrowserOpenedOrAborted = false;
             }
         }
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -465,6 +465,7 @@ public class Distribute extends AbstractAppCenterService {
             PreferencesStorage.remove(PREFERENCE_KEY_POSTPONE_TIME);
             PreferencesStorage.remove(PREFERENCE_KEY_UPDATE_SETUP_FAILED_PACKAGE_HASH_KEY);
             PreferencesStorage.remove(PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY);
+            PreferencesStorage.remove(PREFERENCE_KEY_TESTER_APP_UPDATE_SETUP_FAILED_MESSAGE_KEY);
 
             /* Disable the distribute info tracker. */
             mChannel.removeListener(mDistributeInfoTracker);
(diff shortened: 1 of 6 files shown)


sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -728,14 +736,30 @@ else if (mUnknownSourcesDialog != null) {
...
-            if (!mBrowserOpenedOrAborted) {
+            /* If not, open native app (if installed) to update setup, unless it already failed. */
+            String testerAppUpdateSetupFailedMessage = PreferencesStorage.getString(PREFERENCE_KEY_TESTER_APP_UPDATE_SETUP_FAILED_MESSAGE_KEY);
+            boolean testerAppUpdateSetupFailed = testerAppUpdateSetupFailedMessage != null && !testerAppUpdateSetupFailedMessage.isEmpty();
+            if (isAppCenterTesterAppInstalled() && !mTesterAppOpenedOrAborted && !testerAppUpdateSetupFailed) {
+                mTesterAppOpenedOrAborted = DistributeUtils.updateSetupUsingTesterApp(mForegroundActivity, mPackageInfo);
+            }
+
+            /* If the native app could not be opened, use the browser to update setup. */
+            if (!mBrowserOpenedOrAborted && (!mTesterAppOpenedOrAborted || testerAppUpdateSetupFailed)) {
```
guperrot: do you really need 2 different booleans?

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -728,14 +736,30 @@ else if (mUnknownSourcesDialog != null) {
...
                 mBrowserOpenedOrAborted = true;
             }
         }
     }
 
+    private boolean isAppCenterTesterAppInstalled() {
+        try {
+            PackageInfo testerApp = mContext.getPackageManager().getPackageInfo("com.microsoft.hockeyapp.testerapp", 0);
+            return testerApp != null;
+        } catch (PackageManager.NameNotFoundException e) {
```
guperrot: Log something or renamed `e` as `ignored` to resolve the warning.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
JessicaYeh: do you mean what's the reason to have both `mTesterAppOpenedOrAborted` and `testerAppUpdateSetupFailed`? or which two are you referring to because there are 3 here :D

guperrot: those 2 new ones, was wondering if logic could be simplified

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -816,6 +838,16 @@ synchronized void storeUpdateSetupFailedParameter(@NonNull String requestId, @No
         }
     }
 
+    /**
+     * Store a flag for failure to enable updates from the tester apps, to later reattempt using the
```
MatkovIvan: An intriguing ending :smile:

JessicaYeh: lol, oops :D

JessicaYeh:
retest this

JessicaYeh:
test

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/DistributeUtils.java
```diff
@@ -2,11 +2,13 @@
 
 import android.app.Activity;
 import android.app.Notification;
+import android.content.ActivityNotFoundException;
```
guperrot: unusued import, please cleanup imports using Android Studio.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/DistributeUtils.java
```diff
@@ -96,6 +104,40 @@ static String computeReleaseHash(@NonNull PackageInfo packageInfo) {
         return HashUtils.sha256(packageInfo.packageName + ":" + packageInfo.versionName + ":" + packageInfo.versionCode);
     }
 
+    /**
+     * Update setup using native tester app.
+     *
+     * @param activity    activity from which to start tester app.
+     * @param packageInfo package info.
+     *
+     * @return whether the tester app was opened or not.
```
guperrot: There is a warning here as there is no return parameter but you document one.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeBeforeApiSuccessTest.java
```diff
@@ -2,13 +2,15 @@
 
 import android.app.Activity;
 import android.app.Dialog;
+import android.content.ActivityNotFoundException;
```
guperrot: cleanup imports

PR #626, opened 2018-02-20
Log the missing `distributionGroupId` for the first session

roland-chernov-akvelon:
This PR addresses the issue with the first session log which is sent after app installation. The `distributionGroupId` field is missing in the first session log because it's sent before in-app updates are enabled. So this PR:

- Adds sending of distribution start session log after enabling in-app updates (first app launch after installation).
- Adds check to ignore distribution start session logs without session ID to handle cases when Analytics is disabled.
- Implements tests.

### DIFF
--- a/sdk/appcenter-distribute/src/androidTest/java/com/microsoft/appcenter/distribute/DistributeSerializerTest.java
+++ b/sdk/appcenter-distribute/src/androidTest/java/com/microsoft/appcenter/distribute/DistributeSerializerTest.java
@@ -1,0 +1,48 @@
+package com.microsoft.appcenter.distribute;
+
+import com.microsoft.appcenter.distribute.ingestion.models.DistributionStartSessionLog;
+import com.microsoft.appcenter.distribute.ingestion.models.json.DistributionStartSessionLogFactory;
+import com.microsoft.appcenter.ingestion.models.Log;
+import com.microsoft.appcenter.ingestion.models.LogContainer;
+import com.microsoft.appcenter.ingestion.models.json.DefaultLogSerializer;
+import com.microsoft.appcenter.ingestion.models.json.LogSerializer;
+import com.microsoft.appcenter.utils.UUIDUtils;
+
+import junit.framework.Assert;
+
+import org.json.JSONException;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.UUID;
+
+@SuppressWarnings("unused")
+public class DistributeSerializerTest {
+
+    @Test
+    public void serialize() throws JSONException {
+        LogContainer expectedContainer = new LogContainer();
+        List<Log> logs = new ArrayList<>();
+        {
+            DistributionStartSessionLog log = new DistributionStartSessionLog();
+            log.setTimestamp(new Date());
+            logs.add(log);
+        }
+        expectedContainer.setLogs(logs);
+        UUID sid = UUIDUtils.randomUUID();
+        for (Log log : logs) {
+            log.setSid(sid);
+        }
+
+        /* Serialize and deserialize logs container. */
+        LogSerializer serializer = new DefaultLogSerializer();
+        serializer.addLogFactory(DistributionStartSessionLog.TYPE, new DistributionStartSessionLogFactory());
+        String payload = serializer.serializeContainer(expectedContainer);
+        LogContainer actualContainer = serializer.deserializeContainer(payload);
+
+        /* Verify that logs container successfully deserialized. */
+        Assert.assertEquals(expectedContainer, actualContainer);
+    }
+}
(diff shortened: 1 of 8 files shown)


sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/channel/DistributeInfoTracker.java
```diff
@@ -26,14 +30,25 @@ public DistributeInfoTracker(String distributionGroupId) {
...
 
         /* Set current distribution group ID. */
         log.setDistributionGroupId(mDistributionGroupId);
     }
 
+    @Override
+    public boolean shouldFilter(@NonNull Log log) {
+
+        /* Filter out distribution start session log if session ID is not set (Analytics is disabled). */
+        if (log instanceof DistributionStartSessionLog && log.getSid() == null) {
```
guperrot: This works but will be confusing for aria team as they will also receive this log but will be filtered out without them knowing.

Another solution to it sooner is to use SessionContext from core and check if the last session has a sessionId or null.

roland-chernov-akvelon: Implemented using `SessionContext`. Thank you.

Please take a look.

sdk/appcenter-distribute/src/androidTest/java/com/microsoft/appcenter/distribute/DistributeSerializerTest.java
```diff
@@ -0,0 +1,45 @@
...
+            log.setTimestamp(new Date());
+            logs.add(log);
+        }
+        expectedContainer.setLogs(logs);
+        UUID sid = UUIDUtils.randomUUID();
+        for (Log log : logs) {
+            log.setSid(sid);
+        }
+
+        LogSerializer serializer = new DefaultLogSerializer();
```
guperrot: Always add a comment after a blank line in code

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1768,4 +1801,31 @@ synchronized void setInstalling(ReleaseDetails releaseDetails) {
...
+
+        /*
+         * Session starts before in-app updates setup (using browser) so the first start session log
+         * is sent without distributionGroupId value.
+         *
+         * Send the distribution start session log if start session log without distributionGroupId
+         * value was sent before
+         */
+        SessionContext.SessionInfo lastSession = SessionContext.getInstance().getSessionAt(System.currentTimeMillis());
+        if (lastSession == null) {
```
guperrot: lastSession can have null sessionid if analytics disabled, you need to test that also (a launch session is created by the core before analytics adds one)

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeBeforeApiSuccessTest.java
```diff
@@ -1498,4 +1501,40 @@ public boolean matches(Object argument) {
...
+        /* Enable in-app updates. */
+        start();
+        Distribute.getInstance().storeRedirectionParameters("r", "g", null);
+
+        /* Verify the log was sent. */
+        verify(mChannel).enqueue(any(DistributionStartSessionLog.class), eq(Distribute.getInstance().getGroupName()));
+    }
+
+    @Test
+    public void DontEnqueueDistributionStartSessionLogIfNoSessionsWereLoggedBefore() throws Exception {
```
guperrot: camelCase for method names, not PascalCase

sdk/appcenter-distribute/src/androidTest/java/com/microsoft/appcenter/distribute/DistributeSerializerTest.java
roland-chernov-akvelon: Added. Thank you.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
roland-chernov-akvelon: Fixed. Thank you.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeBeforeApiSuccessTest.java
roland-chernov-akvelon: Fixed. Also added a test to check for a `null` SID. Thank you.

guperrot:
@roland-chernov-akvelon please avoid rewriting the history in PR as we do incremental reviews when possible. I have to review the entire changes since it was forced pushed.

roland-chernov-akvelon:
@guperrot Okay, got it.

PR #627, opened 2018-02-21
Update distribution group ID if new ID is sent in the latest release details

roland-chernov-akvelon:
This PR implements updating of distribution group ID if in-app updated release has different distribution group.

Group ID may change if one user is added to different distribution groups and a new release was distributed to another group.

### DIFF
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/CheckDownloadTask.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/CheckDownloadTask.java
@@ -196,7 +196,7 @@ class CheckDownloadTask extends AsyncTask<Void, Void, DownloadProgress> {
         }
         String groupId = mReleaseDetails.getDistributionGroupId();
         String releaseHash = mReleaseDetails.getReleaseHash();
-        int releaseId= mReleaseDetails.getId();
+        int releaseId = mReleaseDetails.getId();
         AppCenterLog.debug(LOG_TAG, "Store downloaded group id=" + groupId + " release hash=" + releaseHash + " release id=" + releaseId);
         StorageHelper.PreferencesStorage.putString(PREFERENCE_KEY_DOWNLOADED_DISTRIBUTION_GROUP_ID, groupId);
         StorageHelper.PreferencesStorage.putString(PREFERENCE_KEY_DOWNLOADED_RELEASE_HASH, releaseHash);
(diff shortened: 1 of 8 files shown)


guperrot:
Will wait the PR to be green before reviewing.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/CheckDownloadTask.java
```diff
@@ -191,4 +184,22 @@ public void run() {
...
+     * After app update and restart, this info is used to report new download and to update group ID (if it's changed).
+     */
+    private void storeDownloadedReleaseDetails() {
+        if (mReleaseDetails == null) {
+            AppCenterLog.debug(LOG_TAG, "Downloaded release details are missing or broken, won't store.");
+            return;
+        }
+        String groupId = mReleaseDetails.getDistributionGroupId();
+        String releaseHash = mReleaseDetails.getReleaseHash();
+        int releaseId= mReleaseDetails.getId();
```
guperrot: please format file with Android Studio. It's recommended to use android studio commit UI to have the automated actions for that (cleanup, organize imports, reformat, code analysis).

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeBeforeDownloadTest.java
```diff
@@ -928,6 +930,103 @@ public ServiceCall answer(InvocationOnMock invocation) throws Throwable {
         PreferencesStorage.remove(PREFERENCE_KEY_DOWNLOADED_RELEASE_ID);
     }
 
+    @Test
+    public void shouldChangeDistributionGroupIdIfStoredIdDoesntMatchDownloadedId() throws Exception {
```
guperrot: We are verifying this behavior by mocking a previous download, so this behavior is right after start and before an api call is made, it should be in another test class if possible (like before api).

Or this could be in a new class.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/CheckDownloadTask.java
roland-chernov-akvelon: Fixed. Thank you for the advice.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeBeforeDownloadTest.java
roland-chernov-akvelon: Thanks for clarification. Moved to `DistributeBeforeApiSuccessTest` class.

PR #628, opened 2018-02-26
Update minor version in Android SDK

nikitag-ms:


nikitag-ms:
@guperrot, could you take a look?

PR #637, opened 2018-03-24
Feature/transmission target

TroubleMakerBen:
- don't init persistence and ingestion if no app secret has been provided but don't disable
- add default transmission target
- add add ability to call getTransmissionTarget
- add transmission target to log

PR #639, opened 2018-03-25
Fix tests for transmission target branch

MatkovIvan:
Tests for #637

### DIFF
--- a/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/AppCenterAndroidTest.java
+++ b/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/AppCenterAndroidTest.java
@@ -23,6 +23,11 @@ import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicReference;
 
 import static org.junit.Assert.*;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Matchers.isNull;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
 
 @SuppressWarnings("unused")
 public class AppCenterAndroidTest {
--- a/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/AppCenterAndroidTest.java
+++ b/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/AppCenterAndroidTest.java
@@ -22,6 +22,10 @@ import java.util.UUID;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicReference;
 
+import static com.microsoft.appcenter.AppCenter.APP_SECRET_KEY;
+import static com.microsoft.appcenter.AppCenter.KEY_VALUE_DELIMITER;
+import static com.microsoft.appcenter.AppCenter.PAIR_DELIMITER;
+import static com.microsoft.appcenter.AppCenter.TRANSMISSION_TARGET_TOKEN_KEY;
 import static org.junit.Assert.*;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.eq;
@@ -52,7 +56,7 @@ public class AppCenterAndroidTest {
     }
 
     @Test
-    public void getInstallId() throws IllegalAccessException, ClassNotFoundException, InstantiationException {
+    public void getInstallId() {
         assertNull(AppCenter.getInstallId().get());
         StorageHelper.initialize(mApplication);
         StorageHelper.PreferencesStorage.remove(PrefStorageConstants.KEY_INSTALL_ID);
@@ -82,7 +86,7 @@ public class AppCenterAndroidTest {
     }
 
     @Test
-    public void setDefaultLogLevelDebug() throws IllegalAccessException, ClassNotFoundException, InstantiationException {
+    public void setDefaultLogLevelDebug() {
         AppCenterLog.setLogLevel(Log.ASSERT);
         AppCenter.start(mApplication, UUIDUtils.randomUUID().toString());
         assertEquals(Log.WARN, AppCenter.getLogLevel());
@@ -90,7 +94,7 @@ public class AppCenterAndroidTest {
 
     @Test
     public void appSecretWithDelimiter() {
-        String secret = DUMMY_APP_SECRET + AppCenter.PAIR_DELIMITER;
+        String secret = DUMMY_APP_SECRET + PAIR_DELIMITER;
         AppCenter.start(mApplication, secret, DummyService.class);
         assertTrue(AppCenter.isEnabled().get());
         verify(DummyService.getInstance()).onStarted(any(Context.class), eq(DUMMY_APP_SECRET), isNull(String.class), any(Channel.class));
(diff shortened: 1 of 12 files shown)


guperrot:
@TroubleMakerBen looks good to me but since its going to land on your branch I let you merge.

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/AppCenterAndroidTest.java
```diff
@@ -22,11 +22,7 @@
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicReference;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
guperrot: No star import, please always to the following with android commit dialog: organize imports, reformat, code analysis etc... Organize imports will automatically replace start with actual imports.

MatkovIvan: This is a change came from the base branch

guperrot: I commented on that PR too on other places, but since it appears green (new changes), it also applies on yours.

guperrot: Looks like a github bug, the full diff does not show it.

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/AppCenterAndroidTest.java
```diff
@@ -23,15 +23,25 @@
...
+    private static final String DUMMY_APP_SECRET = "123e4567-e89b-12d3-a456-426655440000";
+
+    private static final String DUMMY_TRANSMISSION_TARGET_TOKEN = "snfbse234jknf";
+
     private Application mApplication;
 
     @Before
     public void setUp() throws IllegalAccessException, ClassNotFoundException, InstantiationException {
         AppCenter.unsetInstance();
+        DummyService.sharedInstance = null;
```
guperrot: unsetInstance is the pattern we use.

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/AppCenterAndroidTest.java
```diff
@@ -78,9 +88,84 @@ public void setDefaultLogLevelDebug() throws IllegalAccessException, ClassNotFou
...
+    public void targetTokenWithAppSecretTest() {
+        String secret = AppCenter.TRANSMISSION_TARGET_TOKEN_KEY + AppCenter.KEY_VALUE_DELIMITER + DUMMY_TRANSMISSION_TARGET_TOKEN + AppCenter.PAIR_DELIMITER + DUMMY_APP_SECRET;
+        AppCenter.start(mApplication, secret, DummyService.class);
+        assertTrue(AppCenter.isEnabled().get());
+        verify(DummyService.getInstance()).onStarted(any(Context.class), eq(DUMMY_APP_SECRET), eq(DUMMY_TRANSMISSION_TARGET_TOKEN), any(Channel.class));
+    }
+
+    @Test
+    public void targetTokenWithUnKeyedAppSecretTest() {
+        String secret = AppCenter.TRANSMISSION_TARGET_TOKEN_KEY + AppCenter.KEY_VALUE_DELIMITER + DUMMY_TRANSMISSION_TARGET_TOKEN +  AppCenter.PAIR_DELIMITER + AppCenter.APP_SECRET_KEY + AppCenter.KEY_VALUE_DELIMITER + DUMMY_APP_SECRET;
```
guperrot: Import statics?

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/AppCenterAndroidTest.java
MatkovIvan: Not for dummy ones in tests.
https://github.com/Microsoft/AppCenter-SDK-Android/blob/b4d77a54661fb55a5dd5ec997ab50818a6f0480d/sdk/appcenter/src/test/java/com/microsoft/appcenter/AppCenterTest.java#L130-L131

Do you want to change it everywhere?

guperrot: Ah dummy ones, ok. Somehow I thought it was on the new transmission classes. Dummy is fine.

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/AppCenterAndroidTest.java
```diff
@@ -72,15 +86,90 @@ public void accept(UUID uuid) {
...
+    @Test
+    public void targetTokenWithUnKeyedAppSecretTest() {
+        String secret = TRANSMISSION_TARGET_TOKEN_KEY + KEY_VALUE_DELIMITER + DUMMY_TRANSMISSION_TARGET_TOKEN + PAIR_DELIMITER + APP_SECRET_KEY + KEY_VALUE_DELIMITER + DUMMY_APP_SECRET;
+        AppCenter.start(mApplication, secret, DummyService.class);
+        assertTrue(AppCenter.isEnabled().get());
+        verify(DummyService.getInstance()).onStarted(any(Context.class), eq(DUMMY_APP_SECRET), eq(DUMMY_TRANSMISSION_TARGET_TOKEN), any(Channel.class));
+    }
+
+    @Test
+    public void unknownKeyTest() {
```
clpolet: Good test to have thanks.

sdk/appcenter/src/test/java/com/microsoft/appcenter/AppCenterTest.java
```diff
@@ -200,6 +198,27 @@ public Void answer(InvocationOnMock invocation) throws Throwable {
 
         /* Mock network state helper. */
         when(NetworkStateHelper.getSharedInstance(any(Context.class))).thenReturn(mNetworkStateHelper);
+
+        /* Mock string splitter. */
+        whenNew(TextUtils.SimpleStringSplitter.class).withAnyArguments().then(new Answer<TextUtils.SimpleStringSplitter>() {
```
clpolet: I pushed a new implementation using pure java instead of Android string splitter here: feature/transmission-token-parsing. That way there is no need for mocking. Can you please take a look and merge it to this branch ?

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -442,9 +442,7 @@ private synchronized boolean instanceConfigure(Application application, String a
 
             /* A value with no key is default to the app secret. */
             if (!keyValueSplitter.hasNext()) {
-                if (keyValue.indexOf(KEY_VALUE_DELIMITER) == -1) {
```
clpolet: The reason behind this test was in the case of "appsecret=xxxxxxx;target=", the "target" would become the app secret without the extra check. But I think you are right in a way, no need for multiplying checks like that for corner cases we should instead validate the whole string via regex before parsing it.  If that regex check is really quick to implement for you then go ahead otherwise I'll create a task for it, just let me know.

sdk/appcenter/src/test/java/com/microsoft/appcenter/AppCenterTest.java
clpolet: I should have told you about this implementation before, it was lying down in a local branch for more than a week :sweat_smile:.

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
MatkovIvan: It didn't check anything. This check was fully equal to check above.

sdk/appcenter/src/test/java/com/microsoft/appcenter/AppCenterTest.java
MatkovIvan: These PRs are targeted to different branches. I think we need to merge some of them before merging the changes between them.

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
MatkovIvan: Yes. I mean `if (!keyValueSplitter.hasNext())`  does the same thing

sdk/appcenter/src/test/java/com/microsoft/appcenter/AppCenterTest.java
clpolet: True :/.

clpolet:
Close because of duplicate work, will keep `feature/transmission-token-parsing` branch but we'll still port the few unit tests from here.

PR #642, opened 2018-04-04
Fix ability to run push test in Android Studio

guperrot:


sdk/appcenter-push/src/test/java/com/microsoft/appcenter/push/PushTest.java
```diff
@@ -57,10 +57,9 @@
 import static org.powermock.api.mockito.PowerMockito.verifyStatic;
 import static org.powermock.api.mockito.PowerMockito.when;
 
-
+@RunWith(PowerMockRunner.class)
 @SuppressWarnings({"unused", "MissingPermission"})
 @PrepareForTest({
-        Push.class,
```
guperrot: That's the key to make it work with the annotation. Otherwise if we needed to use whenNew for Push we could not rely on this fix, hopefully not using whenNew on Push for this test class.

PR #638, opened 2018-03-24
PR containing transmission target logic

TroubleMakerBen:

PR #643, opened 2018-04-04
Fix tests for transmission target branch

MatkovIvan:
Tests for #638

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
@@ -453,12 +453,12 @@ public class AppCenter {
             for (String pair : pairs) {
 
                 /* Split key and value. */
-                String[] keyValue = pair.split(KEY_VALUE_DELIMITER);
+                String[] keyValue = pair.split(KEY_VALUE_DELIMITER, -1);
                 String key = keyValue[0];
 
                 /* A value with no key is default to the app secret. */
                 if (keyValue.length == 1) {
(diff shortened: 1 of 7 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -453,12 +453,12 @@ private synchronized boolean instanceConfigure(Application application, String a
             for (String pair : pairs) {
 
                 /* Split key and value. */
-                String[] keyValue = pair.split(KEY_VALUE_DELIMITER);
+                String[] keyValue = pair.split(KEY_VALUE_DELIMITER, -1);
```
MatkovIvan: @clpolet https://stackoverflow.com/questions/14602062

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -299,9 +301,6 @@ private void suspend(boolean deleteLogs, Exception exception) {
     }
 
     private void deleteLogsOnSuspended(final GroupState groupState) {
-        if (mPersistence == null) {
```
TroubleMakerBen: Why remove the null check?

MatkovIvan: Yes, I know. this checked before. In *this* place no way to get `mPersistence == null` => no test to cover it => not 100% coverage

PR #645, opened 2018-04-04
Fix push standard/custom data management

guperrot:


sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/PushIntentUtils.java
```diff
@@ -41,14 +38,15 @@
     @VisibleForTesting
     static final String EXTRA_GOOGLE_MESSAGE_ID = "google.message_id";
```
MatkovIvan: `EXTRA_GOOGLE_PREFIX + "message_id"`

PR #646, opened 2018-04-05
Increased event and handled error max property count and item length

stankovski:


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -41,7 +41,7 @@
      * Max length of properties.
      */
     @VisibleForTesting
-    static final int MAX_PROPERTY_ITEM_LENGTH = 64;
```
clpolet: The limits are also in the doc of the trackEvent/Page  methods see `public static void trackEvent(String name)`

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelper.java
```diff
@@ -86,12 +86,12 @@
...
      * Max number of properties.
      */
-    public static final int MAX_PROPERTY_COUNT = 5;
+    public static final int MAX_PROPERTY_COUNT = 20;
 
     /**
      * Max length of properties.
      */
-    public static final int MAX_PROPERTY_ITEM_LENGTH = 64;
+    public static final int MAX_PROPERTY_ITEM_LENGTH = 125;
```
guperrot: Why this specific number instead of something like 128?

stankovski: For feature parity with Xamarin Insights

stankovski:
@guperrot I've updated tests

PR #647, opened 2018-04-05
Fix gradle warning

MatkovIvan:
```
Configuration 'compile' is obsolete and has been replaced with 'implementation'.
It will be removed at the end of 2018
```


sdk/appcenter-analytics/build.gradle
```diff
@@ -2,5 +2,5 @@ project.description = 'This package contains functionalities to collect session,
 evaluationDependsOn(':sdk')
 
 dependencies {
-    compile project(':sdk:appcenter')
+    implementation project(':sdk:appcenter')
```
guperrot: api, we don't want people to have to include core explicitly.
Same comment for others.

guperrot:
Can you trigger publish internal job and test if you can consume the aar without issues in a blank app with jump to sources/mouse over docs working as expected? Without having to add the core.

guperrot:
@MatkovIvan could you test in a blank app with the aar uploaded from vsts to snapshot repo?

PR #653, opened 2018-04-10
Update jacoco, gradle and android gradle plugin versions

TroubleMakerBen:
* verified using both build jobs to create artifacts
* verified by integrating artifacts in simple app (sources available)
* verified by building test app in App Center

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
+        classpath 'com.android.tools.build:gradle:3.1.2'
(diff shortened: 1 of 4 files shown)


TroubleMakerBen:
This is work in progress, needs more verification.

guperrot:
@TroubleMakerBen I pushed force history to catch up with develop and new Android Studio. Jacoco version needs to match Android debug instrumentation tests which is now 0.7.9, we cannot use any other version otherwise combining code coverage binaries between junit and android device tests fail (hence why we had artificial decreased coverage).

Now the remaining problem is bintrayUpload failing if some other gradle command was run before (it works when we dont run tests with the other job). I think a clean step in vsts might be a workaround, will try.

guperrot:
git clean on VSTS worked...

PR #665, opened 2018-05-05
Fix deadlock between AppCenter and channel classes

guperrot:


guperrot:
This should resolve the deadlock reported at https://github.com/Microsoft/AppCenter-SDK-DotNet/issues/649.

PR #668, opened 2018-05-07
Register a one collector listener when AppCenter starts

clpolet:


jwallra:
We should be vigilant about adding descriptions to our PRs here as well ;)

guperrot:
The title says it all, do you want more information? @jwallra

guperrot:
For now the channel listener implementation is empty and further PRs will add the desired behavior of redirecting some logs to One Collector when a target token is used.

PR #669, opened 2018-05-08
Add one collector log group when sending eligible appcenter logs 

MatkovIvan:
Merge #668 and change target branch before merge

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
@@ -175,6 +175,8 @@ public class DefaultChannel implements Channel {
         if (mPersistence == null) {
             return;
         }
+
+        /* TODO Remove this check once onGroupAdded callback is implemented. */
         if (mGroupStates.containsKey(groupName)) {
             return;
         }
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
@@ -175,6 +175,9 @@ public class DefaultChannel implements Channel {
         if (mPersistence == null) {
             return;
         }
+        if (mGroupStates.containsKey(groupName)) {
+            return;
+        }
 
         /* Init group. */
         AppCenterLog.debug(LOG_TAG, "addGroup(" + groupName + ")");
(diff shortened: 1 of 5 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/OneCollectorChannelListener.java
```diff
@@ -9,13 +9,51 @@
  */
 public class OneCollectorChannelListener implements Channel.Listener {
 
+    /**
+     * Maximum time interval in milliseconds after which a synchronize will be triggered, regardless of queue size.
+     */
+    static final int ONE_COLLECTOR_TRIGGER_INTERVAL = 3 * 1000;
```
guperrot: Should add @VisibleForTesting to all the fields if the package private visibility is used only for testing.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/OneCollectorChannelListener.java
```diff
@@ -9,13 +9,51 @@
...
+
+    /**
+     * Maximum number of requests being sent for the group.
+     */
+    static final int ONE_COLLECTOR_TRIGGER_MAX_PARALLEL_REQUESTS = 3;
+
+    /**
+     * Postfix for One Collector's groups.
+     */
+    static final String ONE_COLLECTOR_GROUP_NAME_POSTFIX = "/one";
```
guperrot: According to me, suffix is more common than postfix.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/OneCollectorChannelListener.java
```diff
@@ -9,13 +9,51 @@
...
+        String oneCollectorGroupName = getOneCollectorGroupName(groupName);
+        mChannel.addGroup(oneCollectorGroupName, ONE_COLLECTOR_TRIGGER_COUNT, ONE_COLLECTOR_TRIGGER_INTERVAL, ONE_COLLECTOR_TRIGGER_MAX_PARALLEL_REQUESTS, null);
     }
 
     @Override
     public boolean shouldFilter(@NonNull Log log) {
         return false;
     }
+
+    private String getOneCollectorGroupName(@NonNull String groupName) {
```
guperrot: Should add 1 liner "javadoc" comment to all the private fields/constructors/methods.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -175,6 +175,9 @@ public synchronized void addGroup(final String groupName, int maxLogsPerBatch, l
         if (mPersistence == null) {
             return;
         }
+        if (mGroupStates.containsKey(groupName)) {
```
guperrot: We discussed this with Clement and we will introduce a onGroupAdded callback instead in listener to register group. This will make sure pending logs of one collector are resumed on app restart even if no app center log enqueued (and will also remove the need for this check).

This would be a new work item but for the time being I would add a `/* TODO Remove this check once onGroupAdded callback is implemented. */`.

PR #672, opened 2018-05-09
Fix notification message being truncated

guperrot:
Apply big text style like Distribute and also like Firebase SDK does.


guperrot:
Somehow build was canceled near the end, restarting.

PR #674, opened 2018-05-10
Alternate Ingestion

Abpanwar:
Work in Progress for alternate ingestion.


guperrot:
Add ability to specify an alternate ingestion implementation for log groups (not yet used for real, this will be in other PRs).

PR #684, opened 2018-05-16
Move AppCenter Analytics validation to shouldFilter callback

MatkovIvan:


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/channel/AnalyticsValidator.java
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Max number of properties.
+     */
+    @VisibleForTesting
+    static final int MAX_PROPERTY_COUNT = 20;
+
+    @Override
+    public boolean shouldFilter(@NonNull Log log) {
+        if (log instanceof EventLog) {
```
guperrot: Works well, last comment though that I missed, since you introduced a base type, use it here to combine the 2 if conditions.

MatkovIvan: This was done intentionally to exclude types like
```java
new LogWithNameAndProperties() {
    @Override
    public String getType() {
        return "";
    }
})
```

PR #685, opened 2018-05-16
Combine the 2 if conditions

MatkovIvan:


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/channel/AnalyticsValidator.java
```diff
@@ -37,10 +37,9 @@
 
     @Override
     public boolean shouldFilter(@NonNull Log log) {
-        if (log instanceof EventLog) {
-            return !validateLog((EventLog) log);
-        } else if (log instanceof PageLog) {
-            return !validateLog((PageLog) log);
+        //noinspection SimplifiableIfStatement
```
MatkovIvan: Suggested option:
```java
return (log instanceof EventLog || log instanceof PageLog) && !validateLog((LogWithNameAndProperties) log);
```

guperrot: Yeah I was curious and tested in Android Studio before seeing this comment ;) I agree we can disable the warning for this one as it would make it less readable.

PR #687, opened 2018-05-18
Common Schema conversion and JSON serialization

Abpanwar:
Part A conversion of event logs.


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/ingestion/models/one/json/CommonSchemaEventLogFactory.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public Collection<CommonSchemaLog> toCommonSchemaLogs(Log log) {
+
+        /* Convert events only for now. */
+        if (log instanceof EventLog) {
+
+            /* 1 App Center log can map to multiple common schema logs if multiple keys. */
+            Collection<CommonSchemaLog> commonSchemaLogs = new LinkedList<>();
```
clpolet: In general yes CS logs can have multiple keys but at this stage we know the log is an EventLog and it will only contain one token so should we still go with the multiple keys approach for this log type?

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/Channel.java
```diff
@@ -95,12 +96,16 @@
     interface Listener {
 
         /**
-         * Called whenever a new group is added.
+         * Called whenever a new group is added
```
clpolet: Missing a `.` in the end here but that's quite picky from me :sweat_smile:

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/ingestion/models/one/json/CommonSchemaEventLogFactory.java
clpolet: Not a problem since it's already enforced by Analytics just wanted to highlight it.

guperrot: This code will be likely shared with session log.

PR #690, opened 2018-05-21
Add onGloballyEnabled callback to channel listener

bmourat:


sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/DefaultChannelTest.java
```diff
@@ -811,6 +811,12 @@ public void listener() {
         verify(listener).shouldFilter(log);
         verifyNoMoreInteractions(listener);
 
+        /* Check enable/disable. */
```
guperrot: We have a method called @Test setEnabled, I think you can do listener checks there instead.

PR #706, opened 2018-05-25
Persist target token

guperrot:
With encryption.

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/AbstractLog.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/AbstractLog.java
@@ -188,31 +188,27 @@ public abstract class AbstractLog implements Log {
         }
     }
 
-    @Override
     @SuppressWarnings("SimplifiableIfStatement")
+    @Override
     public boolean equals(Object o) {
-        if (this == o) {
-            return true;
-        }
-        if (o == null || getClass() != o.getClass()) {
-            return false;
-        }
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
         AbstractLog that = (AbstractLog) o;
-        if (timestamp != null ? !timestamp.equals(that.timestamp) : that.timestamp != null) {
+
+        if (!transmissionTargetTokens.equals(that.transmissionTargetTokens)) return false;
+        if (timestamp != null ? !timestamp.equals(that.timestamp) : that.timestamp != null)
             return false;
-        }
-        if (sid != null ? !sid.equals(that.sid) : that.sid != null) {
+        if (sid != null ? !sid.equals(that.sid) : that.sid != null) return false;
+        if (distributionGroupId != null ? !distributionGroupId.equals(that.distributionGroupId) : that.distributionGroupId != null)
             return false;
-        }
-        if (distributionGroupId != null ? !distributionGroupId.equals(that.distributionGroupId) : that.distributionGroupId != null) {
-            return false;
-        }
         return device != null ? device.equals(that.device) : that.device == null;
     }
 
     @Override
     public int hashCode() {
-        int result = timestamp != null ? timestamp.hashCode() : 0;
+        int result = transmissionTargetTokens.hashCode();
+        result = 31 * result + (timestamp != null ? timestamp.hashCode() : 0);
         result = 31 * result + (sid != null ? sid.hashCode() : 0);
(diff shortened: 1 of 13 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -96,37 +115,35 @@
...
      *
-     * @param database The database name
-     * @param table    The table name
-     * @param version  The version of current schema.
+     * @param context application context.
      */
-    DatabasePersistence(String database, String table, @SuppressWarnings("SameParameterValue") int version) {
-        this(database, table, version, Persistence.DEFAULT_CAPACITY);
+    public DatabasePersistence(Context context) {
+        this(context, 2, SCHEMA, Persistence.DEFAULT_CAPACITY);
```
MatkovIvan: Let's use a constant instead of a magic number

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/AbstractLog.java
```diff
@@ -188,31 +188,27 @@ public void read(JSONObject object) throws JSONException {
...
     @SuppressWarnings("SimplifiableIfStatement")
+    @Override
     public boolean equals(Object o) {
-        if (this == o) {
-            return true;
-        }
-        if (o == null || getClass() != o.getClass()) {
-            return false;
-        }
+        if (this == o) return true;
```
MatkovIvan: Are there any exceptions to "Use `{}` even if you have single operation in block."?

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -539,11 +538,22 @@ void setSQLiteOpenHelper(@NonNull SQLiteOpenHelper helper) {
...
     /**
-     * Error listener for DatabaseManager.
+     * Database listener.
      */
-    interface ErrorListener {
+    public interface Listener {
+
+        /**
+         * Called when upgrade is performed on the database.
+         *
```
MatkovIvan: Missing `@param`s

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -96,37 +115,35 @@
...
         mPendingDbIdentifiersGroups = new HashMap<>();
         mPendingDbIdentifiers = new HashSet<>();
-        mDatabaseStorage = DatabaseStorage.getDatabaseStorage(database, table, version, SCHEMA, maxRecords,
-                new DatabaseStorage.DatabaseErrorListener() {
+        mDatabaseStorage = DatabaseStorage.getDatabaseStorage(DATABASE, TABLE, version, schema, maxRecords,
+                new DatabaseManager.Listener() {
+
+                    @Override
+                    public boolean onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
+                        db.execSQL("ALTER TABLE " + TABLE + " ADD COLUMN `" + COLUMN_TARGET_TOKEN + "` TEXT");
```
MatkovIvan: Add versions check.

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/AbstractLog.java
guperrot: If you generate equals/hashcode yes. Its' owned by idea code generator so we don't want to deal with that.

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
guperrot: This is not called on downgrade and for this version we don't need checks yet.
We can introduce version check the next time we change schema, would be dead code to cover in unit tests otherwise, I will add a comment instead.

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/AbstractLog.java
guperrot: Since it's generated code this is also why we ignore "simplify if statement" with an annotation.

PR #717, opened 2018-06-08
Compress payload if large

guperrot:


sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClient.java
```diff
@@ -74,6 +86,11 @@
      */
     private static final int READ_TIMEOUT = 20000;
 
+    /**
+     * Minimum length to use gzip.
+     */
+    private static final long MIN_GZIP_LENGTH = (long) (1024 * 1.4);
```
MatkovIvan: I guess it should be 1400, not 1433

guperrot: Ok let's simplify.

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClient.java
```diff
@@ -134,26 +151,55 @@ private static String doHttpCall(String urlString, String method, Map<String, St
...
-            if (method.equals(METHOD_POST) && callTemplate != null) {
-                String payload = callTemplate.buildRequestBody();
+            /* Send payload. */
+            if (binaryPayload != null) {
+
+                /* Compress payload if large enough to be worth it. */
                 AppCenterLog.verbose(LOG_TAG, payload);
+                if (shouldCompress) {
+                    ByteArrayOutputStream gzipBuffer = new ByteArrayOutputStream(binaryPayload.length);
+                    GZIPOutputStream gzipStream = new GZIPOutputStream(gzipBuffer);
```
jaelim-ms: You don't need `try` - `catch` to close output stream in error case?

jaelim-ms: **it never fails** -> How do you guarantee?

jaelim-ms: It should be okay since ByteArrayOutputStream.close doesn't do anything.

guperrot: Its in the source code of the 2 classes used in that block, it's all in memory. Gzip will only fail if wrapping a stream that fails (like a network stream), but here we wrap in memory.

PR #718, opened 2018-06-08
Fix analytics listener when common schema is used

guperrot:
@jwallra thanks for having noticed the toast missing in Sasquatch.


apps/sasquatch/src/projectDependency/java/com/microsoft/appcenter/sasquatch/listeners/SasquatchAnalyticsListener.java
```diff
@@ -0,0 +1,75 @@
+package com.microsoft.appcenter.sasquatch.listeners;
```
guperrot: Fork of the file as jCenter dependency does not have CommonSchemaEventLog type yet.
Easier than using reflection.
Will merge the files after release.

PR #719, opened 2018-06-08
Validate null property value for Common Schema

guperrot:


sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/ingestion/models/one/PartCUtilsAndroidTest.java
```diff
@@ -40,6 +40,14 @@ public void nullKey() {
         PartCUtils.addPartCFromLog(properties, log);
     }
 
+    @Test(expected = IllegalArgumentException.class)
```
jaelim-ms: Can you also add a test with empty string value?

guperrot: Done

PR #736, opened 2018-06-22
Introduce AppCenter.startFromLibrary

guperrot:
The feature is not 100% complete yet but we want to split in multiple PRs if possible or at least start review now before it gets too big.

What works with the feature:

* startFromLibrary before or after the app and track events.
* make sure start service log is sent after the app specifies app secret.

What does not work yet:

* Send start session log if app starts analytics after library (Fixed by #737)
* Preventing multiple calls to AppCenter.start with null appSecret. There is a pending discussion.

### DIFF
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
@@ -1,6 +1,7 @@
 package com.microsoft.appcenter.sasquatch.activities;
 
 import android.annotation.SuppressLint;
+import android.app.Application;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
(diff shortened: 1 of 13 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/Channel.java
```diff
@@ -10,6 +10,14 @@
  */
 public interface Channel {
 
+    /**
+     * Set app secret. Intended usage is to use that only if there was no app secret at initialization time.
+     * The behavior is undefined if trying to update app secret a second time.
```
jwallra: Why would the behavior be undefined?

guperrot: Because we don't try to handle that case since it does not happen. Undefined as not defined by the spec. We validate inputs and flow in AppCenter already (configure no more than once).

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
```diff
@@ -239,9 +240,13 @@ private PushListener getPushListener() {
         return sPushListener;
     }
 
-    private void startAppCenter(StartType startType) {
-        String appId = sSharedPreferences.getString(APP_SECRET_KEY, getString(R.string.app_secret));
-        String targetId = sSharedPreferences.getString(TARGET_KEY, getString(R.string.target_id));
+    static void startAppCenter(Application application, String startTypeString) {
```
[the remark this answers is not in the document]
guperrot: There is no breaking change, it's just a test app thing to allow different start modes. Not a SDK change.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/Channel.java
guperrot: (this is an internal class not meant for public usage)

PR #737, opened 2018-06-25
Multitenancy: Fix start session / page tracking and static track calls behavior

guperrot:
* Improve #736 
* Fixes sending start session if library starts Analytics then app starts Analytics (sent when app starts).
* Fixes automatic page tracking the same way as previous point.
* If Crashes started at app level with app secret, and Analytics started only from library, make sure we don't allow static trackEvent/trackPage calls (only AnalyticsTransmissionTarget calls are allowed for libraries).

### DIFF
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
@@ -445,7 +445,7 @@ public class Analytics extends AbstractAppCenterService {
         /* If we enabled the service. */
         if (enabled) {
 
-            /* Check if we have an app secret and enable corresponding features. */
+            /* Check if service started at application level and enable corresponding features. */
             startAppLevelFeatures();
         }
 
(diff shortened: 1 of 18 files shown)


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -544,10 +585,24 @@ private synchronized void setInstanceListener(AnalyticsListener listener) {
     }
 
     @Override
-    public synchronized void onStarted(@NonNull Context context, String appSecret, String transmissionTargetToken, @NonNull Channel channel) {
-        super.onStarted(context, appSecret, transmissionTargetToken, channel);
+    public synchronized void onStarted(@NonNull Context context, @NonNull Channel channel, String appSecret, String transmissionTargetToken, boolean startFromApp) {
```
dhei: `startedFromApp` would be more consistent with the other variables. Is it intentional or a typo?

guperrot: I can rename

PR #742, opened 2018-06-27
Nest transmission targets

guperrot:
For now it's not useful as no setProperty or isEnabled/setEnabled but will come later and will benefit from nesting.

### DIFF
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
@@ -2,6 +2,7 @@ package com.microsoft.appcenter.analytics;
 
 import android.support.annotation.NonNull;
 
+import java.util.HashMap;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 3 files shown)


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
```diff
@@ -46,6 +55,23 @@ public void trackEvent(String name, Map<String, String> properties) {
...
+    /**
+     * Create a new transmission target based on the properties of the current target.
+     *
+     * @param transmissionTargetToken The transmission target token of the new transmission target.
+     * @return The new transmission target.
+     */
+    public synchronized AnalyticsTransmissionTarget getTransmissionTarget(String transmissionTargetToken) {
+
+        /* Reuse instance if a child with the same token has already been created. */
+        AnalyticsTransmissionTarget childTarget = mChildrenTargets.get(transmissionTargetToken);
```
clpolet: There is one edge case covered in Apple, basically if transmissionTargetToken and mTransmissionTargetToken are the same then getTransmissionTarget returns self. It's opened to discussion.

clpolet: So we won't do anything actually in that case so this implementation is good. The nested target having the same token can still have a different behavior than its parent.

PR #743, opened 2018-06-27
Nesting is/setEnabled in Analytics transmissions

guperrot:
First PR for nesting.
TODO: delete logs for disabled targets but this first PR can already be merged.

### DIFF
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
@@ -1,9 +1,20 @@
 package com.microsoft.appcenter.analytics;
 
 import android.support.annotation.NonNull;
+import android.support.annotation.WorkerThread;
+
+import com.microsoft.appcenter.utils.AppCenterLog;
+import com.microsoft.appcenter.utils.async.AppCenterFuture;
+import com.microsoft.appcenter.utils.async.DefaultAppCenterFuture;
+import com.microsoft.appcenter.utils.storage.StorageHelper;
 
 import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.ListIterator;
 import java.util.Map;
+
+import static com.microsoft.appcenter.analytics.Analytics.LOG_TAG;
 
 /**
  * Target for advanced transmission target usage.
@@ -16,6 +27,11 @@ public class AnalyticsTransmissionTarget {
     private final String mTransmissionTargetToken;
 
     /**
+     * Parent target if any.
+     */
+    private final AnalyticsTransmissionTarget mParentTarget;
+
+    /**
      * Children targets for nesting.
      */
     private final Map<String, AnalyticsTransmissionTarget> mChildrenTargets = new HashMap<>();
@@ -24,9 +40,11 @@ public class AnalyticsTransmissionTarget {
      * Create a new instance.
      *
      * @param transmissionTargetToken The token for this transmission target.
+     * @param parentTarget            Parent transmission target.
      */
-    AnalyticsTransmissionTarget(@NonNull String transmissionTargetToken) {
+    AnalyticsTransmissionTarget(@NonNull String transmissionTargetToken, final AnalyticsTransmissionTarget parentTarget) {
         mTransmissionTargetToken = transmissionTargetToken;
+        mParentTarget = parentTarget;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
```diff
@@ -80,4 +147,29 @@ public synchronized AnalyticsTransmissionTarget getTransmissionTarget(String tra
...
+    }
+
+    @WorkerThread
+    private boolean isEnabledInStorage() {
+        return StorageHelper.PreferencesStorage.getBoolean(getEnabledPreferenceKey(), true);
+    }
+
+    @WorkerThread
+    private boolean areAncestorsEnabled() {
+        for (AnalyticsTransmissionTarget target = mParentTarget; target != null; target = target.mParentTarget) {
```
jwallra: @clpolet and I talked about this and you shouldn't have to check anything other than the immediate parent.  If the immediate parent is disabled then we should not allow children to be enabled.  If the immediate parent is enabled, that means that nothing in the chain above it could be disabled, because otherwise the immediate parent would be disabled.

This logic should be able to be simplified this way.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
```diff
@@ -66,12 +84,61 @@ public synchronized AnalyticsTransmissionTarget getTransmissionTarget(String tra
...
+            public void run() {
+
+                /*
+                 * Like the relation between AppCenter and Analytics, we cannot change state if one of the parent is disabled.
+                 * If this callback is called then it was already checked that AppCenter and Analytics are both enabled.
+                 */
+                if (areAncestorsEnabled()) {
+
+
+                    /* Propagate state to this instance then all descendants without a recursive call. */
```
clpolet: Any reasons why we are not doing cascading calls to child.setEnabled(enabled)?

clpolet: Or is it just optimisation ?

guperrot: We are

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
jwallra: Adding a comment here so I can request changes.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
clpolet: Ah maybe because child will call isEnabled on their parent.

guperrot: Yes not using recursive calls is a habit.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
guperrot: Checking that now.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
```diff
@@ -66,12 +84,61 @@ public synchronized AnalyticsTransmissionTarget getTransmissionTarget(String tra
...
+        Analytics.getInstance().postCommand(new Runnable() {
+
+            @Override
+            public void run() {
+
+                /*
+                 * Like the relation between AppCenter and Analytics, we cannot change state if one of the parent is disabled.
+                 * If this callback is called then it was already checked that AppCenter and Analytics are both enabled.
+                 */
+                if (areAncestorsEnabled()) {
```
jwallra: We don't have to do any of this if this.isEnabled != enabled (pseudocode).  Also we shouldn't have to update to enabled if Analytics is disabled.

guperrot: I added a unit test that proves otherwise. If you check only 1 parent, this new unit test will fail.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
```diff
@@ -24,9 +40,11 @@
      * Create a new instance.
      *
      * @param transmissionTargetToken The token for this transmission target.
+     * @param parentTarget            Parent transmission target.
      */
-    AnalyticsTransmissionTarget(@NonNull String transmissionTargetToken) {
+    AnalyticsTransmissionTarget(@NonNull String transmissionTargetToken, final AnalyticsTransmissionTarget parentTarget) {
         mTransmissionTargetToken = transmissionTargetToken;
+        mParentTarget = parentTarget;
```
clpolet: Shouldn't we initialize the state of this transmission target with the state of its parents to align both? Then later we could trust the immediate parent state without checking other parents.

guperrot: We can't do that in the main thread in Android so it would actually be more complex as you would have to post a command and do the get and set in background, and in the meantime you can add children before the background task executes so you would probably do too many storage operations if that happens.

clpolet: I see, we don't have async public APIs on iOS yet so it would work there.

PR #744, opened 2018-06-27
Reintroduce firebase dependency

MatkovIvan:


sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -173,7 +173,12 @@ public static void checkLaunchedFromNotification(Activity activity, Intent inten
      * Sender ID of your project before starting the Push service.
      *
      * @param senderId sender ID of your project.
+     *
+     * @deprecated Starting Android P at release date and later all versions of Android after April 2019,
+     * Firebase SDK is required to use push. Please follow the migration guide at https://aka.ms/...
+     * TODO: Set migration guide link.
```
guperrot: You can use https://aka.ms/acfba, thanks in advance (content will be updated or moved, for now it's the reversed migration page).

sdk/appcenter-push/build.gradle
```diff
@@ -5,7 +5,7 @@ dependencies {
     api project(':sdk:appcenter')
     def fcm = "com.google.firebase:firebase-messaging:${rootProject.ext.firebaseLibVersion}"
     def fc = "com.google.firebase:firebase-core:${rootProject.ext.firebaseLibVersion}"
-    compileOnly fcm
+    api fcm
     compileOnly fc
     testImplementation fcm
```
guperrot: I don't think we need that one anymore.

sdk/appcenter-push/build.gradle
```diff
@@ -5,7 +5,7 @@ dependencies {
     api project(':sdk:appcenter')
     def fcm = "com.google.firebase:firebase-messaging:${rootProject.ext.firebaseLibVersion}"
     def fc = "com.google.firebase:firebase-core:${rootProject.ext.firebaseLibVersion}"
-    compileOnly fcm
+    api fcm
     compileOnly fc
```
guperrot: There was an issue in a projects not pulling firebase core and just pulling firebase messaging, can you check again? We might need to expose that one too.

apps/sasquatch/build.gradle
```diff
@@ -75,21 +75,17 @@ dependencies {
...
 }
 
-/*
- * This will also bypass google-services.json if executing assembleDebug wildcard task.
- * This is needed to build from App Center build as we don't commit google-services.json and
- * App Center build cannot detect our gradle tasks so we make it execute assembleDebug with a
- * work around. We are interested in only a vanilla build variant (no Firebase) for such builds.
- */
-if (getGradle().getStartParameter().getTaskRequests().toString().contains("Firebase")) {
+// We are interested in excluding the plugin only for the vanilla sasquatch build variant.
```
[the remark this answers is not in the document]
MatkovIvan: I don't think that it's still valid.
- `This will also bypass google-services.json if executing assembleDebug wildcard task.` - it isn't true;
- `This is needed to build from App Center build as we don't commit google-services.json and App Center build cannot detect our gradle tasks so we make it execute assembleDebug with a work around.` - not true too;

apps/sasquatch/google-services.json
```diff
@@ -0,0 +1,35 @@
+{
```
jaelim-ms: Why do you check in this file? I know this is mandatory but we used to have a local copy of this one.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
```diff
@@ -126,6 +126,7 @@ protected void onCreate(Bundle savedInstanceState) {
         if (!TextUtils.isEmpty(apiUrl)) {
             Distribute.setApiUrl(apiUrl);
         }
+        //noinspection deprecation
```
jaelim-ms: New line before this one.

sdk/appcenter-push/src/test/java/com/microsoft/appcenter/push/PushTest.java
```diff
@@ -622,20 +623,24 @@ public void registerWithoutFirebaseOrSenderId() {
...
+        verifyStatic();
+        AppCenterLog.warn(anyString(), anyString());
         Push.getInstance().onActivityResumed(mock(Activity.class));
         verify(contextMock, times(2)).startService(any(Intent.class));
     }
 
     @Test
     public void registerWithoutFirebaseStartServiceThrowsRuntimeException() {
         when(FirebaseInstanceId.getInstance()).thenThrow(new IllegalStateException());
+        //noinspection deprecation
```
jaelim-ms: New line before this one.

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -173,7 +173,11 @@ public static void checkLaunchedFromNotification(Activity activity, Intent inten
      * Sender ID of your project before starting the Push service.
      *
      * @param senderId sender ID of your project.
+     *
+     * @deprecated Starting Android P at release date and later all versions of Android after April 2019,
+     * Firebase SDK is required to use push. Please follow the migration guide at https://aka.ms/acfba.
```
jwallra: This link shows how to remove firebase dependencies, not add them.  I don't know if we have documentation for this yet.

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -430,7 +429,10 @@ private synchronized void registerPushToken() {
             onTokenRefresh(FirebaseUtils.getToken());
             AppCenterLog.info(LOG_TAG, "Firebase SDK is available, using Firebase SDK registration.");
         } catch (FirebaseUtils.FirebaseUnavailableException e) {
-            AppCenterLog.info(LOG_TAG, "Firebase SDK is not available, using built in registration. cause: " + e.getMessage());
+            AppCenterLog.warn(LOG_TAG, "Firebase SDK is not available, using built in registration. " +
+                    "Starting Android P at release date and later all versions of Android after April 2019, " +
+                    "Firebase SDK is required to use push. Please follow the migration guide at https://aka.ms/acfba.\n" +
```
jwallra: Link only tells how to remove firebase dependency.

sdk/appcenter-push/src/test/java/com/microsoft/appcenter/push/PushTest.java
```diff
@@ -584,6 +584,7 @@ public void validateCheckLaunchedFromNotification() {
     public void registerWithoutFirebase() {
         IllegalStateException exception = new IllegalStateException();
         when(FirebaseInstanceId.getInstance()).thenThrow(exception);
+        //noinspection deprecation
```
jwallra: Add newlines before comments

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
MatkovIvan: See @guperrot's comment above https://github.com/Microsoft/AppCenter-SDK-Android/pull/744#discussion_r198592561

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
MatkovIvan: See @guperrot's comment above https://github.com/Microsoft/AppCenter-SDK-Android/pull/744#discussion_r198592561

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
MatkovIvan: @guperrot Should we do it with `noinspection` comments? I don't see a single approach to this in the project code

jwallra: Our convention is to have a newline before every comment.  We've had conversations about why this is/isn't a good rule before, but that's what we've chosen to do for the time being.

dhei: Looks like the convention in this repo is to *NOT* add a newline before compiler warning suppressions.

MatkovIvan: I'm aware of that. My question is: Does this rule apply to the service marks for warnings suppressing

guperrot: We add it when we don't forget but it was missed during some reviews, especially on unit tests. I don't usually nitpick on those `//` ones but if you can do the change that would be good I guess.

MatkovIvan: Will do

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/PushReceiver.java
```diff
@@ -38,6 +42,13 @@ public void onReceive(Context context, Intent intent) {
         /* Received message action. */
         else if (INTENT_ACTION_RECEIVE.equals(action)) {
             Push.getInstance().onMessageReceived(context, intent);
+
+            /* Prevent handling message by not initialized firebase. */
+            if (!FirebaseUtils.isFirebaseAvailable()) {
+                AppCenterLog.warn(LOG_TAG, "Abort \"" + INTENT_ACTION_RECEIVE + "\" broadcast " +
+                        "because firebase is not available.");
+                abortBroadcast();
```
guperrot: Does not work on my Pixel device with Android P I still have the crash.

Another way is to call       

```
PackageManager pm = getPackageManager();
        pm.setComponentEnabledSetting(new ComponentName(this, "com.google.firebase.messaging.FirebaseMessagingService"),
                PackageManager.COMPONENT_ENABLED_STATE_DISABLED, PackageManager.DONT_KILL_APP);
```

say in `Push.registerPushToken`.

Since the state is persisted, we should enable again it if we detect firebase available in that method, so that it works when they migrate.

sdk/appcenter-push/src/test/java/com/microsoft/appcenter/push/PushTest.java
```diff
@@ -184,7 +192,7 @@ public void setEnabled() throws InterruptedException {
         Push push = Push.getInstance();
         Channel channel = mock(Channel.class);
         when(mFirebaseInstanceId.getToken()).thenReturn(testToken);
-        start(mock(Context.class), push, channel);
+        start( push, channel);
```
jwallra: remove space before "push"

apps/sasquatch/build.gradle
```diff
@@ -75,21 +75,17 @@ dependencies {
     //noinspection GradleDependency
     jcenterDependencyImplementation "com.microsoft.appcenter:appcenter-push:${appCenterSdkVersion}"
 
-    firebaseImplementation "com.google.firebase:firebase-core:${rootProject.ext.firebaseLibVersion}"
-    firebaseImplementation "com.google.firebase:firebase-messaging:${rootProject.ext.firebaseLibVersion}"
+    firebaseImplementation "com.google.firebase:firebase-core:16.0.1"
```
guperrot: We can TODO remove this after SDK available in jCenter. Can also add jCenter to the directive for now.

MatkovIvan: I tried it before but in fact `gradle` cannot handle it for some reason :confused:
`> Could not find method jcenterDependencyFirebaseImplementation()`

guperrot: Let's just add a TODO then.

PR #746, opened 2018-06-28
Add nested transmission target testing in Sasquatch

guperrot:


apps/sasquatch/src/main/res/values/env.xml
```diff
@@ -9,9 +9,11 @@
     <string-array name="target_id_values" tools:ignore="InconsistentArrays,MissingTranslation">
         <item />
         <item>0eac643c981e48e8afeb29d9df5593e1-cb0ac72d-a4c4-41a3-b032-9f5ae7b64460-7166</item>
+        <item>4ee5ab442b354e6abebd166820bd8fef-88dbb064-f5f9-4dec-a428-d34062307511-8016</item>
     </string-array>
     <string-array name="target_id_names" tools:ignore="InconsistentArrays,MissingTranslation">
         <item>Default transmission</item>
-        <item>Aria - Sasquatch Android 2</item>
+        <item>Sasquatch Android 2</item>
```
jwallra: Would prefer to see the targets included still (Keep the `Aria - `)

PR #753, opened 2018-07-16
Use new SDK in demo app

MatkovIvan:


apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
```diff
@@ -128,9 +125,6 @@ protected void onCreate(Bundle savedInstanceState) {
             Distribute.setApiUrl(apiUrl);
         }
 
-        //noinspection deprecation
-        Push.setSenderId(SENDER_ID);
```
guperrot: Please keep that build variant so we can still test on people not following migrate guide.

PR #758, opened 2018-07-18
Override Common Schema Part A properties

jwallra:
This change adds a `PropertyConfigurator` to `AnalyticsTransmissionTarget`, which is used to override common schema properties.  Properties are inherited in the same way that event properties are.

### DIFF
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
@@ -1,51 +1,57 @@
 package com.microsoft.appcenter.analytics;
 
 import android.support.annotation.NonNull;
-import android.support.annotation.VisibleForTesting;
-import android.util.Property;
 
-import com.microsoft.appcenter.analytics.ingestion.models.one.CommonSchemaEventLog;
-import com.microsoft.appcenter.analytics.ingestion.models.one.json.CommonSchemaEventLogFactory;
 import com.microsoft.appcenter.channel.AbstractChannelListener;
 import com.microsoft.appcenter.channel.Channel;
-import com.microsoft.appcenter.channel.OneCollectorChannelListener;
-import com.microsoft.appcenter.ingestion.models.CustomPropertiesLog;
 import com.microsoft.appcenter.ingestion.models.Log;
-import com.microsoft.appcenter.ingestion.models.StartServiceLog;
-import com.microsoft.appcenter.ingestion.models.json.CustomPropertiesLogFactory;
-import com.microsoft.appcenter.ingestion.models.json.DefaultLogSerializer;
-import com.microsoft.appcenter.ingestion.models.json.LogSerializer;
-import com.microsoft.appcenter.ingestion.models.json.StartServiceLogFactory;
 import com.microsoft.appcenter.ingestion.models.one.AppExtension;
 import com.microsoft.appcenter.ingestion.models.one.CommonSchemaLog;
-import com.microsoft.appcenter.ingestion.models.one.SdkExtension;
-import com.microsoft.appcenter.utils.AppCenterLog;
-import com.microsoft.appcenter.utils.UUIDUtils;
-
-import java.util.Collection;
-
-import static com.microsoft.appcenter.utils.AppCenterLog.LOG_TAG;
 
 /**
  * Allow overriding Part A properties.
  */
 public class PropertyConfigurator extends AbstractChannelListener {
 
+    /**
+     * App name to override common schema part A 'app.name'.
+     */
     private String mAppName;
 
+    /**
+     * App name to override common schema part A 'app.ver'.
+     */
     private String mAppVersion;
 
+    /**
+     * App name to override common schema part A 'app.locale'.
+     */
     private String mAppLocale;
 
+    /**
+     * The transmission target which this configurator belongs to.
+     */
     private AnalyticsTransmissionTarget mTransmissionTarget;
 
+    /**
+     * Create a new property configurator.
+     *
+     * @param channel            The channel for this listener.
+     * @param transmissionTarget The tranmission target of the configurator.
+     */
     PropertyConfigurator(Channel channel, AnalyticsTransmissionTarget transmissionTarget) {
         mTransmissionTarget = transmissionTarget;
         if (channel != null) {
             channel.addListener(this);
         }
     }
-    
+
+    /**
+     * Override or inherit common schema properties while preparing log.
+     *
+     * @param log       A log.
+     * @param groupName The group name.
+     */
     @Override
     public void onPreparingLog(@NonNull Log log, @NonNull String groupName) {
         if (log instanceof CommonSchemaLog && mTransmissionTarget.isEnabled()) {
@@ -55,14 +61,12 @@ public class PropertyConfigurator extends AbstractChannelListener {
             if (mAppName != null) {
                 app.setId(mAppName);
             } else {
-                AnalyticsTransmissionTarget parent = mTransmissionTarget.mParentTarget;
-                while (parent != null) {
-                    String parentAppName = parent.getPropertyConfigurator().getAppName();
+                for (AnalyticsTransmissionTarget target = mTransmissionTarget.mParentTarget; target != null; target = target.mParentTarget) {
+                    String parentAppName = target.getPropertyConfigurator().getAppName();
                     if (parentAppName != null) {
                         app.setId(parentAppName);
                         break;
                     }
-                    parent = parent.mParentTarget;
                 }
             }
 
@@ -70,14 +74,12 @@ public class PropertyConfigurator extends AbstractChannelListener {
             if (mAppVersion != null) {
                 app.setVer(mAppVersion);
             } else {
-                AnalyticsTransmissionTarget parent = mTransmissionTarget.mParentTarget;
-                while (parent != null) {
-                    String parentAppVersion = parent.getPropertyConfigurator().getAppVersion();
+                for (AnalyticsTransmissionTarget target = mTransmissionTarget.mParentTarget; target != null; target = target.mParentTarget) {
+                    String parentAppVersion = target.getPropertyConfigurator().getAppVersion();
                     if (parentAppVersion != null) {
                         app.setVer(parentAppVersion);
                         break;
                     }
-                    parent = parent.mParentTarget;
                 }
             }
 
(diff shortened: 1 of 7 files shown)


sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTargetTest.java
```diff
@@ -436,4 +442,87 @@ public void eventPropertiesCascading() {
...
+        pc.onPreparingLog(log, "groupName");
+
+        /* Assert properties set on common schema. */
+        assertEquals("appVersion", log.getExt().getApp().getVer());
+        assertEquals("appName", log.getExt().getApp().getId());
+        assertEquals("appLocale", log.getExt().getApp().getLocale());
+    }
+
+    @Test
+    public void inheritCommonSchemaProperties() {
```
jwallra: This test should be removed because it's redundant with `inheritCommonSchemaPropertiesFromGrandparent`.  It misses a small bit of code coverage which is covered by the other test

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
```diff
@@ -0,0 +1,144 @@
...
+
+    PropertyConfigurator(Channel channel, AnalyticsTransmissionTarget transmissionTarget) {
+        mTransmissionTarget = transmissionTarget;
+        if (channel != null) {
+            channel.addListener(this);
+        }
+    }
+
+    /**
+     *
```
dhei: nit - remove the empty line; add description to `@param`.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.Collection;
+
+import static com.microsoft.appcenter.utils.AppCenterLog.LOG_TAG;
+
+/**
+ * Allow overriding Part A properties.
+ */
+public class PropertyConfigurator extends AbstractChannelListener {
+
+    private String mAppName;
```
guperrot: Missing Javadoc comment for fields and constructor

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
```diff
@@ -0,0 +1,144 @@
...
+    PropertyConfigurator(Channel channel, AnalyticsTransmissionTarget transmissionTarget) {
+        mTransmissionTarget = transmissionTarget;
+        if (channel != null) {
+            channel.addListener(this);
+        }
+    }
+
+    /**
+     *
+     * @param log
```
guperrot: Missing javadoc comment for parameters, only the key is present

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * Override common schema Part A property App.Name.
+     *
+     * @param appName   App name
+     */
+    public void setAppName(String appName) {
+        mAppName = appName;
+    }
+
+    protected String getAppName() {
```
guperrot: Missing Javadoc comments

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTargetTest.java
```diff
@@ -436,4 +442,87 @@ public void eventPropertiesCascading() {
         expectedProperties.put("g", "7777");
         assertEquals(expectedProperties, log.getProperties());
     }
+
```
guperrot: Class becomes big, we should move the new test to a PropertyConfiguratorTest

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTargetTest.java
```diff
@@ -436,4 +442,87 @@ public void eventPropertiesCascading() {
...
+    @Test
+    public void commonSchemaPropertiesNotSetWhenDisabled() {
+        CommonSchemaLog log = new CommonSchemaEventLog();
+        log.setExt(new Extensions());
+        log.getExt().setApp(new AppExtension());
+
+        /* Get target, disable it, and set properties. */
+        AnalyticsTransmissionTarget target = Analytics.getTransmissionTarget("test");
+        target.setEnabledAsync(false);
```
guperrot: add .get() to each setEnabledAsync calls to make sure its synced (to wait on the future task to complete)

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
```diff
@@ -0,0 +1,144 @@
...
+                    }
+                    parent = parent.mParentTarget;
+                }
+            }
+
+            /* Override app version if not null, else use the version of the nearest parent. */
+            if (mAppVersion != null) {
+                app.setVer(mAppVersion);
+            } else {
+                AnalyticsTransmissionTarget parent = mTransmissionTarget.mParentTarget;
```
guperrot: you can use for loop like in Analytics transmission target

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
```diff
@@ -0,0 +1,144 @@
...
+                    parent = parent.mParentTarget;
+                }
+            }
+        }
+    }
+
+    /**
+     * Override common schema Part A property App.Name.
+     *
+     * @param appName   App name
```
dhei: nit - remove the extra white space

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTargetTest.java
```diff
@@ -436,4 +442,87 @@ public void eventPropertiesCascading() {
         expectedProperties.put("g", "7777");
         assertEquals(expectedProperties, log.getProperties());
     }
+
+    /* Property Configurator tests */
+    @Test
+    public void setCommonSchemaProperties() {
```
guperrot: You need a test where child has no parent and also no overrides to fix code coverage on the while loops (in that the common schema log is not altered and thus checking for null values)

guperrot: Ok you did something equivalent in the meantime while I was typing this.

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTargetTest.java
```diff
@@ -436,4 +443,98 @@ public void eventPropertiesCascading() {
...
+        Log log = new EventLog();
+
+        /* Get property configurator and set properties. */
+        PropertyConfigurator pc = Analytics.getTransmissionTarget("test").getPropertyConfigurator();
+        pc.setAppVersion("appVersion");
+        pc.setAppName("appName");
+        pc.setAppLocale("appLocale");
+        pc.onPreparingLog(log, "groupName");
+
+        verify(mChannel, never()).enqueue(isA(EventLog.class), anyString());
```
guperrot: there is no enqueue involved here, you can verifyZeroInteractions(log) instead

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
jwallra: I had meant to remove this.  It's not listed in the One Collector listener so I'll follow that pattern.

guperrot: We usually never override Javadoc (`@Override` methods) unless the meaning actually changes. by default Javadoc is inherited.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
guperrot: dots in parameter names and make sure to commit using Android studio dialog actions, it will decide which spacing to use with the reformat action.

jwallra: This is how auto-reformat formats the code so I'll leave it

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
jwallra: Will update for consistency

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
```diff
@@ -0,0 +1,154 @@
...
+            AppExtension app = ((CommonSchemaLog) log).getExt().getApp();
+
+            /* Override app name if not null, else use the name of the nearest parent. */
+            if (mAppName != null) {
+                app.setId(mAppName);
+            } else {
+                for (AnalyticsTransmissionTarget target = mTransmissionTarget.mParentTarget; target != null; target = target.mParentTarget) {
+                    String parentAppName = target.getPropertyConfigurator().getAppName();
+                    if (parentAppName != null) {
+                        app.setId(parentAppName);
```
jaelim-ms: This is `app.id`, not `app.name`.

guperrot: You mean the opposite? It should call `setName`

jwallra: Yeah I added those

jaelim-ms: I meant the implementation is overriding `app.id`, not `app.name` and it should be `app.name`.
Sorry for confusion but I think we are on the same page. 😄

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/one/AppExtension.java
```diff
@@ -60,6 +70,24 @@ public void setId(String id) {
         this.id = id;
     }
 
+    /**
```
guperrot: As one collector recommends, we keep ordered JSON as in https://osgwiki.com/wiki/Common_Schema_3.0 please move name everywhere (methods, within methods for writing/reading json) between ver and locale.

Please also modify the unit tests for code coverage and regenerate equals/hashcode after moving the field to the right position.

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/one/AppExtension.java
```diff
@@ -100,13 +128,15 @@ public void setLocale(String locale) {
...
         setId(object.optString(ID, null));
+        setName(object.optString(NAME, null));
         setVer(object.optString(VER, null));
         setLocale(object.optString(LOCALE, null));
     }
 
     @Override
     public void write(JSONStringer writer) throws JSONException {
         JSONUtils.write(writer, ID, getId());
+        JSONUtils.write(writer, NAME, getName());
```
guperrot: order here too

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/one/AppExtension.java
```diff
@@ -100,13 +128,15 @@ public void setLocale(String locale) {
     @Override
     public void read(JSONObject object) {
         setId(object.optString(ID, null));
+        setName(object.optString(NAME, null));
```
guperrot: order here too

jaelim-ms:
@guperrot checked One Collector spec and it allows empty app name. @jwallra ignore my comment.

PR #761, opened 2018-07-20
Moving set/removeEventProperty to PropertyConfigurator

jwallra:
Since they behave generally the same way as other properties, they are being moved to the `PropertyConfigurator`.

### DIFF
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventPropertiesActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventPropertiesActivity.java
@@ -19,11 +19,11 @@ import android.widget.Spinner;
 import android.widget.TextView;
 
 import com.microsoft.appcenter.analytics.AnalyticsTransmissionTarget;
-import com.microsoft.appcenter.analytics.PropertyConfigurator;
 import com.microsoft.appcenter.sasquatch.R;
 import com.microsoft.appcenter.sasquatch.util.EventActivityUtil;
 
 import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


jaelim-ms:
You have to use reflection to access property configurator. Look at build failure and fix it.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
```diff
@@ -151,4 +159,39 @@ private String getAppLocale() {
...
+    @SuppressWarnings("WeakerAccess")
+    public synchronized void removeEventProperty(String key) {
+        mEventProperties.remove(key);
+    }
+
+    /**
+     * Extracted method to synchronize on each level at once while reading properties.
+     * Nesting synchronize between parent/child could lead to deadlocks.
+     */
+    protected synchronized void mergeEventProperties(Map<String, String> mergedProperties) {
```
guperrot: can you explain why private does not work? Who uses the method besides internally?

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventPropertiesActivity.java
```diff
@@ -87,52 +88,76 @@ public boolean onCreateOptionsMenu(Menu menu) {
...
-                mAddPropertyLayout.findViewById(R.id.add_button).setOnClickListener(new View.OnClickListener() {
-
-                    @Override
-                    public void onClick(View v) {
-                        CharSequence key = keyView.getText();
-                        CharSequence value = valueView.getText();
-                        if (!TextUtils.isEmpty(key) && !TextUtils.isEmpty(value)) {
-                            getSelectedTarget().setEventProperty(key.toString(), value.toString());
-                            updatePropertyList();
+        AnalyticsTransmissionTarget target = getSelectedTarget();
```
guperrot: Add todo comment to remove reflection after release.

Maybe select the method as it was before for jCenter? Does the app run without crashing on jCenter?
An alternative to reflection is to have the 2 files in 2 build variants (2 versions of the file), we used that in the past when reflection was becoming too complex.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
jwallra: It's used in `trackEvent` from `AnalyticsTransmissionTarget` when we're iterating through parents and merging event properties: https://github.com/Microsoft/AppCenter-SDK-Android/blob/develop/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java#L91

guperrot: You can remove protected then and use no keyword

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventPropertiesActivity.java
jwallra: Using reflection was a bit of a pain here bit since it's already being used I think it's OK to continue the pattern.  As long as we keep on top of removing it we should be fine.  Is this already documented in our release process?

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/PropertyConfiguratorTest.java
```diff
@@ -127,4 +138,152 @@ public void appCenterLogDoesNotOverride() {
         /* Verify no interactions with the log. */
         verifyNoMoreInteractions(log);
     }
+
+
```
jaelim-ms: Remove empty line.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventPropertiesActivity.java
guperrot: It is documented yes, but also mentions to check for files in multiple build variants as this is also a pattern we used in the past.

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/PropertyConfiguratorTest.java
dhei: nit - remove extra new line

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
```diff
@@ -151,4 +159,39 @@ private String getAppLocale() {
...
+     * Removes the given key from the common event properties.
+     *
+     * @param key The property key to be removed.
+     */
+    @SuppressWarnings("WeakerAccess")
+    public synchronized void removeEventProperty(String key) {
+        mEventProperties.remove(key);
+    }
+
+    /**
```
jaelim-ms: This is not a javadoc. Use `/* */` comment style or adding a full javadoc with discussion

PR #766, opened 2018-07-24
Override cs properties in test app and fix logic

guperrot:


apps/sasquatch/src/projectDependency/java/com/microsoft/appcenter/sasquatch/activities/CommonSchemaPropertiesActivity.java
```diff
@@ -0,0 +1,117 @@
...
+    static final boolean IS_SUPPORTED = true;
+
+    private Spinner mCommonSchemaPropertiesSpinner;
+
+    private EditText mCommonSchemaPropertyValue;
+
+    private PropertyConfigurator mPropertyConfigurator;
+
+    private enum PropertyName {
+
```
jwallra: remove extra line

PR #771, opened 2018-08-07
One Collector MSA token support for internal customers

guperrot:


sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/TicketCache.java
```diff
@@ -0,0 +1,30 @@
+package com.microsoft.appcenter.utils;
+
+import android.support.annotation.VisibleForTesting;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class TicketCache {
```
jaelim-ms: Can't you make all methods static and remove `getInstance`?

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AuthenticationProvider.java
```diff
@@ -0,0 +1,170 @@
...
+         * @param ticketKey The ticket key that is used to get an updated token.
+         * @param callback  callback to provide the result.
+         */
+        void getToken(String ticketKey, AuthenticationCallback callback);
+
+    }
+
+    public interface AuthenticationCallback {
+
+        void onAuthenticationResult(String tokenValue, Date expiresAt);
```
jaelim-ms: Can you add a doc here?

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
```diff
@@ -192,12 +236,27 @@ String getTransmissionTargetToken() {
...
+
+            @Override
+            public void onPreparingLog(@NonNull Log log, @NonNull String groupName) {
+                addTicketToLog(log);
+            }
+        };
+    }
+
+    /**
+     * Add ticket to common schema logs.
```
jwallra: What is a ticket again, exactly?  Is "ticket" universal terminology that is used for other authentication frameworks?  If not, it's worth clarifying in comments.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/TicketCache.java
guperrot: Tha's not the pattern we usually use.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AuthenticationProvider.java
```diff
@@ -0,0 +1,170 @@
...
+import com.microsoft.appcenter.utils.TicketCache;
+
+import java.util.Date;
+
+import static com.microsoft.appcenter.analytics.Analytics.LOG_TAG;
+
+public class AuthenticationProvider {
+
+    /**
+     * Ratio threshold of expiry time to refresh token a bit before it's expired.
```
jwallra: Current wording doesn't really make sense... I would go with "Ratio of refresh interval to token lifetime.  Used to refresh token a bit before it's expired."

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AuthenticationProvider.java
```diff
@@ -0,0 +1,170 @@
...
+     */
+    private static final double REFRESH_THRESHOLD = 0.9;
+
+    /**
+     * The type.
+     */
+    private final Type mType;
+
+    /**
+     * The ticket key for this authentication provider.
```
jwallra: This would be a good place to describe what a ticket is if needed.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AuthenticationProvider.java
```diff
@@ -0,0 +1,170 @@
...
+
+public class AuthenticationProvider {
+
+    /**
+     * Ratio threshold of expiry time to refresh token a bit before it's expired.
+     */
+    private static final double REFRESH_THRESHOLD = 0.9;
+
+    /**
+     * The type.
```
jwallra: "The type *of authentication provider*", yes?

jwallra: I would even just rename this to `mProviderType`

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/OneCollectorIngestion.java
```diff
@@ -117,6 +134,35 @@ public ServiceCall sendAsync(String appSecret, UUID installId, LogContainer logC
...
+        /* Gather tokens from logs. */
+        JSONObject tickets = new JSONObject();
+        for (Log log : logContainer.getLogs()) {
+            List<String> ticketKeys = ((CommonSchemaLog) log).getExt().getProtocol().getTicketKeys();
+            if (ticketKeys != null) {
+                for (String ticketKey : ticketKeys) {
+                    String token = TicketCache.getInstance().getTicket(ticketKey);
+                    if (token != null) {
+                        try {
+                            tickets.put(ticketKey, "d:" + token);
```
jwallra: Why do we add the `d:` prefix to the token?

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/TicketCache.java
jwallra: This seems very generic, a cache of a string-string mapping... Could we need this in other areas in the future?  Should it be refactored as a generic cache?  Just a thought.

guperrot: It should contain only tickets, nothing else to avoid colliding.

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/OneCollectorIngestion.java
guperrot: It's from email discussion, the protocol documents `"p:"` instead but seems like bearer token is actually `"d:"`. There are prefixes for each type.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AuthenticationProvider.java
guperrot: It's an inner enum already, not a top level one. The benefit of having inner types is to avoid being redundant.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
guperrot: This is the right word in One Collector world.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AuthenticationProvider.java
guperrot: No the right place to add details is the contructor. The field is private. Will add details to constuctor.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AuthenticationProvider.java
guperrot: Updated

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AuthenticationProvider.java
guperrot: Done

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AuthenticationProvider.java
```diff
@@ -0,0 +1,199 @@
...
+        }
+    }
+
+    /**
+     * The supported types of the authentication provider.
+     */
+    public enum Type {
+
+        /**
+         * Microsoft authentication.
```
dhei: nit - "Microsoft account authentication"

Technically "MSA" stands for "Microsoft Account", see https://en.wikipedia.org/wiki/Microsoft_account.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AuthenticationProvider.java
```diff
@@ -0,0 +1,199 @@
...
+
+        /**
+         * Implement this method and pass a fresh authentication token using the callback.
+         * This will be called 1 time right after registering and also whenever the token is about to expire.
+         *
+         * @param ticketKey The ticket key that is used to get an updated token.
+         * @param callback  callback to provide the result.
+         */
+        void getToken(String ticketKey, AuthenticationCallback callback);
+
```
dhei: nit - remove new line

apps/sasquatch/src/projectDependency/java/com/microsoft/appcenter/sasquatch/activities/MSALoginActivity.java
```diff
@@ -0,0 +1,299 @@
...
+        }
+    }
+
+    /**
+     * Get initial access token.
+     */
+    private void getToken(final String code) {
+        Map<String, String> headers = new HashMap<>();
+        headers.put(DefaultHttpClient.CONTENT_TYPE_KEY, "application/x-www-form-urlencoded");
+        sHttpClient.callAsync("https://login.live.com/oauth20_token.srf",
```
dhei: nit - use `TOKEN_URL` defined at the top of the class

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
```diff
@@ -61,6 +72,34 @@
...
 
+    /**
+     * Add an authentication provider to associate logs with user identifiers.
+     *
+     * @param authenticationProvider The authentication provider.
+     */
+    public static synchronized void addAuthenticationProvider(AuthenticationProvider authenticationProvider) {
+
+        /* Validate input. */
+        if (authenticationProvider.getType() == null) {
```
achocron: For my understanding, what does it mean for the type to be `null`? Why not just check if the object is `null` here?

guperrot: The object is user input so it can be null by mistake.

achocron: Oh got it. I was for some reason thinking that getType was like `typeof`. Either way, null check should be first (and now is)

achocron: Also why do we use type fields instead of inheritance? Not that we need to change it at this point

guperrot: There is no code specific to type, it will only be used later to have 1 provider per type allowed.

PR #788, opened 2018-08-25
Use 1 client ID even for compact MSA

guperrot:


apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MSALoginActivity.java
```diff
@@ -56,9 +56,7 @@
 
     private static final String SIGN_OUT_URL = URL_PREFIX + "logout.srf?";
 
-    private static final String CLIENT_ID_COMPACT = "000000004C1D3F6C";
-
-    private static final String CLIENT_ID_DELEGATE = "06181c2a-2403-437f-a490-9bcb06f85281";
+    private static final String CLIENT_ID = "06181c2a-2403-437f-a490-9bcb06f85281";
```
jaelim-ms: We need to move this to XML later.

PR #793, opened 2018-09-06
 Add UI for device id to test application

MatkovIvan:
screenshot_1536648323

### DIFF
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/DeviceInfoActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/DeviceInfoActivity.java
@@ -1,6 +1,8 @@
 package com.microsoft.appcenter.sasquatch.activities;
 
+import android.annotation.SuppressLint;
 import android.os.Bundle;
+import android.provider.Settings;
 import android.support.annotation.NonNull;
 import android.support.annotation.VisibleForTesting;
 import android.support.v7.app.AppCompatActivity;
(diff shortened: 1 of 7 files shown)


apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventActivity.java
```diff
@@ -123,22 +127,58 @@ public void toggleTransmissionEnabled(View view) {
...
+
+            // TODO remove reflection once new APIs available in jCenter.
+            {
+                try {
+                    PropertyConfigurator.class.getMethod("collectDeviceId");
+                } catch (Exception ignored) {
+                    return;
+                }
             }
+            mDeviceIdEnabledCheckBox.setVisibility(View.VISIBLE);
```
jaelim-ms: It shouldn't reset settings to disabled. `collectDeviceId` can only be called once before `start`, there is no disable or reset device Id functionalities.

jaelim-ms: This isn't preventing from unchecking. It is enabling the checkbox.
If you enable device on a transmission target and come back to the target after switching to another target, it will be "check-able" which shouldn't be.

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/one/PartAUtils.java
```diff
@@ -92,5 +92,8 @@ public static void addPartAFromLog(Log src, CommonSchemaLog dest, String transmi
                 Math.abs(device.getTimeZoneOffset() / 60),
                 Math.abs(device.getTimeZoneOffset() % 60));
         dest.getExt().getLoc().setTz(timezoneOffset);
+
+        /* Add device extension. */
+        dest.getExt().setDevice(new DeviceExtension());
```
guperrot: Please reflect that change in unit tests in some way, like a assert/verify statement.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/DeviceInfoActivity.java
```diff
@@ -50,7 +52,12 @@ protected void onCreate(Bundle savedInstanceState) {
         }
 
         final List<DeviceInfoDisplayModel> list = getDeviceInfoDisplayModelList(log);
-
+        @SuppressLint("HardwareIds")
```
guperrot: Please add a blank line before annotations.

guperrot:
UI works as expected now.

jwallra:
Please add screenshots

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/one/PartAUtils.java
MatkovIvan: Done. This was supposed to be part #794 :wink:

PR #794, opened 2018-09-07
Allow collection of device ID from property configurator

jwallra:
Device ID will not be inherited by children.

### DIFF
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
@@ -316,7 +316,8 @@ public class Analytics extends AbstractAppCenterService {
                 AppCenterLog.debug(LOG_TAG, "Returning transmission target found with token " + transmissionTargetToken);
                 return transmissionTarget;
             }
-            transmissionTarget = new AnalyticsTransmissionTarget(transmissionTargetToken, null, mChannel);
+            transmissionTarget = new AnalyticsTransmissionTarget(transmissionTargetToken, null,
+                    mCurrentActivity.get().getBaseContext(), mChannel);
             AppCenterLog.debug(LOG_TAG, "Created transmission target with token " + transmissionTargetToken);
             mTransmissionTargets.put(transmissionTargetToken, transmissionTarget);
             return transmissionTarget;
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
@@ -317,7 +317,7 @@ public class Analytics extends AbstractAppCenterService {
             return null;
         } else if (mContext == null) {
 
-            // Context and channel have the same lifecycle so we only have to check one.
+            /* Context and channel have the same lifecycle so we only have to check one. */
             AppCenterLog.error(LOG_TAG, "App context is null, App Center has not been started.");
             return null;
         } else {
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
@@ -81,6 +81,11 @@ public class Analytics extends AbstractAppCenterService {
     private WeakReference<Activity> mCurrentActivity;
 
     /**
+     * Application context, if not null it means onStart was called.
+     */
+    private Context mContext;
+
+    /**
      * True if started from app, false if started only from a library or not yet started at all.
      */
     private boolean mStartedFromApp;
@@ -316,8 +321,7 @@ public class Analytics extends AbstractAppCenterService {
                 AppCenterLog.debug(LOG_TAG, "Returning transmission target found with token " + transmissionTargetToken);
                 return transmissionTarget;
             }
-            transmissionTarget = new AnalyticsTransmissionTarget(transmissionTargetToken, null,
-                    mCurrentActivity.get().getBaseContext(), mChannel);
+            transmissionTarget = new AnalyticsTransmissionTarget(transmissionTargetToken, null, mContext, mChannel);
             AppCenterLog.debug(LOG_TAG, "Created transmission target with token " + transmissionTargetToken);
             mTransmissionTargets.put(transmissionTargetToken, transmissionTarget);
             return transmissionTarget;
@@ -596,6 +600,7 @@ public class Analytics extends AbstractAppCenterService {
 
     @Override
     public synchronized void onStarted(@NonNull Context context, @NonNull Channel channel, String appSecret, String transmissionTargetToken, boolean startedFromApp) {
+        mContext = context;
         mStartedFromApp = startedFromApp;
         super.onStarted(context, channel, appSecret, transmissionTargetToken, startedFromApp);
         setDefaultTransmissionTarget(transmissionTargetToken);
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
@@ -313,7 +313,12 @@ public class Analytics extends AbstractAppCenterService {
      */
     private synchronized AnalyticsTransmissionTarget getInstanceTransmissionTarget(String transmissionTargetToken) {
         if (transmissionTargetToken == null || transmissionTargetToken.isEmpty()) {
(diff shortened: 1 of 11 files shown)


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -308,15 +313,20 @@ private static String generatePageName(Class<?> activityClass) {
      */
     private synchronized AnalyticsTransmissionTarget getInstanceTransmissionTarget(String transmissionTargetToken) {
         if (transmissionTargetToken == null || transmissionTargetToken.isEmpty()) {
-            AppCenterLog.error(LOG_TAG, "Transmission target may not be null or empty.");
+            AppCenterLog.error(LOG_TAG, "Transmission target token may not be null or empty.");
+            return null;
+        } else if (mContext == null) {
+
+            // Context and channel have the same lifecycle so we only have to check one.
```
guperrot: We use `/*` comments when its not generated by Android Studio to ignore warnings.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
```diff
@@ -58,17 +59,23 @@
...
+     * App context.
+     */
+    private final Context mContext;
+
     /**
      * Create a new instance.
      *
      * @param transmissionTargetToken The token for this transmission target.
      * @param parentTarget            Parent transmission target.
      * @param channel                 The channel for this transmission target.
```
guperrot: Update Javadoc

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
```diff
@@ -225,6 +232,15 @@ public void run() {
         return future;
     }
 
+    /**
+     * Getter for base context.
+     *
+     * @return the context.
+     */
+    Context getContext() {
```
guperrot: We are passing channel to PropertyConfigurator and not using a getter, please also pass Context to avoid this and make it consistent with channel field.

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTest.java
```diff
@@ -87,14 +93,9 @@ public void notInit() {
         Analytics.trackEvent("test", new HashMap<String, String>());
         Analytics.trackPage("test");
         Analytics.trackPage("test", new HashMap<String, String>());
```
guperrot: Why removing those lines? The code should just simulate onStart with context before to restore those singleton tests.

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/ingestion/models/one/CommonSchemaLogSerializerTest.java
```diff
@@ -108,6 +108,12 @@ public void serializeAndDeserialize() throws JSONException {
         log.getExt().getLoc().setTz("-08:00");
         checkSerialization(serializer, log);
 
+        /* Device extension. */
```
guperrot: We opted to keep same JSON order as bond as One Collector team some Asimov customers are sensitive to JSON order of fields.

Device extension is located between user and os extensions in Bond.

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/one/DeviceExtension.java
```diff
@@ -0,0 +1,67 @@
...
+        setLocalId(object.optString(LOCAL_ID, null));
+    }
+
+    @Override
+    public void write(JSONStringer writer) throws JSONException {
+        JSONUtils.write(writer, LOCAL_ID, getLocalId());
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
guperrot: Do you have code analysis warning simplifiable if statement?

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/one/Extensions.java
```diff
@@ -46,6 +46,11 @@
      */
     private static final String LOC = "loc";
 
+    /**
```
guperrot: We opted to keep same JSON order as bond as One Collector team some Asimov customers are sensitive to JSON order of fields.

Device extension is located between user and os extensions in Bond.

guperrot: (Should reorder everything, the constant, the getter/setter etc... though what really matters it the order of the write method)

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/one/Extensions.java
```diff
@@ -328,7 +370,8 @@ public boolean equals(Object o) {
         if (app != null ? !app.equals(that.app) : that.app != null) return false;
         if (net != null ? !net.equals(that.net) : that.net != null) return false;
         if (sdk != null ? !sdk.equals(that.sdk) : that.sdk != null) return false;
-        return loc != null ? loc.equals(that.loc) : that.loc == null;
+        if (loc != null ? !loc.equals(that.loc) : that.loc != null) return false;
```
guperrot: regenerate equals/hashcode after reordering

sdk/appcenter/src/test/java/com/microsoft/appcenter/ingestion/models/one/ExtensionsTest.java
```diff
@@ -91,5 +91,15 @@ public void equalsHashCode() {
         checkNotEquals(a, b);
         b.setLoc(a.getLoc());
         checkEquals(a, b);
+
```
guperrot: Move between user and os after reordering

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
```diff
@@ -16,21 +19,31 @@
  */
 public class PropertyConfigurator extends AbstractChannelListener {
 
+    /**
+     * Common schema prefix for Android device IDs.
+     */
+    private final String ANDROID_DEVICE_ID_PREFIX = "a:";
```
jaelim-ms: private **static** final

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/one/Extensions.java
jaelim-ms: We should keep in mind that generated methods like `hashCode` and `equals` shouldn't be manually modified. Remove those methods and regenerate.

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/one/DeviceExtension.java
jaelim-ms: Not for this one. It has nothing to simplify.

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTest.java
jaelim-ms: This isn't valid since this PR also fixes the bug that you mentioned. If it isn't initialized then `getTransmissionTarget` will return `null` which is covered by another unit test.

guperrot: I was talking about initializing.
This is the test that checks we always get same instance, I don't see any other test doing that, please add the onStart mock and restore the tests.

guperrot: Oh Github UI tricked me into thinking it was the singleton tests while its another test...

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
jaelim-ms: [proposal.txt](https://github.com/Microsoft/AppCenter-SDK-Android/files/2362695/proposal.txt)

guperrot: That would also work indeed.

PR #796, opened 2018-09-15
Set max storage size

dhei:
Adding setMaxStorageSize API to core module with the following behaviors:

- Set the storage size to a given amount, in bytes, rounded up to the nearest 4KB.
- Return value is "true" if the storage size was set to the next multiple of 4kB or false otherwise.
- If the number of logs currently stored is already larger than the new storage size, the size won't be set.
- The minimum storage size allowed is 20KB (20480 bytes) (SQL limitation).
- When new logs are inserted while the storage is full, the oldest logs will be discarded until the new log fits.
- If a log is larger than the entire storage size, all logs will be removed from storage, and the new log will be discarded.  This decision was made because calculating the size of individual logs already in storage is a very expensive operation.

### DIFF
--- a/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
+++ b/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
@@ -233,7 +233,7 @@ public class DatabasePersistenceAndroidTest {
     public void putLargeLogFailsToRead() throws PersistenceException {
 
         /* Initialize database persistence. */
-        DatabasePersistence persistence = new DatabasePersistence(sContext, 1, DatabasePersistence.SCHEMA, Persistence.DEFAULT_MAX_STORAGE_SIZE_IN_BYTES);
+        DatabasePersistence persistence = new DatabasePersistence(sContext, 1, DatabasePersistence.SCHEMA, DatabasePersistence.DEFAULT_MAX_STORAGE_SIZE_IN_BYTES);
 
         /* Set a mock log serializer. */
         LogSerializer logSerializer = new DefaultLogSerializer();
@@ -342,7 +342,7 @@ public class DatabasePersistenceAndroidTest {
     public void deleteLogs() throws PersistenceException {
 
         /* Initialize database persistence. */
-        DatabasePersistence persistence = new DatabasePersistence(sContext, 1, DatabasePersistence.SCHEMA, Persistence.DEFAULT_MAX_STORAGE_SIZE_IN_BYTES);
+        DatabasePersistence persistence = new DatabasePersistence(sContext, 1, DatabasePersistence.SCHEMA, DatabasePersistence.DEFAULT_MAX_STORAGE_SIZE_IN_BYTES);
 
         /* Set a mock log serializer. */
         LogSerializer logSerializer = new DefaultLogSerializer();
--- a/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
+++ b/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
@@ -233,7 +233,7 @@ public class DatabasePersistenceAndroidTest {
     public void putLargeLogFailsToRead() throws PersistenceException {
 
         /* Initialize database persistence. */
-        DatabasePersistence persistence = new DatabasePersistence(sContext, 1, DatabasePersistence.SCHEMA, DatabasePersistence.DEFAULT_MAX_STORAGE_SIZE_IN_BYTES);
+        DatabasePersistence persistence = new DatabasePersistence(sContext, 1, DatabasePersistence.SCHEMA);
 
         /* Set a mock log serializer. */
         LogSerializer logSerializer = new DefaultLogSerializer();
@@ -283,7 +283,8 @@ public class DatabasePersistenceAndroidTest {
     public void putTooManyLogs() throws PersistenceException {
 
         /* Initialize database persistence. */
-        DatabasePersistence persistence = new DatabasePersistence(sContext, 2, DatabasePersistence.SCHEMA, MAX_STORAGE_SIZE_IN_BYTES);
+        DatabasePersistence persistence = new DatabasePersistence(sContext, 2, DatabasePersistence.SCHEMA);
+        assertTrue(persistence.setMaxStorageSize(MAX_STORAGE_SIZE_IN_BYTES));
 
         /* Set a mock log serializer. */
         LogSerializer logSerializer = new DefaultLogSerializer();
@@ -346,7 +347,7 @@ public class DatabasePersistenceAndroidTest {
     public void deleteLogs() throws PersistenceException {
 
         /* Initialize database persistence. */
-        DatabasePersistence persistence = new DatabasePersistence(sContext, 1, DatabasePersistence.SCHEMA, DatabasePersistence.DEFAULT_MAX_STORAGE_SIZE_IN_BYTES);
+        DatabasePersistence persistence = new DatabasePersistence(sContext, 1, DatabasePersistence.SCHEMA);
 
         /* Set a mock log serializer. */
         LogSerializer logSerializer = new DefaultLogSerializer();
--- a/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
+++ b/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
@@ -62,7 +62,7 @@ public class DatabasePersistenceAndroidTest {
     /**
      * Maximum storage size in bytes for unit test case.
      */
-    private static final int MAX_STORAGE_SIZE_IN_BYTES = 16385;
+    private static final int MAX_STORAGE_SIZE_IN_BYTES = 20480;
 
     /**
      * Context instance.
--- a/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
+++ b/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
@@ -88,6 +88,13 @@ public class DatabasePersistenceAndroidTest {
         sContext.deleteDatabase(DatabasePersistence.DATABASE);
     }
 
+    @After
+    public void tearDown() {
+
+        /* Clean up database. */
+        sContext.deleteDatabase("test-persistence");
+    }
+
     private static int getIteratorSize(Iterator iterator) {
         int count = 0;
         for (; iterator.hasNext(); iterator.next())
@@ -124,8 +131,6 @@ public class DatabasePersistenceAndroidTest {
             assertEquals(log, outputLogs.get(0));
             assertEquals(1, persistence.countLogs("test-p1"));
         } finally {
-
-            /* Close. */
             persistence.close();
         }
     }
@@ -182,8 +187,6 @@ public class DatabasePersistenceAndroidTest {
             assertFalse(file.exists());
             assertFalse(file.getParentFile().exists());
         } finally {
-
-            /* Close. */
             persistence.close();
         }
     }
@@ -223,8 +226,6 @@ public class DatabasePersistenceAndroidTest {
             /* Make sure database entry has been removed. */
             assertEquals(0, persistence.countLogs("test-p1"));
         } finally {
-
-            /* Close. */
             persistence.close();
 
             /* Restore path. */
@@ -276,8 +277,6 @@ public class DatabasePersistenceAndroidTest {
             assertEquals(0, outputLogs.size());
             assertEquals(0, persistence.countLogs("test-p1"));
         } finally {
-
-            /* Close. */
             persistence.close();
         }
     }
@@ -408,7 +407,6 @@ public class DatabasePersistenceAndroidTest {
             assertEquals(0, persistence.countLogs("test-p1"));
         } finally {
 
-            /* Close. */
             //noinspection ThrowFromFinallyBlock
             persistence.close();
         }
@@ -430,8 +428,6 @@ public class DatabasePersistenceAndroidTest {
             Log log = AndroidTestUtils.generateMockLog();
             persistence.putLog("test-p1", log);
         } finally {
-
-            /* Close. */
             persistence.close();
         }
     }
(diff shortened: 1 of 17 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -84,6 +89,7 @@
      */
     private long mIMDBAutoInc;
 
+
```
MatkovIvan: Extra line

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -53,6 +53,18 @@
...
+     * Default maximum storage size for SQLite database.
+     */
+    @VisibleForTesting
+    static final long DEFAULT_MAX_STORAGE_SIZE_IN_BYTES = 10 * 1024 * 1024;
+
+    /**
+     * Minimum size allowed for set maximum size.
+     */
+    @VisibleForTesting
+    static final long MINIMUM_STORAGE_SIZE = 16385;
```
achocron: Where does this number come from? It's *slightly* over 4 pages. So I suggest either 4 or 5 pages should be used for this

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -365,7 +387,23 @@ public static void startFromLibrary(Context context, Class<? extends AppCenterSe
...
+     * the new size, then the operation will fail and a warning will be emitted. Can only be called
+     * once per app lifetime and only before AppCenter.start(...).
+     * <p>
+     * If the size is not a multiple of 4096 bytes, the next multiple of 4096 is used as the new maximum size.
+     *
+     * @param storageSizeInBytes New size for the SQLite db in bytes.
+     * @return Future with true result if succeeded, otherwise future with false result.
+     */
+    @SuppressWarnings("WeakerAccess") // TODO remove annotation when updating demo app for release.
+    public static AppCenterFuture<Boolean> setMaxStorageSize(long storageSizeInBytes) {
```
achocron: Doesn't the spec indicate that we need to use a callback?

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -537,6 +510,45 @@ void setSQLiteOpenHelper(@NonNull SQLiteOpenHelper helper) {
...
+     *
+     * @param maxStorageSizeInBytes Maximum SQLite database size.
+     * @return true if database size was set, otherwise false.
+     */
+    boolean setMaxStorageSize(long maxStorageSizeInBytes) {
+        SQLiteDatabase db = getDatabase();
+        long newMaxSize = db.setMaximumSize(maxStorageSizeInBytes);
+
+        /* SQLite always use the next multiple of 4KB as maximum size. */
+        long expectedMultipleMaxSize = maxStorageSizeInBytes / ALLOWED_SIZE_MULTIPLE * ALLOWED_SIZE_MULTIPLE;
```
achocron: This is a little confusing, can a simple ceiling function do the job too?

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
achocron: But also, shouldn't it be higher than 2MiB?

jwallra: Our decision was to allow setting any size; 2MB is just the max for common schema logs.  There is an undocumented exception that occurs when you set it to 16384 or lower, though, so it's to prevent that exception from occurring.  I agree that since you can't in practice use 16385-20479 byte sizes then it just makes sense to use 20480.  I'll update the value and the comment to be more explicit.

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
jwallra: If I understand correctly, a future is analagous to a callback in that we're saving it here, and completing either immediately (if there is some parameter/timing violation) or saving and completing it [here](https://github.com/Microsoft/AppCenter-SDK-Android/pull/796/files#diff-918508e799dc541fd29d897160e9b0a5R749)

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
jwallra: Unless there's documentation for `ceil()` that I'm missing, it would still have to be in terms of multiples of 4 pages so something like `long res = ceil(max/ALLOWED_MULTIPLE)*ALLOWED_MULTIPLE` which is essentially the same.  Do you think the `ceil()` version is more understandable?

achocron: I personally do, bu it isn't a huge deal to me either way.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -537,6 +509,42 @@ void setSQLiteOpenHelper(@NonNull SQLiteOpenHelper helper) {
         mSQLiteOpenHelper = helper;
     }
 
+    /**
+     * Set maximum SQLite database size.
+     *
+     * @param maxStorageSizeInBytes Maximum SQLite database size.
+     * @return true if database size was set, otherwise false.
+     */
+    boolean setMaxStorageSize(long maxStorageSizeInBytes) {
```
dhei: nit - to align with `public long getMaxSize()`, maybe we should can this `boolean setMaxSize(long maxStorageSizeInBytes)`. It's already in DatabaseManager class so no need to mention "storage".

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -34,9 +37,16 @@
...
      */
-    private static final String PRIMARY_KEY_SELECTION = "oid = ?";
+    @VisibleForTesting
+    static final int ALLOWED_SIZE_MULTIPLE = 4096;
+
+    /**
+     * Maximum number of entries of in memory database.
+     */
+    @VisibleForTesting
+    static final long IN_MEMORY_MAX_SIZE = 300;
```
jaelim-ms: Should we keep this one for IMDB? It is okay but we probably need to consider removing IMDB. Just an opinion. 😄

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -650,6 +718,15 @@ private void finishConfiguration() {
...
         mLogSerializer.addLogFactory(CustomPropertiesLog.TYPE, new CustomPropertiesLogFactory());
         mChannel = new DefaultChannel(mApplication, mAppSecret, mLogSerializer, mHandler);
+
+        /* Complete set maximum storage size future if starting from app. */
+        if (configureFromApp) {
+            applyStorageMaxSize();
+        } else {
+
+            /* If from library, we apply storage size only later, we have to try using the default value in the mean time. */
+            mChannel.setMaxStorageSize(DEFAULT_MAX_STORAGE_SIZE_IN_BYTES);
```
jaelim-ms: Can you explain me about this line? Why do we need to set a storage limit to default value when it is started from lib?

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
```diff
@@ -279,35 +281,133 @@ public void putLargeLogFailsToRead() throws PersistenceException {
...
+                fail("Inserting large common schema log is not supposed to work");
+            } catch (PersistenceException e) {
+
+                /* Count logs is still 0 */
+                e.printStackTrace();
+                assertEquals(0, persistence.countLogs("test-p1"));
+            }
+        } finally {
+
+            /* Close. */
```
jaelim-ms: Remove this comment. Meaningless.

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
```diff
@@ -279,35 +281,133 @@ public void putLargeLogFailsToRead() throws PersistenceException {
...
-            assertEquals(2, outputLogs.size());
-            assertEquals(log3, outputLogs.get(0));
-            assertEquals(log4, outputLogs.get(1));
-            assertEquals(2, persistence.countLogs("test-p1"));
+            persistence.getLogs("test-p1", 7, outputLogs);
+            assertEquals(expectedLogs.size(), persistence.countLogs("test-p1"));
+            assertEquals(expectedLogs, outputLogs);
+        } finally {
+
+            /* Close. */
```
jaelim-ms: Remove this comment. Meaningless.

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/utils/storage/StorageHelperAndroidTest.java
```diff
@@ -769,31 +760,26 @@ public boolean onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
 
             @Override
             public void onError(String operation, RuntimeException e) {
-                throw e;
+                /* Do not handle any errors. This is simulating errors so this is expected. */
```
jaelim-ms: New line before a comment.

sdk/appcenter/src/test/java/com/microsoft/appcenter/persistence/DatabasePersistenceTest.java
```diff
@@ -260,4 +260,19 @@ public Log answer(InvocationOnMock invocation) {
         /* Verify that the only log we deleted in the entire test was the one from previous test (id=1). */
         verify(databaseStorage).delete(anyLong());
     }
+
+    @Test
+    public void checkSetStorageSizeForwarding() {
```
jaelim-ms: I think we can combine two tests.

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
```diff
@@ -279,35 +281,133 @@ public void putLargeLogFailsToRead() throws PersistenceException {
         }
     }
 
+    @Test
+    public void putLargeLogNotSupportedOnCommonSchema() throws JSONException {
```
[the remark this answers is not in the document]
jwallra: It's just making sure that the log doesn't get stored to the database if it's too big.

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
```diff
@@ -85,13 +94,6 @@ private static int getIteratorSize(Iterator iterator) {
         return count;
     }
 
-    @After
-    public void tearDown() {
-
-        /* Clean up database. */
-        sContext.deleteDatabase("test-persistence");
```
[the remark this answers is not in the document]
jwallra: We're already cleaning it up in `setUp`

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
jwallra: Because we still store logs when started from a library, but we don't allow it to be set manually when starting from a library, so we use the default value.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
jwallra: IMDB is used if there is a SQLite catastrophic error, so we just decided to keep the same logic as we do right now instead of writing a bunch more logic to ensure sizing... Although I guess that means that we might go over the manual limit? But we also don't really have any control over that if SQLite fails...

sdk/appcenter/src/test/java/com/microsoft/appcenter/persistence/DatabasePersistenceTest.java
jwallra: The `setMaxStorageSize` method is passed down through both the StorageHelper and DefaultChannel layers, and that code needs to be covered.  It makes sense to put the testing code in the test files for their respective class

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
jwallra: We might need to revisit this if we're seeing that people run into this.  The problem is that we could break our contract on storage size if the 300 logs to be stored in the in-memory DB (which is only used in case of a SQLite failure/access error/etc.) go over the storage size.  It's a lot of extra logic for a pretty extreme edge case.

dhei: @jaelim-ms We decide to keep the same 300 size limit logic for in memory database for simplicity. It is tricky to calculate the actual storage size for in memory data, because you will have to somehow compute the size by writing the data in the database, which is an expensive operation. It is a good enough for the purpose of a fail back solution. In case we find a better solution, we can always improve it later.

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
```diff
@@ -28,6 +28,7 @@
 import com.microsoft.appcenter.utils.storage.StorageHelper.DatabaseStorage.DatabaseScanner;
 
 import org.json.JSONException;
+import org.junit.After;
```
dhei: nit - Is this import really needed?

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
```diff
@@ -322,8 +323,6 @@ public void putLargeLogNotSupportedOnCommonSchema() throws JSONException {
                 assertEquals(0, persistence.countLogs("test-p1"));
             }
         } finally {
-
-            /* Close. */
```
dhei: nit - what is the reason of removing this comment everywhere? Looks like to me it's fine to keep it.

PR #797, opened 2018-09-17
Fix test application crash when start is skipped

MatkovIvan:


guperrot:
We have a design bug in the SDK where getTransmissionTarget needs to be asynchronous, the real fix is to move startFromLibrary earlier and have getTransmissionTarget go through the handler thread so it will have a context instance for sure. This is kind of critical bug.

PR #798, opened 2018-09-17
 When a part C property is invalid, warn/skip instead of rejecting entire log

MatkovIvan:


sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/one/PartCUtils.java
```diff
@@ -31,18 +34,21 @@ public static void addPartCFromLog(Map<String, String> properties, CommonSchemaL
...
-                    throw new IllegalArgumentException("Property key cannot be null.");
+                    AppCenterLog.warn(LOG_TAG, "Property key cannot be null.");
+                    continue;
                 }
 
                 /* Validate value not null. */
                 String value = entry.getValue();
                 if (value == null) {
-                    throw new IllegalArgumentException("Property value cannot be null.");
+                    AppCenterLog.warn(LOG_TAG, "Property value cannot be null.");
```
jaelim-ms: Can we log "key" of the property?

PR #803, opened 2018-09-19
Feature/pause resume

jaelim-ms:
Adding `pause` and `resume` APIs at Analytics level.  Calling `pause` will stop the SDK from sending logs to App Center backend and saves them until `resume` is called.  Then we resume sending the logs

### DIFF
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
@@ -621,14 +621,14 @@ public class Analytics extends AbstractAppCenterService {
      */
     private synchronized AppCenterFuture<Void> pauseInstanceAsync() {
         final DefaultAppCenterFuture<Void> future = new DefaultAppCenterFuture<>();
-        post(new Runnable() {
+        postAsyncGetter(new Runnable() {
 
             @Override
             public void run() {
                 mChannel.pauseGroup(ANALYTICS_GROUP);
                 future.complete(null);
             }
-        });
+        }, future, null);
         return future;
     }
 
@@ -637,14 +637,14 @@ public class Analytics extends AbstractAppCenterService {
      */
     private synchronized AppCenterFuture<Void> resumeInstanceAsync() {
         final DefaultAppCenterFuture<Void> future = new DefaultAppCenterFuture<>();
-        post(new Runnable() {
+        postAsyncGetter(new Runnable() {
 
             @Override
             public void run() {
                 mChannel.resumeGroup(ANALYTICS_GROUP);
                 future.complete(null);
             }
-        });
+        }, future, null);
         return future;
     }
 
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
@@ -167,6 +167,24 @@ public class Analytics extends AbstractAppCenterService {
     }
 
     /**
+     * Pauses log transmission.
+     *
+     * @return future with null result to monitor when the operation completes.
+     */
+    public static AppCenterFuture<Void> pause() {
+        return getInstance().pauseInstanceAsync();
+    }
+
+    /**
+     * Resumes log transmission if paused.
+     *
+     * @return future with null result to monitor when the operation completes.
+     */
+    public static AppCenterFuture<Void> resume() {
+        return getInstance().resumeInstanceAsync();
+    }
+
+    /**
      * Sets an analytics listener.
      * <p>
      * Note: it needs to be protected for Xamarin to change it back to public in bindings.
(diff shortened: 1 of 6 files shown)


sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTest.java
```diff
@@ -332,6 +332,43 @@ public void disablePersisted() {
...
+        verify(channel).addListener(isA(SessionTracker.class));
+        verify(channel).addListener(isA(AnalyticsValidator.class));
+
+        /* Pause Analytics. */
+        Analytics.pause().get();
+
+        /* Check if Analytics group is paused. */
+        verify(channel).pauseGroup(eq(analytics.getGroupName()));
+
+        /* Now try to use all methods. Should work. */
```
jwallra: This comment and the one on line 366 should be updated.  We don't need to test all methods here; it would be redundant.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -598,6 +616,40 @@ private synchronized void setInstanceListener(AnalyticsListener listener) {
         mAnalyticsListener = listener;
     }
 
+    /**
+     * Implements {@link #pause()}}.
+     */
+    private synchronized AppCenterFuture<Void> pauseInstanceAsync() {
+        final DefaultAppCenterFuture<Void> future = new DefaultAppCenterFuture<>();
+        Runnable runnable = new Runnable() {
```
guperrot: nit: can inline variable

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -598,6 +616,40 @@ private synchronized void setInstanceListener(AnalyticsListener listener) {
...
+        post(runnable);
+        return future;
+    }
+
+    /**
+     * Implements {@link #resume()}}.
+     */
+    private synchronized AppCenterFuture<Void> resumeInstanceAsync() {
+        final DefaultAppCenterFuture<Void> future = new DefaultAppCenterFuture<>();
+        Runnable runnable = new Runnable() {
```
guperrot: nit: can inline variable

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTest.java
```diff
@@ -332,6 +332,43 @@ public void disablePersisted() {
         verify(channel, never()).addListener(any(Channel.Listener.class));
     }
 
+    @Test
+    public void notSendingLogsOnPause() {
+
+        /* Before start it does not work to change state, it's disabled. */
```
guperrot: then we should have a test for what happens if calling before start and happens when calling pause/resume while disabled.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -232,6 +233,26 @@ public synchronized void removeGroup(String groupName) {
...
+        GroupState groupState = mGroupStates.get(groupName);
+        if (groupState != null && !groupState.mPaused) {
+            groupState.mPaused = true;
+            cancelTimer(groupState);
+        }
+    }
+
+    @Override
+    public synchronized void resumeGroup(String groupName) {
+        AppCenterLog.debug(LOG_TAG, "resumeGroup(" + groupName + ")");
```
guperrot: I would move the debug log to inside the if.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -232,6 +233,26 @@ public synchronized void removeGroup(String groupName) {
         }
     }
 
+    @Override
+    public synchronized void pauseGroup(String groupName) {
+        AppCenterLog.debug(LOG_TAG, "pauseGroup(" + groupName + ")");
```
guperrot: I would move the debug log to inside the if.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -755,6 +793,7 @@ public void run() {
             mMaxParallelBatches = maxParallelBatches;
             mIngestion = ingestion;
             mListener = listener;
+            mPaused = false;
```
guperrot: nit: by convention we never initialize explicitly to default value, since false is already default

jwallra: I'm ok with changing it back if it's convention, but personally I can't remember default values for variables for a ton of different languages, so in my mind it becomes hidden logic.  I very much dislike this convention.

guperrot: Its null, 0 and false, there is nothing else actually.

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTest.java
jwallra: I'll add the one for start but we already have tests for being disabled [here](https://github.com/Microsoft/AppCenter-SDK-Android/pull/803/files#diff-fb7e581dbd71041b785eca0c6a15b437R807)

guperrot: No thats not testing that when Analytics disabled it's not even trying to call channel APIs.

PR #804, opened 2018-09-19
Add a warning log when it overrides exist values in Part C properties

jaelim-ms:


sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/one/PartCUtils.java
```diff
@@ -60,6 +60,8 @@ public static void addPartCFromLog(Map<String, String> properties, CommonSchemaL
                     if (subObject == null) {
                         subObject = new JSONObject();
                         destProperties.put(keys[i], subObject);
+                    } else {
+                        AppCenterLog.warn(LOG_TAG, "Property key '" + keys[i] + "' already has a value, the old value will be overridden.");
```
guperrot: we need the unit test for adding “a.b” in a linked hash map, then override with top level “a” property

and verify only top level property is sent
and optionally another test that does it in reverse order

MatkovIvan: It prints the warning on the next case:
```json
{
  "a.b": "1",
  "a.c": "2"
}
```

that's wrong I guess

PR #807, opened 2018-09-20
Fix creating a target right after AppCenter.start

guperrot:
Use the fact that any operation are queued into background looper so it is guaranteed that the `initInBackground` will be called in a sequence after the `Analytics.onStarted` that set context/channel to non null, in other terms if you execute anything in post, it will always have non null channel/context.

### DIFF
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventActivity.java
@@ -48,6 +48,9 @@ public class EventActivity extends LogActivity {
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
+        /* Test start from library. */
+        AppCenter.startFromLibrary(this, Analytics.class);
+
         /* Transmission target views init. */
         mTransmissionTargetSpinner = findViewById(R.id.transmission_target);
         ArrayAdapter<String> targetAdapter = new ArrayAdapter<>(this, android.R.layout.simple_spinner_dropdown_item, getResources().getStringArray(R.array.target_id_names));
@@ -100,9 +103,6 @@ public class EventActivity extends LogActivity {
                 startActivity(intent);
             }
         });
-
-        /* Test start from library. */
-        AppCenter.startFromLibrary(this, Analytics.class);
     }
 
     @Override
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
@@ -657,6 +657,11 @@ public class Analytics extends AbstractAppCenterService {
         postAsyncGetter(runnable, future, valueIfDisabledOrNotStarted);
     }
 
+    /**
+     * Post a command that will run on background even if SDK disabled (needs to be configured though).
+     *
+     * @param runnable command.
+     */
     void postCommandEvenIfDisabled(Runnable runnable) {
         post(runnable, runnable, runnable);
     }
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
@@ -320,8 +320,6 @@ public class Analytics extends AbstractAppCenterService {
             AppCenterLog.error(LOG_TAG, "Transmission target token may not be null or empty.");
             return null;
         } else if (!AppCenter.isConfigured()) {
-
-            /* Context and channel have the same lifecycle so we only have to check one. */
             AppCenterLog.error(LOG_TAG, "Cannot create transmission target, AppCenter is not configured or started.");
             return null;
         } else {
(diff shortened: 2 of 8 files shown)


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
```diff
@@ -208,9 +212,12 @@ public synchronized void removeEventProperty(String key) {
...
     }
 
-    @SuppressLint("HardwareIds")
+    /**
+     * Enable collection of the Android device identifier for this target.
+     * This does not have any effect on child transmission targets.
+     */
     public void collectDeviceId() {
-        mDeviceId = Secure.getString(mTransmissionTarget.mContext.getContentResolver(), Secure.ANDROID_ID);
+        mDeviceIdEnabled = true;
```
guperrot: Context can be null when calling this early so we just remember the call and apply later.

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTest.java
```diff
@@ -507,12 +507,17 @@ public boolean matches(Object item) {
...
+        when(AppCenter.isConfigured()).thenReturn(false);
+
+        /* When creating a target, it returns null. */
         assertNull(Analytics.getTransmissionTarget("t1"));
 
-        /* Verify log. */
+        /* And prints an error. */
         verifyStatic();
-        AppCenterLog.error(anyString(), contains("App context is null, App Center has not been started."));
+        AppCenterLog.error(anyString(), contains("AppCenter is not configured"));
```
jwallra: Add period at end of log message

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventActivity.java
```diff
@@ -48,6 +48,9 @@
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
+        /* Test start from library. */
+        AppCenter.startFromLibrary(this, Analytics.class);
+
```
guperrot: SDK checks SDK started before you can create targets so had to move earlier.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -624,7 +633,9 @@ public void onConfigurationUpdated(String appSecret, String transmissionTargetTo
      */
     @WorkerThread
     private void setDefaultTransmissionTarget(String transmissionTargetToken) {
-        mDefaultTransmissionTarget = getInstanceTransmissionTarget(transmissionTargetToken);
+        if (transmissionTargetToken != null) {
+            mDefaultTransmissionTarget = getInstanceTransmissionTarget(transmissionTargetToken);
```
guperrot: This is already covered in unit test. Without the if, mDefaultTransmissionTarget would still be null but with an error log saying that you need to pass non null. Since we do it internally its confusing for users (this happens when starting with no secret from an app).

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTest.java
guperrot: its not the log but the test verification with contains a substring

jwallra: Whoops, should pay closer attention to what I'm looking at 😅

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
jaelim-ms: Understand this change but we aligned the iOS implementation matching with Android.
Not a big deal but we need to change it in iOS, too.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -59,7 +61,8 @@
     /**
      * Shared instance.
      */
-    private static Analytics sInstance = null;
+    @SuppressLint("StaticFieldLeak")
+    private static Analytics sInstance;
```
clpolet: Just curious, what's the reason for this change?

jaelim-ms:
Isn't is a fix for the comment that you mentioned in #797?

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
guperrot: It was a warning we always ignore, also no need to init with null as its default value.

jaelim-ms: We don't initialize anything to null since it is by default. And it added `SuppressLint` to ignore lint complains.

guperrot:
@jaelim-ms yes I should close #797 after merging this.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
guperrot: I don't think we should change iOS as this constraint is a bug fix for Android.

PR #808, opened 2018-09-20
Add UI for pause and resume

guperrot:


achocron:
Small nit, instead of saying "Analytics is resumed" can it say "Analytics is not paused"? It's a bit strange for something to be "resumed" (and I'd suggest "Analytics is running" but that can get confusing since we have enabled/disabled state as well)

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -86,6 +89,9 @@ public void onCreate(Bundle savedInstanceState) {
                 @Override
                 public void setEnabled(boolean enabled) {
                     AppCenter.setEnabled(enabled);
+                    if (enabled) {
```
jaelim-ms: `If` isn't necessary.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -142,13 +148,51 @@ public void onEvent(int event, @Nullable String path) {
                 @Override
                 public void setEnabled(boolean enabled) {
                     Analytics.setEnabled(enabled);
+                    if (enabled) {
```
jaelim-ms: `If` isn't necessary.

guperrot:
@achocron but the API we used is resume so it makes sense that the state is resumed, pretty much like activity life cycle in Android.

PR #809, opened 2018-09-20
Improve storage size UI

guperrot:


apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
```diff
@@ -231,19 +231,23 @@ private void setMaxStorageSize() {
...
                             MainActivity.this.getString(R.string.max_storage_size_change_success),
                             Formatter.formatFileSize(MainActivity.this, expectedMultipleMaxSize)), Toast.LENGTH_SHORT).show();
                     sSharedPreferences.edit().putLong(MAX_STORAGE_SIZE_KEY, expectedMultipleMaxSize).apply();
                 } else {
+
+                    /* SQLite shrinks to fileSize rounded to next page size in that case. */
                     Toast.makeText(MainActivity.this, R.string.max_storage_size_change_failed, Toast.LENGTH_SHORT).show();
-                    sSharedPreferences.edit().putLong(MAX_STORAGE_SIZE_KEY, 0).apply();
+                    String DATABASE_NAME = "com.microsoft.appcenter.persistence";
+                    long fileSize = getDatabasePath(DATABASE_NAME).length();
```
jaelim-ms: This is an actual file size. It is showing the value that isn't set by a user so it is a wrong value. Can we just reset the value?

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
```diff
@@ -231,19 +231,23 @@ private void setMaxStorageSize() {
...
+                    long maxSize = (long) Math.ceil((double) fileSize / (double) ALLOWED_SIZE_MULTIPLE) * ALLOWED_SIZE_MULTIPLE;
```
jaelim-ms: This is redundant since you are reading an actual size. It will always be incremental of 4K.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
guperrot: No I verified its the real max size set by SDK it actually shrinks that way.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
guperrot: Sure, will simplify code.

PR #813, opened 2018-09-21
Adding UI for transmission target pause/resume

jwallra:
image

### DIFF
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventActivity.java
@@ -119,14 +119,17 @@ public class EventActivity extends LogActivity {
             @Override
             public void onClick(View v) {
                 AnalyticsTransmissionTarget target = getSelectedTarget();
+                Method pauseMethod;
                 try {
-                    final Method pauseMethod = target.getClass().getMethod("pause");
-
-                    @SuppressWarnings("unchecked")
+                    pauseMethod = target.getClass().getMethod("pause");
+                } catch (Exception ignored) {
+                    return;
+                }
+                try {
                     AppCenterFuture<Void> future = (AppCenterFuture<Void>) pauseMethod.invoke(null);
                     future.get();
-                } catch (Exception e) {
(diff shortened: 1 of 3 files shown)


jwallra:
Still needs to wait on feature implementation

guperrot:
@jwallra since it will use reflection anyway you can already code the calls.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventActivity.java
```diff
@@ -103,6 +110,41 @@ public void onClick(View v) {
...
+        /* Init pause/resume buttons. */
+        /* TODO remove reflection once SDKs are released. */
+        mPauseTransmissionButton = findViewById(R.id.pause_transmission_button);
+        mPauseTransmissionButton.setOnClickListener(new View.OnClickListener() {
+
+            @Override
+            public void onClick(View v) {
+                AnalyticsTransmissionTarget target = getSelectedTarget();
+                try {
+                    final Method pauseMethod = target.getClass().getMethod("pause");
```
guperrot: In java using final variables in methods is very unusual unless it is used into an inner anonymous class which is not the case here.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventActivity.java
```diff
@@ -103,6 +110,41 @@ public void onClick(View v) {
...
+
+                    @SuppressWarnings("unchecked")
+                    AppCenterFuture<Void> future = (AppCenterFuture<Void>) pauseMethod.invoke(null);
+                    future.get();
+                } catch (Exception e) {
+                    e.printStackTrace();
```
guperrot: Please align reflection code with collectdeviceId, also we do not log using e.printStackTrace as it does use the System log tag instead of app log tag (and some devices disable it).

jwallra:
@achocron It's to allow testing calling `pause` or `resume` more than once in a row.  This also prevents us from having to save and modify a list of states for each transmission target.  This PR still needs improvements to indicate if a transmission target is paused/unpaused, though.

PR #814, opened 2018-09-24
Change log level to info

jaelim-ms:


sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -528,7 +528,7 @@ boolean setMaxSize(long maxStorageSizeInBytes) {
             return false;
         }
         if (maxStorageSizeInBytes != newMaxSize) {
-            AppCenterLog.warn(LOG_TAG, "Could change database size but is slightly larger than expected (next multiple of 4KB), requestedMaxSize=" +
+            AppCenterLog.info(LOG_TAG, "Could change database size but is slightly larger than expected (next multiple of 4KB), requestedMaxSize=" +
```
achocron: The text is confusing. If i see that I might think that it didn't actually change anything and won't try to. I suggest we say `Will attempt to change database size to {size in bytes} (next multiple of 4KiB)`

ElektrojungeMS: I like this suggestion!

jwallra: I like the text used in [iOS](https://github.com/Microsoft/AppCenter-SDK-Apple/pull/1125/files), we should align these anyways

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -520,16 +520,15 @@ boolean setMaxSize(long maxStorageSizeInBytes) {
...
 
         /* So to check the resize works, we need to check new max size against the next multiple of 4KB. */
         if (newMaxSize != expectedMultipleMaxSize) {
             AppCenterLog.error(LOG_TAG, "Could not change database size to " + maxStorageSizeInBytes + " bytes, current max size is " + newMaxSize + " bytes.");
             return false;
         }
         if (maxStorageSizeInBytes != newMaxSize) {
-            AppCenterLog.warn(LOG_TAG, "Could change database size but is slightly larger than expected (next multiple of 4KB), requestedMaxSize=" +
-                    maxStorageSizeInBytes + " actualMaxSize=" + newMaxSize);
+            AppCenterLog.info(LOG_TAG, "Will attempt to change database size to " + newMaxSize + " bytes (next multiple of 4KiB).");
```
guperrot: On Android the operation is already done at this point, it's not "will". Is that different on ios?

PR #819, opened 2018-09-27
Integration for coveralls

guperrot:


guperrot:
Need 2 other minor changes actually, badge and accept coveralls token as variable as well.

guperrot:
Does not work from VSTS PRs because of missing branch name.

PR #828, opened 2018-10-05
Fix CustomProperty serialization error if using NaN or Infinity for doubles

MatkovIvan:


sdk/appcenter/src/main/java/com/microsoft/appcenter/CustomProperties.java
```diff
@@ -165,4 +161,17 @@ private boolean isValidStringValue(String key, String value) {
...
     }
+
+    private boolean isValidNumberValue(String key, Number value) {
+        if (value == null) {
+            AppCenterLog.error(AppCenter.LOG_TAG, VALUE_NULL_ERROR_MESSAGE);
+            return false;
+        }
+        double doubleValue = value.doubleValue();
+        if (Double.isInfinite(doubleValue) || Double.isNaN(doubleValue)) {
+            AppCenterLog.error(AppCenter.LOG_TAG, "Custom property \""+ key + "\" value is not a valid number.");
```
jaelim-ms: `"Custom property \"" + key + "\" cannot be NaN or infinite"`?

MatkovIvan: `"Custom property \""+ key + "\" value cannot be NaN or infinite."` is better

PR #836, opened 2018-10-09
Use Android Studio 3.2

MatkovIvan:


sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/StorageHelper.java
```diff
@@ -586,6 +586,7 @@ public static DatabaseStorage getDatabaseStorage(@NonNull String database,
          * Set maximum SQLite database size.
          *
          * @param maxStorageSizeInBytes Maximum SQLite database size.
+         * @return true if database size was set, otherwise false.
```
ElektrojungeMS: Technically not in scope of the PR but it's always good to leave the campground cleaner ;)

guperrot:
Testing more build definitions now.

PR #840, opened 2018-10-11
Fix 2 issues related to thread safety in CustomProperties

guperrot:
* Avoid concurrent modification exceptions while and after calling setCustomProperties by taking a snapshot of properties when we call `setCustomProperties`.
* Avoid bypassing the size limit by inserting keys from multiple threads.


guperrot:
It's not the same fix as in EventProperties as we don't do traversals in multiple places, plus we have the size check that is done at set time and needs to be atomic as well.

PR #850, opened 2018-10-20
Fix part A properties to not be applied retro-actively

guperrot:
* [x] Has `CHANGELOG.md` been updated? (Yes but in https://msasg.visualstudio.com/Shared%20Data/_git/1dsdocs/pullrequest/809581?_a=overview)
* [x] Did you add unit tests?

## Description

* Fix part A properties to not be applied retro-actively.
* Fix distinction between default target and explicit target using the same token.

### DIFF
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
@@ -438,7 +438,7 @@ public class Analytics extends AbstractAppCenterService {
     }
 
     /**
-     * Unconditionally create a new transmission target at root level.
+     * Unconditionally create a new transmission target at root level, even if one exists with the given token.
      *
      * @param transmissionTargetToken the token.
      * @return the created target.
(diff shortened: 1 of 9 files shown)


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -430,21 +431,31 @@ private synchronized AnalyticsTransmissionTarget getInstanceTransmissionTarget(S
...
         }
     }
 
+    /**
+     * Unconditionally create a new transmission target at root level.
+     *
+     * @param transmissionTargetToken the token.
+     * @return the created target.
+     */
+    private AnalyticsTransmissionTarget createAnalyticsTransmissionTarget(String transmissionTargetToken) {
```
jwallra: Since you split it off, can you be more explicit in the method name? `createAnalyticsTransmissionTargetAtRoot` is my suggestion.

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/models/AbstractLog.java
```diff
@@ -136,6 +141,16 @@ public void setDevice(Device device) {
         this.device = device;
     }
 
+    @Override
+    public Object getTag() {
```
jaelim-ms: This method is only used by unit tests. Any other use cases for the method?

PR #854, opened 2018-10-30
Fix warnings in channel

guperrot:
A pr to fix the new code analysis issues found by Android Studio 3.2.
Ideally merged before refactoring enqueueItem to add priority.

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
@@ -518,7 +518,7 @@ public class DefaultChannel implements Channel {
 
                         @Override
                         public void run() {
-                            handleSendingSuccess(groupState, currentState, batchId);
+                            handleSendingSuccess(groupState, batchId);
                         }
                     });
                 }
@@ -529,7 +529,7 @@ public class DefaultChannel implements Channel {
 
                         @Override
                         public void run() {
-                            handleSendingFailure(groupState, currentState, batchId, e);
+                            handleSendingFailure(groupState, batchId, e);
                         }
                     });
                 }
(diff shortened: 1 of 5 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -717,27 +706,26 @@ public synchronized void enqueue(@NonNull Log log, @NonNull final String groupNa
...
+            triggerIngestion(groupState);
         } else if (pendingLogCount > 0 && !groupState.mScheduled) {
             groupState.mScheduled = true;
             mAppCenterHandler.postDelayed(groupState.mRunnable, groupState.mBatchTimeInterval);
         }
     }
 
     @VisibleForTesting
-    GroupState getGroupState(String groupName) {
+    GroupState getGroupState(@SuppressWarnings("SameParameterValue") String groupName) {
```
jaelim-ms: Not a big deal but we used to have method level suppress warning in this case. Can we change?

guperrot: This one is parameter specific though. We have other examples of this one in the repo.

jaelim-ms: I thought this is a kind of rule what we do but you are saying it isn't. OK. I'd prefer this way since this will hide some warnings that we don't expect in other parameters. 
There is an example that we are not doing this one. We probably need to check all of these cases in the future (low priority).
[Example](https://github.com/Microsoft/AppCenter-SDK-Android/blob/develop/sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java#L159)

guperrot: Yes this should have been reviewed as suppressing warnings for all parameters does not help knowing what parameter is always used with 1 value.

One other exception is trycanusefinally that can be at class level as this is more global as we can't use it because of min API level so it does not make sense to put it in each method or statement.

PR #858, opened 2018-10-31
Purge critical logs last

guperrot:

* [x] Did you add unit tests?

## Description

Implement persistence flag for purge of the logs.

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
@@ -199,11 +199,11 @@ public class DatabaseManager implements Closeable {
      * the log is not inserted.
      *
      * @param values         The entry to be stored.
-     * @param priorityColumn When storage full and deleting data,use this order by clause to determine which entries to delete first.
+     * @param priorityColumn When storage full and deleting data, optionally use this order by clause to determine which entries to delete first.
      * @return If a log was inserted, the database identifier. Otherwise -1.
      */
     @SuppressWarnings("TryFinallyCanBeTryWithResources")
-    public long put(@NonNull ContentValues values, @NonNull String priorityColumn) {
+    public long put(@NonNull ContentValues values, @Nullable String priorityColumn) {
         try {
             while (true) {
                 try {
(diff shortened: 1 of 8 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -310,18 +323,18 @@ public final long getRowCount() {
...
      * @param queryBuilder  The query builder that contains SQL query.
+     * @param columns       Columns to select, null for all.
      * @param selectionArgs The array of values for selection.
-     * @param idOnly        Return only row identifier if true, return all fields otherwise.
+     * @param sortOrder     Sorting order (ORDER BY clause without ORDER BY itself).
      * @return A cursor for all rows that matches the given criteria.
      * @throws RuntimeException If an error occurs.
      */
-    public Cursor getCursor(@Nullable SQLiteQueryBuilder queryBuilder, @Nullable String[] selectionArgs, boolean idOnly) throws RuntimeException {
+    public Cursor getCursor(@Nullable SQLiteQueryBuilder queryBuilder, String[] columns, @Nullable String[] selectionArgs, @Nullable String sortOrder) throws RuntimeException {
```
guperrot: During implementation at this commit I needed the new signature because I was selecting only columns oid and priority to avoid loading all data: https://github.com/Microsoft/AppCenter-SDK-Android/pull/858/commits/19f6c5a8fc58d269061b85190c50d6fbd2d01151#diff-9b79a28544bf85990330f7945382e4ebR217

Later I simplified code and now need only id like before. But I kept the refactoring change as it also allows to optimize the count query, which is now included in this change, this could have been part of the storagehelper refactoring but didn't catch that during review.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -198,36 +203,44 @@ public ContentValues nextValues(Cursor cursor) {
...
+                    if (cursor == null) {
+                        String priority = values.getAsString(priorityColumn);
+                        SQLiteQueryBuilder queryBuilder = SQLiteUtils.newSQLiteQueryBuilder();
+                        queryBuilder.appendWhere(priorityColumn + " <= ?");
+                        cursor = getCursor(queryBuilder, SELECT_PRIMARY_KEY, new String[]{priority}, priorityColumn);
+                    }
+                    if (cursor.moveToNext()) {
+                        delete(cursor.getLong(0));
+                    } else {
+                        throw e;
```
guperrot: Now we get an error message in the logs if full. Before it was silenced and just returning -1.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -198,36 +203,44 @@ public ContentValues nextValues(Cursor cursor) {
...
                     }
                 }
             }
         } catch (RuntimeException e) {
+            id = -1L;
             AppCenterLog.error(AppCenter.LOG_TAG, String.format("Failed to insert values (%s) to database.", values.toString()), e);
         }
-        return -1;
+        if (cursor != null) {
+            cursor.close();
```
guperrot: This can throw exception, need to move, oops

guperrot:
Ready for review, just testing from app manually to make sure, that's only reason for don't merge label.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -198,36 +203,47 @@ public ContentValues nextValues(Cursor cursor) {
...
-                        } else {
-                            return -1;
-                        }
-                    } finally {
-                        cursor.close();
+                    if (cursor == null) {
+                        String priority = values.getAsString(priorityColumn);
+                        SQLiteQueryBuilder queryBuilder = SQLiteUtils.newSQLiteQueryBuilder();
+                        queryBuilder.appendWhere(priorityColumn + " <= ?");
+                        cursor = getCursor(queryBuilder, SELECT_PRIMARY_KEY, new String[]{priority}, priorityColumn);
```
jaelim-ms: It would be better to explicitly specify sort order for primary key. `priority` then `oid`

Issue #855, opened 2018-10-30
StackOverflowError on App start


codeknox:
### **Description**

StackOverflowError during app start

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. connect app to app center with provided code
2. after using you app for a while, when it crashes, bad things happen

### **Details**

1. Which SDK version are you using?
    - 1.8.0
2. Which OS version did you experience the issue on?
    - Android 7.0
3. What device version did you see this error on?  Were you using an emulator or a physical device?
    - physical tablet device
4. What third party libraries are you using?
    - lots of them, but I lack to see the importance here

5. Please enable verbose logging for your app using `AppCenter.setLogLevel(Log.VERBOSE)` before your call to `AppCenter.start(...)` and include the logs here:

I will work on collecting the logs, as I need a new deploy, but for now, here is the stack trace:
```
STACK_TRACE: java.lang.StackOverflowError
	at dalvik.system.BlockGuard.getThreadPolicy(BlockGuard.java:140)
	at libcore.io.BlockGuardOs.read(BlockGuardOs.java:148)
	at libcore.io.IoBridge.read(IoBridge.java:432)
	at java.io.FileInputStream.read(FileInputStream.java:179)
	at libcore.io.Streams.readSingleByte(Streams.java:41)
	at java.io.FileInputStream.read(FileInputStream.java:175)
	at java.io.FilterInputStream.read(FilterInputStream.java:114)
	at java.io.ObjectInputStream.checkReadPrimitiveTypes(ObjectInputStream.java:405)
	at java.io.ObjectInputStream.readNonPrimitiveContent(ObjectInputStream.java:770)
	at java.io.ObjectInputStream.readObject(ObjectInputStream.java:2003)
	at java.io.ObjectInputStream.readObject(ObjectInputStream.java:1960)
	at java.io.ObjectInputStream.readFieldValues(ObjectInputStream.java:1137)
	at java.io.ObjectInputStream.defaultReadObject(ObjectInputStream.java:455)
	at java.lang.Throwable.readObject(Throwable.java:444)
	at java.lang.reflect.Method.invokeNative(Native Method)
	at java.lang.reflect.Method.invoke(Method.java:511)
	at java.io.ObjectInputStream.readObjectForClass(ObjectInputStream.java:1354)
	at java.io.ObjectInputStream.readHierarchy(ObjectInputStream.java:1266)
[...]

PR #859, opened 2018-10-31
Preventing stack overflow crash while reading a huge throwable file

MatkovIvan:
* [x] Did you add unit tests?

## Description

Preventing stack overflow crash while reading a huge throwable file.

## Related PRs or issues

#855

### DIFF
--- a/apps/sasquatch/src/main/res/values/crashes.xml
+++ b/apps/sasquatch/src/main/res/values/crashes.xml
@@ -6,8 +6,8 @@
     <string name="description_crash_divide_by_0" tools:ignore="MissingTranslation">Divide by 0</string>
     <string name="title_test_ui_crash" tools:ignore="MissingTranslation">Test some UI crash</string>
     <string name="description_test_ui_crash" tools:ignore="MissingTranslation">A UI coding mistake, has inner exception</string>
-    <string name="title_deep_cause_exception_crash" tools:ignore="MissingTranslation">A exception with plenty of causes</string>
-    <string name="description_deep_cause_exception_crash" tools:ignore="MissingTranslation">The exception that caused the exception that caused the exception that caused the exception…</string>
+    <string name="title_deeply_nested_exception_crash" tools:ignore="MissingTranslation">A exception with plenty of causes</string>
+    <string name="description_deeply_nested_exception_crash" tools:ignore="MissingTranslation">The exception that caused the exception that caused the exception that caused the exception…</string>
     <string name="title_stack_overflow_crash" tools:ignore="MissingTranslation">Stack overflow crash</string>
     <string name="description_stack_overflow_crash" tools:ignore="MissingTranslation">Caused by infinite recursion</string>
     <string name="title_memory_crash" tools:ignore="MissingTranslation">Out of memory crash</string>
--- a/apps/sasquatch/src/main/res/values/crashes.xml
+++ b/apps/sasquatch/src/main/res/values/crashes.xml
@@ -6,6 +6,8 @@
     <string name="description_crash_divide_by_0" tools:ignore="MissingTranslation">Divide by 0</string>
     <string name="title_test_ui_crash" tools:ignore="MissingTranslation">Test some UI crash</string>
     <string name="description_test_ui_crash" tools:ignore="MissingTranslation">A UI coding mistake, has inner exception</string>
+    <string name="title_deep_cause_exception_crash" tools:ignore="MissingTranslation">A exception with plenty of causes</string>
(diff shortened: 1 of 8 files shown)


apps/sasquatch/src/main/res/values/crashes.xml
```diff
@@ -6,6 +6,8 @@
     <string name="description_crash_divide_by_0" tools:ignore="MissingTranslation">Divide by 0</string>
     <string name="title_test_ui_crash" tools:ignore="MissingTranslation">Test some UI crash</string>
     <string name="description_test_ui_crash" tools:ignore="MissingTranslation">A UI coding mistake, has inner exception</string>
+    <string name="title_deep_cause_exception_crash" tools:ignore="MissingTranslation">A exception with plenty of causes</string>
+    <string name="description_deep_cause_exception_crash" tools:ignore="MissingTranslation">The exception that caused the exception that caused the exception that caused the exception…</string>
```
jwallra: 😂

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -752,10 +752,10 @@ ErrorReport buildErrorReport(ManagedErrorLog log) {
                     ErrorReport report = ErrorLogHelper.getErrorReportFromErrorLog(log, throwable);
                     mErrorReportCache.put(id, new ErrorLogReport(log, report));
                     return report;
-                } catch (ClassNotFoundException ignored) {
-                    AppCenterLog.error(LOG_TAG, "Cannot read throwable file " + file.getName(), ignored);
-                } catch (IOException ignored) {
-                    AppCenterLog.error(LOG_TAG, "Cannot access serialized throwable file " + file.getName(), ignored);
+                } catch (ClassNotFoundException | StackOverflowError e) {
```
jwallra: Should these exceptions really have the same message?  I think we can be more explicit here so we can investigate more precisely when for support issues.

apps/sasquatch/src/main/res/values/crashes.xml
jaelim-ms: 😂

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
dhei: `e` is logged in `AppCenterLog.error(LOG_TAG, "Cannot read throwable file " + file.getName(), e)` so it should be fine to reuse the same message.

apps/sasquatch/src/main/res/values/crashes.xml
dhei: naming suggestion: something like "deeply_nested_exception_crash"

guperrot:
Also another comment but might separate PR, the issue caused empty json files to be created and they are never cleaned up, we should also make a patch to delete the JSON file when we fail to read it for any reason.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
guperrot: I would say we can even merge the 2 catch remaining catch blocks as the difference between messages is not useful.

MatkovIvan:
Well, now when SDK try to send it without throwable, it crashes with
```
java.lang.OutOfMemoryError: Failed to allocate a 150994952 byte allocation with 25165824 free bytes and 103MB until OOM
        at org.json.JSONStringer.newline(JSONStringer.java:360)
        at org.json.JSONStringer.beforeKey(JSONStringer.java:390)
        at org.json.JSONStringer.key(JSONStringer.java:374)
        at org.json.JSONObject.writeTo(JSONObject.java:723)
        at org.json.JSONStringer.value(JSONStringer.java:237)
        at org.json.JSONArray.writeTo(JSONArray.java:613)
        at org.json.JSONStringer.value(JSONStringer.java:233)
        at org.json.JSONObject.writeTo(JSONObject.java:723)
        at org.json.JSONStringer.value(JSONStringer.java:237)
        at org.json.JSONArray.writeTo(JSONArray.java:613)
        at org.json.JSONStringer.value(JSONStringer.java:233)
...
        at org.json.JSONObject.toString(JSONObject.java:716)
        at com.microsoft.appcenter.http.DefaultHttpClient.doHttpCall(DefaultHttpClient.java:222)
        at com.microsoft.appcenter.http.DefaultHttpClient.doCall(DefaultHttpClient.java:152)
```

PR #861, opened 2018-11-01
Add flags to trackEvent API

guperrot:

* [x] Did you add unit tests?

## Description

Add flags to trackEvent that take EventProperties as a parameter.
Use Flags.DEFAULT_FLAGS for all previous signatures.

### DIFF
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
@@ -278,7 +278,7 @@ public class Analytics extends AbstractAppCenterService {
      * @param name An event name.
      */
     public static void trackEvent(String name) {
-        trackEvent(name, null, null);
+        trackEvent(name, null, null, Flags.DEFAULT_FLAGS);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -342,14 +342,49 @@ public static void trackEvent(String name, Map<String, String> properties) {
...
+     * <li>The <tt>baseData</tt> and <tt>baseDataType</tt> properties are reserved and thus discarded.</li>
+     * <li>The full event size when encoded as a JSON string cannot be larger than 1.9MB.</li>
+     * </ul>
+     *
+     * @param name       An event name.
+     * @param properties Optional properties.
+     * @param flags      Optional flags. Use {@link Flags#PERSISTENCE_CRITICAL} to send this event
+     *                   before events using that use default flags or {@link Flags#PERSISTENCE_NORMAL}.
+     */
+    public static void trackEvent(String name, EventProperties properties, int flags) {
```
MatkovIvan: Something like `@IntRange` doesn't apply here?

guperrot: We don't use them for public APIs as it triggers warnings for people not having support annotations. So we validate and log at runtime everywhere in the SDK.

PR #871, opened 2018-11-07
Create index on priority table

MatkovIvan:
* [ ] ~Has `CHANGELOG.md` been updated?~
* [x] Did you add unit tests?

## Description

Create index on priority table

### DIFF
--- a/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
+++ b/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
@@ -377,7 +377,7 @@ public class DatabasePersistenceAndroidTest {
                 persistence.putLog(log, group, PERSISTENCE_NORMAL);
                 allLogs.add(log);
 
-                /* Fail if no eviction happen after a long time to avoid infinite loop on bug. */
+                /* Fail if no eviction happens after a long time to avoid infinite loop on bug. */
                 assertTrue("No eviction is happening", i < 1000);
             }
 
@@ -426,7 +426,7 @@ public class DatabasePersistenceAndroidTest {
                 persistence.putLog(log, group, PERSISTENCE_NORMAL);
                 allNormalLogs.add(log);
 
-                /* Fail if no eviction happen after a long time to avoid infinite loop on bug. */
+                /* Fail if no eviction happens after a long time to avoid infinite loop on bug. */
                 assertTrue("No eviction is happening", i < 1000);
             }
 
--- a/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
+++ b/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
@@ -52,6 +52,8 @@ import static com.microsoft.appcenter.Flags.PERSISTENCE_CRITICAL;
 import static com.microsoft.appcenter.Flags.PERSISTENCE_NORMAL;
 import static com.microsoft.appcenter.ingestion.models.json.MockLog.MOCK_LOG_TYPE;
 import static com.microsoft.appcenter.persistence.DatabasePersistence.SCHEMA;
+import static org.hamcrest.CoreMatchers.hasItems;
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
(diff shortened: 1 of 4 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -200,6 +206,11 @@ public boolean onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
                 db.execSQL("ALTER TABLE " + TABLE + " ADD COLUMN `" + COLUMN_PRIORITY + "` INTEGER DEFAULT " + PERSISTENCE_NORMAL);
                 return true;
             }
+
+            @Override
+            public void onOpen(SQLiteDatabase db) {
+                db.execSQL("CREATE INDEX IF NOT EXISTS " + INDEX_PRIORITY + " ON " + TABLE + " (" + COLUMN_PRIORITY + ")");
```
jaelim-ms: Can you create index in onCreate and onUpgrade instead of doing this every time when it opens database?
I know it will only create the index if it doesn't exist but the method call itself isn't necessary to be called multiple times on same DB version.

guperrot: I suggest adding onCreate to the Listener and use add index in both callbacks.

guperrot: We don't have to update version as this was never released.

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -92,6 +92,12 @@
     @VisibleForTesting
     static final String COLUMN_PRIORITY = "priority";
 
+    /**
+     * Priority index.
+     */
+    @VisibleForTesting
+    static final String INDEX_PRIORITY = "idx_" + COLUMN_PRIORITY;
```
guperrot: We opted for another convention we found: `ix_logs_priority`.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -425,7 +433,7 @@ long getMaxSize() {
          * @param newVersion new version of the schema.
          * @return true if upgrade was managed, false to drop/create table.
          */
-        @SuppressWarnings({"BooleanMethodIsAlwaysInverted", "unused"})
+        @SuppressWarnings("BooleanMethodIsAlwaysInverted")
```
dhei: Is `BooleanMethodIsAlwaysInverted` still needed?

MatkovIvan: Yes. Look at the single use:
https://github.com/Microsoft/AppCenter-SDK-Android/blob/42dc17f2b4670849d6e4f9e74297c163046d6e26/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java#L141

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -186,6 +192,11 @@ public DatabasePersistence(Context context) {
         mPendingDbIdentifiers = new HashSet<>();
         mDatabaseManager = new DatabaseManager(context, DATABASE, TABLE, version, schema, new DatabaseManager.Listener() {
 
+            @Override
+            public void onCreate(SQLiteDatabase db) {
+                db.execSQL("CREATE INDEX `" + INDEX_PRIORITY + "` ON " + TABLE + " (`" + COLUMN_PRIORITY + "`)");
```
guperrot: nit: this could be shared private method with the upgrade.

guperrot:
Need some time to review changes on DatabasePersistenceAndroidTest.

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/persistence/DatabasePersistenceAndroidTest.java
```diff
@@ -370,27 +369,26 @@ public void putTooManyLogs() throws PersistenceException {
...
+            List<Log> allLogs = new ArrayList<>();
+            String group = "test-p1";
+            for (int i = 0; allLogs.size() == persistence.countLogs(group); i++) {
                 MockLog log = AndroidTestUtils.generateMockLog();
-                persistence.putLog(log, "test-p1", PERSISTENCE_NORMAL);
-                expectedLogs.add(log);
+                persistence.putLog(log, group, PERSISTENCE_NORMAL);
+                allLogs.add(log);
+
+                /* Fail if no eviction happen after a long time to avoid infinite loop on bug. */
```
dhei: "happen" => "happens", same for https://github.com/Microsoft/AppCenter-SDK-Android/pull/871/commits/aa06618b4c812c673b1cdf1bb60a6aace74475de#diff-13c7c0c5167ac2fcdb51679095067306R429

PR #881, opened 2018-11-15
Fix push foreground listener after re-enabling push service

MatkovIvan:
* [x] Did you add unit tests?

## Description

1. Start Sasquatch
2. Go to settings
3. Disable push
4. Restart app
5. Go to settings
6. Enable push
7. Stay on settings screen, do not turn screen off and on
8. Push => no dialog

Reason is that `mActivity` is null in that case for Push class, we need to store/update `mActivity` reference even when Push is disabled so that it's ready at enabled time.

### DIFF
--- a/sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
+++ b/sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
@@ -326,13 +326,7 @@ public class Push extends AbstractAppCenterService {
 
     @Override
     public void onActivityPaused(Activity activity) {
-        postOnUiThread(new Runnable() {
-
-            @Override
-            public void run() {
-                mActivity = null;
-            }
-        });
+        mActivity = null;
     }
 
     private void checkPushInActivityIntent(Activity activity) {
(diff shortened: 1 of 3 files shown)


guperrot:
There was a need for the post command, need to find history before I can elaborate.

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -348,11 +342,11 @@ private void checkPushInActivityIntent(final Activity activity, final Intent int
             AppCenterLog.error(LOG_TAG, "Push.checkLaunchedFromNotification: intent may not be null");
             return;
         }
+        mActivity = activity;
         postOnUiThread(new Runnable() {
 
             @Override
             public void run() {
-                mActivity = activity;
                 checkPushInIntent(intent);
```
guperrot: The post was making sure activity is not null or mismatching the activity where we received the push in foreground.
So this solution is simplifying but to avoid regression you need to pass (final) activity (not mActivity) to checkPushInIntent, the activity snapshot basically for that callback.

MatkovIvan: This is a literal description of what to do 😄

sdk/appcenter-push/src/test/java/com/microsoft/appcenter/push/PushTest.java
```diff
@@ -300,14 +300,21 @@ public void nullTokenOnStartThenRefresh() {
     @Test
     public void receivedInForeground() {
```
guperrot: It would be better to have 2 tests, 1 like before, 1 like after the changes.

PR #890, opened 2018-11-21
Updating setUserId APIs to match new spec

jwallra:

* [ ] Did you add unit tests?

## Description

* `app.userId` becomes `user.localId` again
* If 'c:' prefix is missing for 1st party user id, it is now added automatically (only when sent to One Collector). If using both secrets, crashes will not get automatic prefix in the payload.
* If using another prefix than `c:` for `AppCenter.setUserId` when using target token or both, or when using `PropertyConfigurator`, emit an error and don’t accept `userId` update.
* The `userId` is invalid if empty or just `c:`.
* Rename `setAppUserId` back to `setUserId` like in V1 (for `PropertyConfigurator`)


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
```diff
@@ -248,14 +248,20 @@ private String getUserId() {
...
     public void setUserId(final String userId) {
-        if (UserIdUtils.checkUserIdValidForOneCollector(userId)) {
-            Analytics.getInstance().postCommandEvenIfDisabled(new Runnable() {
-
-                @Override
-                public void run() {
-                    mUserId = UserIdUtils.getPrefixedUserId(userId);
-                }
-            });
+        if (userId == null) {
```
guperrot: why doing that since the code was returning true for null? This is more complex than before.

sdk/appcenter/src/main/java/com/microsoft/appcenter/Constants.java
```diff
@@ -31,7 +31,7 @@
     /**
      * Maximum allowed length for user identifier for App Center server.
      */
-    static final int USER_ID_APP_CENTER_MAX_LENGTH = 256;
+    public static final int USER_ID_APP_CENTER_MAX_LENGTH = 256;
```
guperrot: We should move to UserIdUtils

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdUtils.java
```diff
@@ -32,9 +32,11 @@ public static boolean checkUserIdForOneCollector(String userId) {
                     AppCenterLog.error(LOG_TAG, String.format("userId prefix must be '%s%s', '%s%s:' is not supported.", CUSTOM_PREFIX, PREFIX_SEPARATOR, prefix, PREFIX_SEPARATOR));
                     return false;
                 }
+                return true;
             }
+            return false;
```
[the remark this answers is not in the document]
jwallra: Yes, but this case is catching user ids like `:alice` - we don't allow that, do we?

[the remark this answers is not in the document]
jwallra: See this test https://github.com/Microsoft/AppCenter-SDK-Android/pull/890/files#diff-44334512bde881655cd850e052666320R20

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdUtils.java
```diff
@@ -32,9 +32,11 @@ public static boolean checkUserIdForOneCollector(String userId) {
...
         }
-        return true;
+        return false;
```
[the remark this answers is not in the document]
jwallra: I don't know what commit you commented on, but this is already fixed

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdUtils.java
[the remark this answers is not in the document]
guperrot: ah ok

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
jwallra: Forgot to revert this, I had a commit or two in there where null was invalid

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
```diff
@@ -248,14 +248,20 @@ private String getUserId() {
...
+            mUserId = userId;
```
guperrot: You cannot update mUserId in that thread, please revert as it was before.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java
guperrot: ok

sdk/appcenter/src/main/java/com/microsoft/appcenter/Constants.java
jwallra: It's used in `AppCenterUserIdTest` though to generate long user ids

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdUtils.java
```diff
@@ -0,0 +1,72 @@
...
+import static com.microsoft.appcenter.AppCenter.LOG_TAG;
+
+/**
+ * Utility to store and retrieve values for user identifiers.
+ */
+public class UserIdUtils {
+
+    private static final String CUSTOM_PREFIX = "c";
+
+    private static final String PREFIX_SEPARATOR = ":";
```
jaelim-ms: We are using the same separator for authentication. In CS 3.0 ":" is used for all prefixed values, we can move this to a constant class and use one definition for everything.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdUtils.java
```diff
@@ -0,0 +1,72 @@
...
+    public static final int USER_ID_APP_CENTER_MAX_LENGTH = 256;
+
+    /**
+     * Check if userId is valid for One Collector.
+     *
+     * @param userId userId.
+     * @return true if valid, false otherwise.
+     */
+    public static boolean checkUserIdValidForOneCollector(String userId) {
+        if (userId != null) {
```
jaelim-ms: You can reduce the nested `if` here.
```
if (userId == null) {
  return true;
}

File: sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdUtils.java (hunk 3 above)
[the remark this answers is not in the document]
Comment (jaelim-ms) [reply sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdUtils.java]: Have you tested `:alice`? `:` can be treated as a normal character then it should be a valid userId.

**Comment by jaelim-ms** [comment]
For my [comment](https://github.com/Microsoft/AppCenter-SDK-Android/pull/890#discussion_r236358407), I tested `:alice` and it is accepted.

File: sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdUtils.java (hunk 6 above)
Comment (jwallra) [reply sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdUtils.java]: We have this same string defined in a few places, but they're not related.  I don't see one defined for the common schema separator anywhere.  Since it's only used here so far, I think it's OK to define here, the deepest layer at which it's used.  Am I missing something?

Or, are you saying that other fields which we don't have yet will also use this so we should put it in constants?

**Comment by jwallra** [comment]
> For my [comment](https://github.com/Microsoft/AppCenter-SDK-Android/pull/890#discussion_r236358407), I tested `:alice` and it is accepted.

@jaelim-ms In our refactor PR (which will be merged first) we test this [here](https://github.com/Microsoft/AppCenter-SDK-Android/pull/891/files#diff-c288effc2dd145443ca3c81fa066775dR15) and verify that it isn't accepted.

File: sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdUtils.java (hunk 3 above)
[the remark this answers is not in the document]
Comment (guperrot) [reply sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdUtils.java]: Invalid, fixed.

File: sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdUtils.java (hunk 6 above)
Comment (guperrot) [reply sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdUtils.java]: Let's move both occurrences to constants, I'm doing the patch now.

File: sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/PropertyConfigurator.java [hunk 8]
```diff
@@ -231,32 +232,33 @@ public void run() {
...
-        return mAppUserId;
+    private String getUserId() {
+        return mUserId;
     }
 
     /**
-     * Override common schema Part A property App.UserId.
+     * Set the user identifier.
+     * The user identifier needs to start with the c: prefix or must not have a prefix.
+     * If the prefix is missing, the c: prefix it will be automatically added.
```
dhei: wording "the c: prefix will be automatically added."

PR #891, opened 2018-11-21
Refactor UserIdUtil into UserIdContext

jwallra:

* [ ] Did you add unit tests?

## Description

A small refactor to the local user id spec change.


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -749,7 +750,7 @@ private void queuePage(String name, Map<String, String> properties) {
      * @param flags              optional flags.
      */
     private synchronized void trackEventAsync(final String name, final List<TypedProperty> properties, final AnalyticsTransmissionTarget transmissionTarget, final int flags) {
-        final String userId = AppCenter.getInstance().getUserId();
+        final String userId = UserIdContext.getInstance().getUserId();
```
jaelim-ms: Not related to this PR but userId assignment could be done in Channel since all logs have userId property.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdContext.java
```diff
@@ -19,6 +19,33 @@
     @VisibleForTesting
     public static final int USER_ID_APP_CENTER_MAX_LENGTH = 256;
 
+    /**
+     * Unique instance.
+     */
+    private static UserIdContext sInstance;
```
jaelim-ms: `UserIdContext` only has one string property which is `userId`. I don't think `getInstance()` is necessary unless we will have other properties here in the future.
`private static String sUserId;`??

jwallra: @jaelim-ms that's how we originally implemented it in #890 - we wanted to align with the Apple implementation.  Do you think we should close this instead of merging it?

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
guperrot: Its not the spec though.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdContext.java
guperrot: @jaelim-ms there could be history later for breakpad and aligning with Apple and best practices for shared objects.

jaelim-ms: @guperrot Makes sense.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/UserIdContext.java
```diff
@@ -7,7 +7,7 @@
 /**
  * Utility to store and retrieve values for user identifiers.
```
dhei: nit-picking, it's not "Utility" in JavaDoc any more after the renaming.

PR #896, opened 2018-11-27
Fix logs duplication

MatkovIvan:
* [x] Did you add unit tests?

## Description

The SDK doesn't cancel http request correctly when a network connection is lost.

### DIFF
--- a/sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/AbstractDistributeTest.java
+++ b/sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/AbstractDistributeTest.java
@@ -15,6 +15,8 @@ import com.microsoft.appcenter.AppCenter;
 import com.microsoft.appcenter.AppCenterHandler;
 import com.microsoft.appcenter.channel.Channel;
 import com.microsoft.appcenter.distribute.channel.DistributeInfoTracker;
+import com.microsoft.appcenter.http.HttpClient;
+import com.microsoft.appcenter.http.HttpUtils;
 import com.microsoft.appcenter.utils.AppCenterLog;
 import com.microsoft.appcenter.utils.AppNameHelper;
 import com.microsoft.appcenter.utils.HandlerUtils;
@@ -32,6 +34,7 @@ import org.mockito.Mock;
 import org.mockito.internal.stubbing.answers.Returns;
 import org.mockito.invocation.InvocationOnMock;
 import org.mockito.stubbing.Answer;
+import org.powermock.api.mockito.PowerMockito;
 import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.rule.PowerMockRule;
 import org.powermock.reflect.Whitebox;
@@ -52,7 +55,23 @@ import static org.powermock.api.mockito.PowerMockito.mockStatic;
 import static org.powermock.api.mockito.PowerMockito.whenNew;
 
 @SuppressWarnings({"WeakerAccess", "CanBeFinal"})
-@PrepareForTest({Distribute.class, SharedPreferencesManager.class, AppNameHelper.class, AppCenterLog.class, AppCenter.class, NetworkStateHelper.class, BrowserUtils.class, UUIDUtils.class, ReleaseDetails.class, TextUtils.class, CryptoUtils.class, InstallerUtils.class, Toast.class, HandlerUtils.class})
+@PrepareForTest({
+        AppCenter.class,
+        AppCenterLog.class,
+        AppNameHelper.class,
+        BrowserUtils.class,
+        CryptoUtils.class,
+        Distribute.class,
+        HandlerUtils.class,
+        HttpUtils.class,
+        InstallerUtils.class,
+        NetworkStateHelper.class,
+        ReleaseDetails.class,
+        SharedPreferencesManager.class,
+        TextUtils.class,
+        Toast.class,
+        UUIDUtils.class
+})
 public class AbstractDistributeTest {
 
     static final String TEST_HASH = HashUtils.sha256("com.contoso:1.2.3:6");
@@ -108,6 +127,9 @@ public class AbstractDistributeTest {
 
     @Mock
     DistributeInfoTracker mDistributeInfoTracker;
+
+    @Mock
+    HttpClient mHttpClient;
 
     @Before
     @SuppressLint("ShowToast")
@@ -177,6 +199,8 @@ public class AbstractDistributeTest {
         mockStatic(NetworkStateHelper.class);
         mNetworkStateHelper = mock(NetworkStateHelper.class, new Returns(true));
         when(NetworkStateHelper.getSharedInstance(any(Context.class))).thenReturn(mNetworkStateHelper);
+        PowerMockito.spy(HttpUtils.class);
+        PowerMockito.doReturn(mHttpClient).when(HttpUtils.class, "createHttpClient", any(Context.class));
 
         /* Mock some statics. */
         mockStatic(BrowserUtils.class);
--- a/sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/AbstractDistributeTest.java
+++ b/sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/AbstractDistributeTest.java
@@ -34,7 +34,6 @@ import org.mockito.Mock;
 import org.mockito.internal.stubbing.answers.Returns;
 import org.mockito.invocation.InvocationOnMock;
 import org.mockito.stubbing.Answer;
-import org.powermock.api.mockito.PowerMockito;
 import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.rule.PowerMockRule;
 import org.powermock.reflect.Whitebox;
@@ -51,7 +50,9 @@ import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 import static org.powermock.api.mockito.PowerMockito.doAnswer;
+import static org.powermock.api.mockito.PowerMockito.doReturn;
 import static org.powermock.api.mockito.PowerMockito.mockStatic;
+import static org.powermock.api.mockito.PowerMockito.spy;
 import static org.powermock.api.mockito.PowerMockito.whenNew;
 
 @SuppressWarnings({"WeakerAccess", "CanBeFinal"})
@@ -199,8 +200,8 @@ public class AbstractDistributeTest {
         mockStatic(NetworkStateHelper.class);
         mNetworkStateHelper = mock(NetworkStateHelper.class, new Returns(true));
         when(NetworkStateHelper.getSharedInstance(any(Context.class))).thenReturn(mNetworkStateHelper);
-        PowerMockito.spy(HttpUtils.class);
-        PowerMockito.doReturn(mHttpClient).when(HttpUtils.class, "createHttpClient", any(Context.class));
+        spy(HttpUtils.class);
+        doReturn(mHttpClient).when(HttpUtils.class, "createHttpClient", any(Context.class));
 
         /* Mock some statics. */
         mockStatic(BrowserUtils.class);
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClient.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClient.java
@@ -11,11 +11,14 @@ import com.microsoft.appcenter.utils.HandlerUtils;
 
 import org.json.JSONObject;
 
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Reader;
+import java.io.UnsupportedEncodingException;
+import java.io.Writer;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.util.Map;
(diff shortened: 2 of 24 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClient.java
```diff
@@ -298,11 +297,21 @@ public void run() {
...
+                    return call.isHttpResult(call.get());
+                } catch (Exception ignored) {
+                    return false;
                 }
             }
+
+            @Override
+            public void cancel() {
+
+                // FIXME This doesn't kill the AsyncTask!
```
guperrot: Yeah and basically we shouldn't cancel on network outage, let call fail, and if it somehow succeeds, accept callback, we don't need all these changes, we should instead avoid cancelling anything from network state decorator if already running, and avoid resuming call if still running.

I don't understand this PR at all overall and should use a different approach of dealing with network state but not in the way you modified so far. Let call run, just pause new calls, forward success callback to channel, let channel accept it.

guperrot: For the cancel part, I would keep the code as before and cancel would be kept only for disabling scenario (where we delete logs), but as said on previous comment, don't call cancel on network down.

MatkovIvan:
Summary for review:
* Checking if task is cancelled during HTTP call;
* Do not cancel the calls on losing network (based on @guperrot feedback);
* Move cancelling the calls on disabling AppCenter to DefaultHttpClient;
* Move call async task into a separate file to avoid circular reference to Tracker interface;
* Handle one more recoverable exception;
* Reorder decorators to avoid retries in offline;

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/AbstractDistributeTest.java
```diff
@@ -177,6 +199,8 @@ public Void answer(InvocationOnMock invocation) {
         mockStatic(NetworkStateHelper.class);
         mNetworkStateHelper = mock(NetworkStateHelper.class, new Returns(true));
         when(NetworkStateHelper.getSharedInstance(any(Context.class))).thenReturn(mNetworkStateHelper);
+        PowerMockito.spy(HttpUtils.class);
```
guperrot: nit: prefer static imports for mock APIs

jaelim-ms: Usually spy takes an instance. Is it working?

sdk/appcenter/src/test/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandlerTest.java
```diff
@@ -95,8 +92,12 @@ public ServiceCall answer(InvocationOnMock invocationOnMock) {
         NetworkStateHelper networkStateHelper = mock(NetworkStateHelper.class);
         when(networkStateHelper.isNetworkConnected()).thenReturn(true);
 
+        /* Simulate state updated events first. It should no affect behaviour. */
```
jaelim-ms: `It should no affect behaviour` -> `It should not affect behavior`

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClientCallTask.java
```diff
@@ -0,0 +1,337 @@
...
+
+    /**
+     * Read buffer size.
+     */
+    private static final int READ_BUFFER_SIZE = 1024;
+
+    /**
+     * Write buffer size.
+     */
+    private static final int WRUTE_BUFFER_SIZE = 1024;
```
guperrot: typo

sdk/appcenter/src/test/java/com/microsoft/appcenter/ingestion/AppCenterIngestionTest.java
```diff
@@ -45,14 +47,26 @@
...
     @Rule
-    public PowerMockRule rule = new PowerMockRule();
+    public PowerMockRule mRule = new PowerMockRule();
+
+    @Mock
+    private HttpClient mHttpClient;
+
+    @Before
+    public void setUp() throws Exception {
+        PowerMockito.spy(HttpUtils.class);
```
guperrot: nit: use static imports

sdk/appcenter/src/test/java/com/microsoft/appcenter/ingestion/OneCollectorIngestionTest.java
```diff
@@ -78,9 +80,14 @@
     @Captor
     private ArgumentCaptor<Map<String, String>> mHeadersCaptor;
 
+    @Mock
+    private HttpClient mHttpClient;
+
     @Before
     public void setUp() throws Exception {
         TicketCache.clear();
+        PowerMockito.spy(HttpUtils.class);
```
guperrot: same here

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClientCallTask.java
```diff
@@ -0,0 +1,337 @@
...
+
+        /* Handle the result even if it was cancelled. */
+        if (isHttpResult(result)) {
+            onPostExecute(result);
+        } else {
+            mTracker.onFinish(this);
+        }
+    }
+
+    interface Tracker {
```
jaelim-ms: This definition seems weird. Should it be defined in DefaultHttpClient.java?

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
```diff
@@ -67,82 +64,34 @@ public void reopen() {
 
     @Override
     public synchronized void onNetworkStateUpdated(boolean connected) {
-        if (connected) {
+        if (connected && mCalls.size() > 0) {
             AppCenterLog.debug(LOG_TAG, "Network is available. " + mCalls.size() + " pending call(s) to submit now.");
-        } else {
-            AppCenterLog.debug(LOG_TAG, "Network is down. Pausing " + mCalls.size() + " network call(s).");
```
jaelim-ms: We still need a log that says network is down.

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
```diff
@@ -67,82 +64,34 @@ public void reopen() {
 
     @Override
     public synchronized void onNetworkStateUpdated(boolean connected) {
-        if (connected) {
+        if (connected && mCalls.size() > 0) {
```
jaelim-ms: Can you log "Network is available" when it is connected? Split below the log message or you can print below log message even though `mCalls.size() == 0`.

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClientCallTask.java
```diff
@@ -0,0 +1,337 @@
...
+
```
jaelim-ms: Can you add a doc?

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
guperrot: According to me, it's better now. When you have multiple apps in background it's always spamming logs otherwise.

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClientCallTask.java
```diff
@@ -0,0 +1,337 @@
...
+import static com.microsoft.appcenter.http.DefaultHttpClient.CHARSET_NAME;
+import static com.microsoft.appcenter.http.DefaultHttpClient.CONTENT_ENCODING_KEY;
+import static com.microsoft.appcenter.http.DefaultHttpClient.CONTENT_ENCODING_VALUE;
+import static com.microsoft.appcenter.http.DefaultHttpClient.CONTENT_TYPE_KEY;
+import static com.microsoft.appcenter.http.DefaultHttpClient.CONTENT_TYPE_VALUE;
+import static com.microsoft.appcenter.http.DefaultHttpClient.METHOD_POST;
+import static java.lang.Math.max;
+import static java.lang.Math.min;
+
+class DefaultHttpClientCallTask extends AsyncTask<Void, Void, Object> {
```
guperrot: Should add Javadoc comments where missing

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClientCallTask.java
```diff
@@ -0,0 +1,337 @@
...
+            /* Generate exception on failure. */
+            throw new HttpException(status, response);
+        } finally {
+
+            /* Release connection. */
+            urlConnection.disconnect();
+        }
+    }
+
+    private boolean isHttpResult(Object result) {
```
guperrot: I would inline this method

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientCallDecorator.java
```diff
@@ -5,7 +5,7 @@
 /**
```
guperrot: This class has warnings `access can be private`.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/AbstractDistributeTest.java
MatkovIvan: The target here is mock static method

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
```diff
@@ -52,9 +52,6 @@ public synchronized ServiceCall callAsync(String url, String method, Map<String,
     @Override
     public synchronized void close() throws IOException {
         mNetworkStateHelper.removeListener(this);
-        for (Call call : mCalls) {
```
guperrot: This is worth commenting here why we don't do that anymore.

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
MatkovIvan: It's logged in `NetworkStateHelper`

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
MatkovIvan: Network availability is logged in `NetworkStateHelper`

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientCallDecorator.java
```diff
@@ -5,7 +5,7 @@
...
  * Helper class used to share logic with multiple decorators.
  */
-abstract class HttpClientCallDecorator implements Runnable, ServiceCall, ServiceCallback {
+class HttpClientCallDecorator implements Runnable, ServiceCall, ServiceCallback {
```
guperrot: I can build just fine with abstract, please restore it.

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
```diff
@@ -67,82 +64,34 @@ public void reopen() {
...
         }
 
         @Override
-        public void run() {
-            callRunAsync(this);
-        }
-
-        @Override
-        public void cancel() {
+        public synchronized void cancel() {
```
guperrot: synchronized not needed here, its the parent class that uses it

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClientCallTask.java
MatkovIvan: Ok. Just in some previous state it was used in two places

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClient.java
```diff
@@ -56,229 +36,31 @@
...
-            urlConnection.disconnect();
-        }
-    }
+    @VisibleForTesting
+    Set<DefaultHttpClientCallTask> mTasks = new HashSet<>();
 
     @Override
-    public ServiceCall callAsync(String url, String method, Map<String, String> headers, CallTemplate callTemplate, final ServiceCallback serviceCallback) {
-        final Call call = new Call(url, method, headers, callTemplate, serviceCallback);
+    public synchronized ServiceCall callAsync(String url, String method, Map<String, String> headers, CallTemplate callTemplate, final ServiceCallback serviceCallback) {
```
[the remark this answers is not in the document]
MatkovIvan: Yep. This is not required now

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClientCallTask.java
MatkovIvan: I don't think so. It's callback interface for this class.

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
MatkovIvan: I believe you mean the comment in `onNetworkStateUpdated`

Issue #917, opened 2018-12-18
TLS 1.2 issue on specific devices below API level 21


bentmar:
### **Description**

Appcenter fails to send data to server when running kitkat with tls 1.2 not enabled by default

### **Repro Steps**

Use appcenter (analytics + crashes is what I use) on a Xamarin client. 
Run on a KitKat android device that doesnt have tls 1.2 enabled by default
Track some data using appcenter api
See these error logs in debug output:

`
12-18 12:19:05.372 W/AppCenter( 4447): Try #3 failed and will be retried in 1077322 ms
12-18 12:19:05.372 W/AppCenter( 4447): javax.net.ssl.SSLException: SSL handshake aborted: ssl=0x6ec7d2f8: I/O error during system call, Connection reset by peer
12-18 12:19:05.372 W/AppCenter( 4447): 	at com.android.org.conscrypt.NativeCrypto.SSL_do_handshake(Native Method)
12-18 12:19:05.372 W/AppCenter( 4447): 	at com.android.org.conscrypt.OpenSSLSocketImpl.startHandshake(OpenSSLSocketImpl.java:486)
12-18 12:19:05.372 W/AppCenter( 4447): 	at com.android.okhttp.Connection.upgradeToTls(Connection.java:146)
12-18 12:19:05.372 W/AppCenter( 4447): 	at com.android.okhttp.Connection.connect(Connection.java:107)
12-18 12:19:05.372 W/AppCenter( 4447): 	at com.android.okhttp.internal.http.HttpEngine.connect(HttpEngine.java:294)
12-18 12:19:05.372 W/AppCenter( 4447): 	at com.android.okhttp.internal.http.HttpEngine.sendSocketRequest(HttpEngine.java:255)
12-18 12:19:05.372 W/AppCenter( 4447): 	at com.android.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:206)
12-18 12:19:05.372 W/AppCenter( 4447): 	at com.android.okhttp.internal.http.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:345)
12-18 12:19:05.372 W/AppCenter( 4447): 	at com.android.okhttp.internal.http.HttpURLConnectionImpl.connect(HttpURLConnectionImpl.java:89)
[...]

PR #918, opened 2018-12-19
Fix TLS 1.2 on some Android 4.x devices

MatkovIvan:
* [x] Did you add unit tests?

## Description

## Related PRs or issues

#917

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClientCallTask.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClientCallTask.java
@@ -174,6 +174,14 @@ class DefaultHttpClientCallTask extends AsyncTask<Void, Void, Object> {
              * Make sure we use TLS 1.2 when the device supports it but not enabled by default.
              * Don't hardcode TLS version when enabled by default to avoid unnecessary wrapping and
              * to support future versions of TLS such as say 1.3 without having to patch this code.
+             *
+             * TLS 1.2 was enabled by default only on Android 5.0:
+             * https://developer.android.com/about/versions/android-5.0-changes#ssl
+             * https://developer.android.com/reference/javax/net/ssl/SSLSocket#default-configuration-for-different-android-versions
+             *
+             * There is a problem that TLS 1.2 is still disabled by default on some Samsung devices
+             * with API 21, so apply the rule to this API level as well.
+             * See https://github.com/square/okhttp/issues/2372#issuecomment-244807676
              */
             if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.LOLLIPOP) {
                 urlConnection.setSSLSocketFactory(new TLS1_2SocketFactory());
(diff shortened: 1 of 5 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClientCallTask.java
```diff
@@ -175,7 +175,7 @@ private String doHttpCall() throws Exception {
              * Don't hardcode TLS version when enabled by default to avoid unnecessary wrapping and
              * to support future versions of TLS such as say 1.3 without having to patch this code.
              */
-            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.KITKAT_WATCH) {
+            if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.LOLLIPOP) {
```
guperrot: It should be `< LOLLIPOP`, not `<=`, TLS 1.2 was enabled on 5.0: https://developer.android.com/about/versions/android-5.0-changes#ssl.

What is weird is that API level 19 is already covered by this patch (kitkat), [kitkat watch is API level 20](https://developer.android.com/reference/android/os/Build.VERSION_CODES#KITKAT_WATCH), not 19.
TLS 1.2 should have been enabled on API level 20 according to https://developer.android.com/reference/javax/net/ssl/SSLSocket#default-configuration-for-different-android-versions.

Have you tried to repro the error?

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/TLS1_2SocketFactory.java
```diff
@@ -14,10 +17,30 @@
...
      * Protocols that we allow.
      */
-    private static final String[] ENABLED_PROTOCOLS = {"TLSv1.2"};
+    private static final String[] ENABLED_PROTOCOLS = { TLS1_2_PROTOCOL };
+
+    private SSLSocketFactory mSocketFactory;
+
+    TLS1_2SocketFactory() {
+        try {
+            SSLContext sc = SSLContext.getInstance(TLS1_2_PROTOCOL);
```
guperrot: Why do we need this?

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/DefaultHttpClientCallTask.java
MatkovIvan: Based on https://github.com/square/okhttp/issues/2372#issuecomment-244807676
> the same issue also happens on some Samsung devices with API 21

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/TLS1_2SocketFactory.java
MatkovIvan: 1. #917
2. All the materials I found on the topic use this

This is the main version of #917. So opposite, I wonder why it wasn't here.
Also, I thought about lack of `ProviderInstaller.installIfNeeded` call. But this requires play-services dependency. So, it should be discussed first.

PR #921, opened 2019-01-09
Don't start session from logs

MatkovIvan:
* [x] Did you add unit tests?

## Description

1. Schedule track event every 30s
2. Minimize the app into the background
3. Observe that every event creates a new session

Expected behavior:  
Like iOS, sending a log just extends current session lifetime but must not trigger a new session.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,16 @@
 # App Center SDK for Android Change Log
 
-## Version 1.11.1
+## Version 1.11.1 (Under active development)
+
+### AppCenterAnalytics
+
+* **[Fix]** Don't start a new session from logs in the background.
 
 ### AppCenterDistribute
 
 * **[Fix]** Fix issue with forcing Chrome to open links when other browsers are the default.
+
+___
 
 ## Version 1.11.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### AppCenterAnalytics
 
-* **[Fix]** Don't start a new session from logs in the background.
+* **[Fix]** Extend the current session instead of starting a new session when sending events from the background.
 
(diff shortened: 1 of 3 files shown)


sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/channel/SessionTrackerTest.java
```diff
@@ -380,15 +387,13 @@ public void sdkConfiguredBetweenPauseAndResume() {
             verify(mChannel).enqueue(expectedStartSessionLog, TEST_GROUP, DEFAULTS);
         }
 
-        /* No usage from background for a long time: new session. */
+        /* No usage from background for a long time: same session. */
```
jaelim-ms: Can you split this test?

CHANGELOG.md
```diff
@@ -1,11 +1,17 @@
 # App Center SDK for Android Change Log
 
-## Version 1.11.1
+## Version 1.11.1 (Under active development)
+
+### AppCenterAnalytics
+
+* **[Fix]** Don't start a new session from logs in the background.
```
dhei: Maybe something like "Extend the current session instead of starting a new session when sending events from the background." for better clarity.

CHANGELOG.md
```diff
@@ -1,11 +1,17 @@
 # App Center SDK for Android Change Log
 
-## Version 1.11.1
+## Version 1.11.1 (Under active development)
+
+### AppCenterAnalytics
+
+* **[Fix]** Extend the current session instead of starting a new session when sending events from the background.
```
guperrot: Sessions are also no longer started in background by sending an event or a log from another service such as push, as a consequence the push registration information will be missing from crash events information.

Issue #929, opened 2019-01-16
Crash with NullPointerException on launch


kvpt:
### **Description**
App crash on start when distribution is activated.

### **Repro Steps**
Please list the steps used to reproduce your issue.

1. Install the app with AppCenter and Distribution activated
2. The browser open
3. Minimize all
4. Return to the app
5. The App Crash (I only reproduce it on one device)

### **Details**
1. Which SDK version are you using?
    - e.g. 1.11.0
2. Which OS version did you experience the issue on?
    - Android 7.1.2
3. What device version did you see this error on?  Were you using an emulator or a physical device?
    - Redmi 4X
4. What third party libraries are you using?
    - Xamarin.Forms
5. Please enable verbose logging for your app using `AppCenter.setLogLevel(Log.VERBOSE)` before your call to `AppCenter.start(...)` and include the logs here:

The stacktrace of the crash :
`java.lang.RuntimeException: Unable to start activity ComponentInfo{com.app/com.microsoft.appcenter.distribute.DeepLinkActivity}: java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.String android.content.SharedPreferences.getString(java.lang.String, java.lang.String)' on a null object reference
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2723)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2784)
	at android.app.ActivityThread.-wrap12(ActivityThread.java)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1523)
	at android.os.Handler.dispatchMessage(Handler.java:102)
	at android.os.Looper.loop(Looper.java:163)
	at android.app.ActivityThread.main(ActivityThread.java:6238)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:794)
Caused by: java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.String android.content.SharedPreferences.getString(java.lang.String, java.lang.String)' on a null object reference
	at com.microsoft.appcenter.utils.storage.SharedPreferencesManager.getString(SharedPreferencesManager.java:205)
	at com.microsoft.appcenter.utils.storage.SharedPreferencesManager.getString(SharedPreferencesManager.java:194)
[...]

PR #931, opened 2019-01-17
Fix exception if we receive deep link intent with setup failure before onStart

MatkovIvan:
* [x] Did you add unit tests?

## Description

Fix exception if we receive deep link intent with setup failure before `onStart`.

## Related PRs or issues

#929


MatkovIvan:
@guperrot There is no need to delay start and prevent to open the browser. Tested.

PR #934, opened 2019-01-23
Fix compatibility with Android Studio 3.3

MatkovIvan:
* [ ] Did you add unit tests?

## Description

Fix compatibility with Android Studio 3.3

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,10 +13,10 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.2.1'
+        classpath 'com.android.tools.build:gradle:3.3.0'
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.0'
         classpath 'com.github.dcendents:android-maven-gradle-plugin:2.0'
-        classpath 'com.google.gms:google-services:4.0.1'
+        classpath 'com.google.gms:google-services:4.2.0'
         classpath 'org.kt3k.gradle.plugin:coveralls-gradle-plugin:2.8.2'
     }
 }
--- a/scripts/publish-github-release.sh
+++ b/scripts/publish-github-release.sh
@@ -5,7 +5,7 @@ GITHUB_ACCESS_TOKEN=${1:-$GITHUB_ACCESS_TOKEN}
 GITHUB_API_URL_TEMPLATE="https://%s.github.com/repos/%s/%s?access_token=%s%s"
 GITHUB_API_HOST="api"
 GITHUB_UPLOAD_HOST="uploads"
-BINARY_FILE_FILTER="*release.aar"
+BINARY_FILE_FILTER="*.aar"
 JQ_COMMAND=jq
 PUBLISH_VERSION="$(grep "versionName = '" versions.gradle | awk -F "[']" '{print $2}')"
 
--- a/sdk/appcenter-push/build.gradle
+++ b/sdk/appcenter-push/build.gradle
@@ -6,4 +6,7 @@ dependencies {
 
     api "com.google.firebase:firebase-core:16.0.1"
     api "com.google.firebase:firebase-messaging:17.0.0"
+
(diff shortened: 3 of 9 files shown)


build.gradle
```diff
@@ -13,10 +13,10 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.2.1'
+        classpath 'com.android.tools.build:gradle:3.3.0'
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.0'
         classpath 'com.github.dcendents:android-maven-gradle-plugin:2.0'
-        classpath 'com.google.gms:google-services:4.0.1'
```
guperrot: Can you comment on this change? Is this necessary for push? Do we have to update docs and portal for the google plugin version?

sdk/appcenter-push/build.gradle
```diff
@@ -6,4 +6,7 @@ dependencies {
 
     api "com.google.firebase:firebase-core:16.0.1"
     api "com.google.firebase:firebase-messaging:17.0.0"
+
+    // Resolve version conflict with firebase-core (support-v4:26.1.0) and compile-only support-annotations
+    implementation "com.android.support:support-v4:${rootProject.ext.supportLibVersion}"
```
guperrot: Did you check bintray upload does not expose the dependency for that one?

sdk/build.gradle
```diff
@@ -90,13 +90,14 @@ subprojects {
...
+        sourceDirectories.from = files(["$projectDir/src/main/java"])
+        classDirectories.from = files([
                 fileTree(dir: "$buildDir/intermediates/classes/debug", excludes: fileFilter),
                 fileTree(dir: "$buildDir/intermediates/javac/debug/compileDebugJavaWithJavac/classes", excludes: fileFilter)
         ])
-        executionData = fileTree(dir: buildDir, includes: [
+        executionData.from = fileTree(dir: buildDir, includes: [
                 'jacoco/testDebugUnitTest.exec',
+                'outputs/code_coverage/debugAndroidTest/connected/*coverage.ec',
                 'outputs/code-coverage/connected/*coverage.ec'
```
guperrot: We keep the old one for previous android versions or is there another reason?

sdk/appcenter-push/build.gradle
guperrot: Exposed but as runtime dependency, not sure about the impact, can you check?

build.gradle
guperrot: It's ok to test latest versions, just wanted to make sure it was not a requirement. If you already tested push we can keep it.

sdk/build.gradle
guperrot: Let's add a comment then.

guperrot: Hum can you test with only the new path? Since we use the latest android plugin it should not have to handle previous versions.

scripts/publish-github-release.sh
```diff
@@ -5,7 +5,7 @@ GITHUB_ACCESS_TOKEN=${1:-$GITHUB_ACCESS_TOKEN}
 GITHUB_API_URL_TEMPLATE="https://%s.github.com/repos/%s/%s?access_token=%s%s"
 GITHUB_API_HOST="api"
 GITHUB_UPLOAD_HOST="uploads"
-BINARY_FILE_FILTER="*release.aar"
+BINARY_FILE_FILTER="*.aar"
```
guperrot: We should look for possible impact on xamarin unity build.sh -t=externals before merging.

guperrot: Yeah we have multiple references to `-release.aar` on the .NET repo, we should have the PR ready (don't merge) before merging this.

We might want to check Unity repo too.

MatkovIvan:
image

scripts/publish-github-release.sh
MatkovIvan: It's a bug: https://github.com/gradle/gradle/issues/8328
Let's revert gradle to 5.0 for now. Updated masks in the scripts are correct for both versions.

PR #936, opened 2019-01-26
Fix push on develop

guperrot:
Push is broken on develop with the changes from Android Studio 3.3.

### DIFF
--- a/sdk/appcenter-push/build.gradle
+++ b/sdk/appcenter-push/build.gradle
@@ -4,6 +4,8 @@ evaluationDependsOn(':sdk')
 dependencies {
     api project(':sdk:appcenter')
 
+    //noinspection GradleDependency cannot update this without sync error, but app can update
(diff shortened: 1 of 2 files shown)


sdk/appcenter-push/build.gradle
```diff
@@ -4,6 +4,8 @@ evaluationDependsOn(':sdk')
 dependencies {
     api project(':sdk:appcenter')
 
+    //noinspection GradleDependency cannot update this without sync error, but app can update
     api "com.google.firebase:firebase-core:16.0.1"
+    //noinspection GradleDependency cannot update this without sync error, but app can update
```
jwallra: Is this something that we need to document?  It could be confusing for people to see AC uses an older firebase version and then not knowing if they can update.

guperrot: No it's not documented, it should be a separate task though, this comment is not adding anything new compared to master and this PR needs to be merged asap as develop broken.

PR #937, opened 2019-01-28
Addition to #932

MatkovIvan:
* [x] Did you add unit tests?

## Description

Addition to #932

## Related PRs or issues

#932

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
@@ -299,7 +299,7 @@ public class DatabaseManager implements Closeable {
     @Override
     public void close() {
         try {
-            getDatabase().close();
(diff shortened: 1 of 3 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -350,9 +352,14 @@ SQLiteDatabase getDatabase() {
         try {
             return mSQLiteOpenHelper.getWritableDatabase();
         } catch (RuntimeException e) {
+            AppCenterLog.warn(LOG_TAG, "Failed to open database. Trying to delete database (may be corrupted).", e);
 
             /* First error, try to delete database (may be corrupted). */
-            mContext.deleteDatabase(mDatabase);
+            if (mContext.deleteDatabase(mDatabase)) {
```
ElektrojungeMS: We should document this behavior to let customers know that we only ever delete data to recover from this fatal error scenario.

PR #817, opened 2018-09-25
Work in progress for pause/resume transmisssion target

guperrot:

PR #940, opened 2019-02-02
Expose title message when background Push notification is pressed

Quan-Jin:

* [ ] Did you add unit tests?

## Description

1. Add a private set to store the received push intents
2. Use message id to get current intent when click the notification from back end

## Related PRs or issues

[#817](https://github.com/Microsoft/AppCenter-SDK-DotNet/issues/817)

### DIFF
--- a/sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
+++ b/sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
@@ -433,10 +433,9 @@ public class Push extends AbstractAppCenterService {
                 PushNotifier.handleNotification(context, pushIntent);
             }
 
-            if(mReceivedIntents == null){
+            if (mReceivedIntents == null) {
                 mReceivedIntents = new HashSet<>();
             }
-
             mReceivedIntents.add(pushIntent);
         } else {
             final PushNotification notification = new PushNotification(pushIntent);
--- a/sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
+++ b/sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
@@ -380,7 +380,7 @@ public class Push extends AbstractAppCenterService {
                 AppCenterLog.debug(LOG_TAG, "Push intent extras=" + intent.getExtras());
 
                 Intent currentPushIntent = null;
-                if (mReceivedIntents != null) {
+                if (mReceivedIntents != null && !mReceivedIntents.isEmpty()) {
                     for (Intent receivedIntent : mReceivedIntents) {
                         if (googleMessageId.equals(PushIntentUtils.getMessageId(receivedIntent))) {
                             currentPushIntent = receivedIntent;
--- a/sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
+++ b/sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
@@ -23,6 +23,7 @@ import com.microsoft.appcenter.utils.UserIdContext;
 import com.microsoft.appcenter.utils.async.AppCenterFuture;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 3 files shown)


jinquan2019:
Can anyone help to share the error log for failed build because I have no permission to view the build log? Then I can try to investigate the build error for it.

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -79,6 +80,12 @@
      */
     private String mFirstGoogleMessageId;
 
+    /**
+     * Save Received message Intents to generate push notification
+     * when click notification from background
+     */
+    private HashSet<Intent> mReceivedIntents;
```
guperrot: Please use Map<String, Intent> where String is the googleMessageId, that way we can avoid the for loop.

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -367,10 +375,21 @@ private synchronized void checkPushInIntent(Activity activity, Intent intent) {
...
                 }
-                PushNotification notification = new PushNotification(intent);
+
                 AppCenterLog.info(LOG_TAG, "Clicked push message from background id=" + googleMessageId);
                 mLastGoogleMessageId = googleMessageId;
                 AppCenterLog.debug(LOG_TAG, "Push intent extras=" + intent.getExtras());
+
+                Intent currentPushIntent = intent;
+                for (Intent receivedIntent : mReceivedIntents) {
+                    if (googleMessageId.equals(PushIntentUtils.getMessageId(receivedIntent))) {
```
guperrot: Switch to map as mentioned in the other comment.
Also please add a comment every time there is a newline to start a new code block.

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -367,10 +375,21 @@ private synchronized void checkPushInIntent(Activity activity, Intent intent) {
...
+                        currentPushIntent = receivedIntent;
+                        mReceivedIntents.remove(receivedIntent);
+                        break;
+                    }
+                }
+
+                PushNotification notification = new PushNotification(currentPushIntent);
```
guperrot: Please add a block comment or remove newline.

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
guperrot: Please also add final and use interface and not implementation, also init it directly here with default constructor.

Quan-Jin: Updated

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
Quan-Jin: updated

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -105,6 +111,7 @@
     private Push() {
         mFactories = new HashMap<>();
         mFactories.put(PushInstallationLog.TYPE, new PushInstallationLogFactory());
+        mReceivedIntents = new HashMap<>();
```
guperrot: This can be moved to declaration.

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -367,10 +374,17 @@ private synchronized void checkPushInIntent(Activity activity, Intent intent) {
...
                 }
-                PushNotification notification = new PushNotification(intent);
+
                 AppCenterLog.info(LOG_TAG, "Clicked push message from background id=" + googleMessageId);
                 mLastGoogleMessageId = googleMessageId;
                 AppCenterLog.debug(LOG_TAG, "Push intent extras=" + intent.getExtras());
+
+                /* Try to get the received push intent via google message id. */
+                Intent currentPushIntent = mReceivedIntents.get(googleMessageId);
+                if (currentPushIntent == null) {
```
guperrot: This is precisely what will make the feature unreliable. If the process is restarted before clicking on the notification will be missing title and message. This is thus not a solution that should get recommended as users will need guarantee if they need to read those fields.

guperrot:
This is more complex than expected and requires some design changes, we are postponing this due to other priorities.

Issue #941, opened 2019-02-04
Issue sending logs on Internal wifi network to App center


sonu1:
### **Description**

Having Issues sending logs to App Center on SSID network. 
**Location :**  appcenter\utils\NetworkStateHelper.javaNetworkStateHelper.java
Class NetworkStateHelper  ->   Method reopen() --> public void onAvailable(Network network) method is not called when the SSID network is connected. Is there a work around for this.

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. Connecting to SSID network.
2. appcenter\utils\NetworkStateHelper.javaNetworkStateHelper.java.
Class NetworkStateHelper  ->   Method reopen() --> public void onAvailable(Network network) method is not called when the SSID network is connected. Is there a work.

### **Details**

1. Which SDK version are you using?
    1.11.2
2. Which OS version did you experience the issue on?
    - e.g. Android 6.0.1
3. What device version did you see this error on?  Were you using an emulator or a physical device?
      all devices tested on.
4. What third party libraries are you using?
    NA
5. Please enable verbose logging for your app using `AppCenter.setLogLevel(Log.VERBOSE)` before your call to `AppCenter.start(...)` and include the logs here:

D/AppCenter: Call triggered with no network connectivity, waiting network to become available...
D/AppCenter: Call triggered with no network connectivity, waiting network to become available...
D/AppCenter: checkPendingLogs(group_core) pendingLogCount=0
D/AppCenter: checkPendingLogs(group_analytics) pendingLogCount=0

sonu1:
Hi @MatkovIvan , can you please advice regarding the timeline for this fix. And why is this happening?  Thanks.

MatkovIvan:
Hi @sonu1,

You can review the fix [here](https://github.com/Microsoft/AppCenter-SDK-Android/pull/943) if you are interested.

As I wrote to you in intercom, we release the SDKs every month, but I'm not sure that this fix will be included in the February one.

Thanks,
Ivan

MatkovIvan:
The fix has been merged and will be available in the March release of the SDK.
Keeping issue open until release with the fix.

guperrot:
If you are interested in testing a pre-release of the SDK you can add:

```
repositories {
    maven {
        url  "https://dl.bintray.com/vsappcenter/appcenter-snapshot" 
    }
}
```

[...]

PR #943, opened 2019-02-06
Fix network connection state tracking

MatkovIvan:
* [x] Did you add unit tests?

## Description

Fix network connection state tracking

## Related PRs or issues

#941

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
@@ -42,6 +42,9 @@ public class HttpClientNetworkStateHandler extends HttpClientDecorator implement
         Call call = new Call(mDecoratedApi, url, method, headers, callTemplate, serviceCallback);
         if (mNetworkStateHelper.isNetworkConnected()) {
             call.run();
+
+            /* Send pending call(s), if any. It can be the case if a network state update event is missing. */
+            onNetworkStateUpdated(true);
         } else {
             mCalls.add(call);
             AppCenterLog.debug(LOG_TAG, "Call triggered with no network connectivity, waiting network to become available...");
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
@@ -42,9 +42,6 @@ public class HttpClientNetworkStateHandler extends HttpClientDecorator implement
         Call call = new Call(mDecoratedApi, url, method, headers, callTemplate, serviceCallback);
         if (mNetworkStateHelper.isNetworkConnected()) {
             call.run();
-
-            /* Send pending call(s), if any. It can be the case if a network state update event is missing. */
-            onNetworkStateUpdated(true);
         } else {
             mCalls.add(call);
             AppCenterLog.debug(LOG_TAG, "Call triggered with no network connectivity, waiting network to become available...");
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
@@ -188,7 +188,8 @@ public class NetworkStateHelper implements Closeable {
      * Handle network available update on API level >= 21.
      */
     @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
-    private synchronized void onNetworkAvailable(@SuppressWarnings("unused") Network network) {
+    private synchronized void onNetworkAvailable(Network network) {
+        AppCenterLog.debug(LOG_TAG, "Network "+ network + " is available.");
         if (!mConnected) {
             mConnected = true;
             notifyNetworkStateUpdated(true);
@@ -200,6 +201,7 @@ public class NetworkStateHelper implements Closeable {
      */
     @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
     private synchronized void onNetworkLost(Network network) {
+        AppCenterLog.debug(LOG_TAG, "Network "+ network + " is lost.");
         Network[] networks = mConnectivityManager.getAllNetworks();
         boolean noNetwork = networks == null || networks.length == 0 ||
                 Arrays.equals(networks, new Network[] { network });
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
@@ -188,8 +188,7 @@ public class NetworkStateHelper implements Closeable {
      * Handle network available update on API level >= 21.
      */
     @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
-    private synchronized void onNetworkAvailable(Network network) {
-        AppCenterLog.debug(LOG_TAG, "Network "+ network + " is available.");
+    private synchronized void onNetworkAvailable(@SuppressWarnings("unused") Network network) {
         if (!mConnected) {
             notifyNetworkStateUpdated(true);
             mConnected = true;
@@ -201,7 +200,6 @@ public class NetworkStateHelper implements Closeable {
      */
     @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
     private synchronized void onNetworkLost(Network network) {
-        AppCenterLog.debug(LOG_TAG, "Network "+ network + " is lost.");
         Network[] networks = mConnectivityManager.getAllNetworks();
         boolean noNetworks = networks == null || networks.length == 0 ||
                 Arrays.equals(networks, new Network[] { network });
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
@@ -176,7 +176,8 @@ public class NetworkStateHelper implements Closeable {
     private synchronized void onNetworkLost(Network network) {
         AppCenterLog.debug(LOG_TAG, "Network "+ network + " is lost.");
         Network[] networks = mConnectivityManager.getAllNetworks();
-        boolean noNetworks = networks.length == 0 || Arrays.equals(networks, new Network[] { network });
+        boolean noNetworks = networks == null || networks.length == 0 ||
+                Arrays.equals(networks, new Network[] { network });
         if (mConnected && noNetworks) {
             notifyNetworkStateUpdated(false);
             mConnected = false;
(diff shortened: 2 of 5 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -149,81 +144,54 @@ private IntentFilter getOldIntentFilter() {
...
-    /**
-     * Handle network available update on API level >= 21.
-     */
-    private synchronized void onNetworkAvailable(Network network) {
-        AppCenterLog.debug(LOG_TAG, "Network available netId: " + network);
-        mAvailableNetworks.add(network);
-        AppCenterLog.debug(LOG_TAG, "Available networks netIds: " + mAvailableNetworks);
+        /* The actual current state should be checked here instead of using a value based on connection change events. */
+        try {
+            NetworkInfo info = mConnectivityManager.getActiveNetworkInfo();
```
guperrot: No this is inaccurate and caused issues in the past, we had to move away from this call. Please revert this change.

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
```diff
@@ -42,6 +42,9 @@ public synchronized ServiceCall callAsync(String url, String method, Map<String,
         Call call = new Call(mDecoratedApi, url, method, headers, callTemplate, serviceCallback);
         if (mNetworkStateHelper.isNetworkConnected()) {
             call.run();
+
+            /* Send pending call(s), if any. It can be the case if a network state update event is missing. */
+            onNetworkStateUpdated(true);
```
guperrot: This is not the right solution I think.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
MatkovIvan: Maybe keep it as a fallback? Because rely just on connection change events is inaccurate too

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
MatkovIvan: I agree that we can play with network capabilities, but this might be saved as a fallback.

guperrot: But isNetworkConnected returns false if there is no event with the modern APIs, also the old APIs might work fine on older devices, the customer is on API level 23 apparently so using the new APIs.

guperrot: (also getActiveNetworkInfo is out of sync with events on some Android versions and has been deprecated, specifically on Android 8 where it can say "blocked" while we received an event for active using old APIs)

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
guperrot: This API has already been proven to not work correctly like returning "blocked" state while the network was up. It is also deprecated. It also does not work correctly when there is more than 1 network as mentioned in the past by a customer who worked on a network library.

MatkovIvan: it's not deprecated, but OK, I got it. Thanks for the info

guperrot:
We can consider the fallbacks later if nothing else seems to work (after investigating the capability filters). But in that case we would better poll all network info and not the "active" one.

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
MatkovIvan: it's not deprecated. I tested on Android 9 and yes, it's out of sync with events, but it's not an issue here.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
guperrot: Ah yeah its `getAllNetworkInfo()` that is deprecated indeed.

MatkovIvan:
List of changes:
* Do not call network connected state change event on change type/number of connected networks (there was no action on it anyway);
* Remove `NetworkCapabilities.NET_CAPABILITY_VALIDATED` to get connection events to restricted networks (the main fix);
* Check all network state as fallback to make sure if we are really offline;
* Send queued logs even if connection event is missed;
* Threat lack of network state permission as "always online";
* Remove logging network ids (can be treated as sensitive information);

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientNetworkStateHandler.java
guperrot: I'm still uncomfortable with this line. It's code executed from a specific consumer of the network state helper lib and was not needed to fix the issue. Can we revert that line?

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -123,17 +121,20 @@ public void onLost(Network network) {
...
 
             /*
              * Can be security exception if permission missing or sometimes another runtime exception
              * on some customized firmwares.
              */
-            AppCenterLog.error(LOG_TAG, "Cannot access network state information", e);
+            AppCenterLog.error(LOG_TAG, "Cannot access network state information.", e);
+
+            /* We should try to send the data, even if we can't get the current network state. */
+            mConnected = true;
```
guperrot: ok

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -149,81 +150,73 @@ private IntentFilter getOldIntentFilter() {
      * @return true for connected, false for disconnected.
      */
     public synchronized boolean isNetworkConnected() {
-        return mNetworkType != null || !mAvailableNetworks.isEmpty();
+        return mConnected || isAnyNetworkConnected();
     }
 
     /**
-     * Handle network available update on API level >= 21.
+     * Check if any network is connected.
```
guperrot: Maybe comment it's used only as a fallback.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -149,81 +150,73 @@ private IntentFilter getOldIntentFilter() {
...
+    private synchronized void onNetworkAvailable(@SuppressWarnings("unused") Network network) {
+        if (!mConnected) {
             notifyNetworkStateUpdated(true);
+            mConnected = true;
         }
     }
 
     /**
-     * Update network type by polling on API level < 21.
+     * Handle network available update on API level >= 21.
```
guperrot: network lost update, not available

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -149,81 +150,73 @@ private IntentFilter getOldIntentFilter() {
...
-
-        /* Update network type. null for not connected. */
-        if (networkInfo != null && networkInfo.isConnected()) {
-            mNetworkType = networkInfo.getTypeName() + networkInfo.getSubtypeName();
-        } else {
-            mNetworkType = null;
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    private synchronized void onNetworkLost(Network network) {
+        Network[] networks = mConnectivityManager.getAllNetworks();
+        boolean noNetworks = networks == null || networks.length == 0 ||
```
guperrot: remove `s` from `noNetworks`

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -149,81 +150,73 @@ private IntentFilter getOldIntentFilter() {
...
+                Arrays.equals(networks, new Network[] { network });
+        if (mConnected && noNetworks) {
+            notifyNetworkStateUpdated(false);
+            mConnected = false;
```
guperrot: update state before notifying

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -149,81 +150,73 @@ private IntentFilter getOldIntentFilter() {
...
-        boolean networkTypeChanged = previousNetworkType == null ? mNetworkType != null : !previousNetworkType.equals(mNetworkType);
-        if (networkTypeChanged) {
-            boolean connected = isNetworkConnected();
-            if (connected && previousNetworkType != null) {
-                notifyNetworkStateUpdated(false);
-            }
+        boolean connected = isAnyNetworkConnected();
+        if (connected != mConnected) {
             notifyNetworkStateUpdated(connected);
+            mConnected = connected;
```
guperrot: update state before notifying

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
MatkovIvan: https://github.com/Microsoft/AppCenter-SDK-Android/pull/943/files#diff-f8b58324d3d38f4997ba0df305858c37R216

guperrot:
What API levels were tested on real device?

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -233,19 +226,19 @@ private synchronized void handleNetworkStateUpdate() {
      * @param connected whether the network is connected or not.
      */
     private void notifyNetworkStateUpdated(boolean connected) {
+        AppCenterLog.debug(LOG_TAG, "Network has been " + (connected ? "connected." : "disconnected."));
```
MatkovIvan: @guperrot It's logged here. Anyway, I'll revert the logs in the callbacks

guperrot: I couldn't see any debug logs when I was testing, will retry.

guperrot:
Looks good, need more testing now.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
guperrot: I see this log now, maybe I missed it last time.

PR #950, opened 2019-02-13
Send token in Authorization header to ingestion

annakocheshkova:
* [X] Did you add unit tests?
* [X] Did you check UI tests on the sample app? They are not executed on CI.

## Description

Implements `AuthTokenContext` to manage identity token. Token context uses `TokenStorage`, which is implemented using Preferences right now.
`AuthTokenContext` is injected into the `Channel`, which passes the token to `Ingestion` and `Ingestion` puts the `AUTHORIZATION` header.
_NOTE_: **Do not merge** this branch before you merge tests into it. They're in a separate PR (#951).

## Related PRs or issues

#951


sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -377,7 +378,9 @@ private synchronized void loginFromUI() {
                 @Override
                 public void onSuccess(IAuthenticationResult authenticationResult) {
                     AppCenterLog.info(LOG_TAG, "User login succeeded. id=" + authenticationResult.getIdToken());
-                    // TODO send id token (not access token) to ingestion, if sdk still enabled.
+                    if (isInstanceEnabled() && AppCenter.isEnabled().get()) {
```
MatkovIvan:
```suggestion
                    if (Identity.isEnabled().get()) {
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -894,10 +907,10 @@ private void finishStartServices(Iterable<AppCenterService> updatedServices, Ite
                 service.setInstanceEnabled(false);
             }
             if (startFromApp) {
-                service.onStarted(mApplication, mChannel, mAppSecret, mTransmissionTargetToken, true);
+                service.onStarted(mApplication, mChannel, mAuthTokenContext, mAppSecret, mTransmissionTargetToken, true);
                 AppCenterLog.info(LOG_TAG, service.getClass().getSimpleName() + " service started from application.");
             } else {
-                service.onStarted(mApplication, mChannel, null, null, false);
+                service.onStarted(mApplication, mChannel, mAuthTokenContext,null, null, false);
```
MatkovIvan:
```suggestion
                service.onStarted(mApplication, mChannel, mAuthTokenContext, null, null, false);
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -122,8 +128,8 @@
      * @param logSerializer    The log serializer.
      * @param appCenterHandler App Center looper thread handler.
      */
-    public DefaultChannel(@NonNull Context context, String appSecret, @NonNull LogSerializer logSerializer, @NonNull Handler appCenterHandler) {
-        this(context, appSecret, buildDefaultPersistence(context, logSerializer), new AppCenterIngestion(context, logSerializer), appCenterHandler);
+    public DefaultChannel(@NonNull Context context, String appSecret, @NonNull LogSerializer logSerializer, @NonNull Handler appCenterHandler, AuthTokenContext authTokenContext) {
```
MatkovIvan:
```suggestion
    public DefaultChannel(@NonNull Context context, String appSecret, @NonNull LogSerializer logSerializer, @NonNull Handler appCenterHandler, @NonNull AuthTokenContext authTokenContext) {
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -0,0 +1,80 @@
...
+    private final Collection<Listener> mListeners;
+
+    /**
+     * Storage that handles saving and encrypting token.
+     */
+    private TokenStorage mTokenStorage;
+
+    /**
+     * Initializes token context with the given token storage.
+     * @param tokenStorage token storage to save token.
```
MatkovIvan: Empty line before

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/PreferenceTokenStorage.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Storage for tokens that uses Shared Preferences. Handles saving and encryption.
+ */
+public class PreferenceTokenStorage implements TokenStorage {
+
+    /**
+     * Context instance.
+     */
+    private Context mContext;
```
MatkovIvan: final?

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/AppCenterIngestion.java
```diff
@@ -55,6 +51,20 @@
     @VisibleForTesting
     static final String APP_SECRET = "App-Secret";
 
+    /**
+     * Auth token format for Authorization header.
+     */
+    @VisibleForTesting
+    @SuppressWarnings("WeakerAccess")
+    static final String AUTH_TOKEN = "Bearer %s";
```
jwallra: This should be `AUTH_TOKEN_FORMAT`

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/AppCenterIngestion.java
```diff
@@ -55,6 +51,20 @@
...
+
+    /**
+     * Authorization HTTP Header.
+     */
+    @VisibleForTesting
+    @SuppressWarnings("WeakerAccess")
+    static final String AUTHORIZATION = "Authorization";
```
jwallra: `AUTHORIZATION_HEADER`

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/AppCenterIngestion.java
jwallra: This and `AUTHORIZATION_HEADER` should be in a shared file since they're duplicated [here](https://github.com/Microsoft/AppCenter-SDK-Android/pull/950/files#diff-147e2a59ce7018d183c480e87a829c84R89)

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/PreferenceTokenStorage.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param context context instance.
+     */
+    public PreferenceTokenStorage(Context context) {
+        mContext = context;
+    }
+
+    /**
+     * Base key for stored preferences.
+     */
+    static final String PREFERENCE_PREFIX = "AppCenter.";
```
jwallra: This doesn't look like it's used anywhere else.  I would prefer that this just be combined with `PREFERENCE_KEY_AUTH_TOKEN`

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -377,7 +378,9 @@ private synchronized void loginFromUI() {
                 @Override
                 public void onSuccess(IAuthenticationResult authenticationResult) {
                     AppCenterLog.info(LOG_TAG, "User login succeeded. id=" + authenticationResult.getIdToken());
-                    // TODO send id token (not access token) to ingestion, if sdk still enabled.
+                    if (Identity.isEnabled().get()) {
```
guperrot: What thread is that since we are calling `.get`? Prefer using `thenAccept`.
Also do we need to fully qualify the static call since we are already in the same class?

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/AppCenterIngestion.java
```diff
@@ -55,6 +51,20 @@
     @VisibleForTesting
     static final String APP_SECRET = "App-Secret";
 
+    /**
+     * Auth token format for Authorization header.
+     */
+    @VisibleForTesting
+    @SuppressWarnings("WeakerAccess")
```
guperrot: Add TODO to remove those warnings, it would be fixed by adding the tests

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AbstractTokenContextListener.java
```diff
@@ -0,0 +1,11 @@
+package com.microsoft.appcenter.utils.context;
+
+/**
+ * Empty implementation to make callbacks optional.
+ */
+public class AbstractTokenContextListener implements AuthTokenContext.Listener {
```
guperrot: We use this pattern when there is more than 1 callback and that some of them are optional.
I don't see the use case for this one.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -410,7 +411,7 @@ protected int getTriggerCount() {
     private static final String DISTRIBUTE_GROUP = "group_distribute";
 
     @Override
-    public synchronized void onStarted(@NonNull Context context, @NonNull Channel channel, String appSecret, String transmissionTargetToken, boolean startedFromApp) {
+    public synchronized void onStarted(@NonNull Context context, @NonNull Channel channel, @NonNull AuthTokenContext authTokenContext, String appSecret, String transmissionTargetToken, boolean startedFromApp) {
```
guperrot: Since only 1 module uses the context, I would prefer we use the pattern of `get(Context)` singleton. NetworkStateHelper is an example. That would avoid modifications to all services and abstractions.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/PreferenceTokenStorage.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public void saveToken(String token) {
+        String encryptedToken = CryptoUtils.getInstance(mContext).encrypt(token);
+        SharedPreferencesManager.putString(PREFERENCE_KEY_AUTH_TOKEN, encryptedToken);
+    }
+
+    @Override
+    public String getToken() {
+        String encryptedToken = SharedPreferencesManager.getString(PREFERENCE_KEY_AUTH_TOKEN, "");
```
guperrot: why not using null as default value? and return null.
This is rather unusual.

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/OneCollectorIngestion.java
```diff
@@ -131,6 +141,9 @@ public ServiceCall sendAsync(String appSecret, UUID installId, LogContainer logC
             apiKey.deleteCharAt(apiKey.length() - 1);
         }
         headers.put(API_KEY, apiKey.toString());
+        if (authToken != null && authToken.length() > 0) {
```
dhei: We don't need `authToken` for One Collector requests, right? I don't think it should be added here.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/PreferenceTokenStorage.java
dhei: @jwallra `PREFERENCE_PREFIX` is used below in `PREFERENCE_KEY_AUTH_TOKEN`

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/OneCollectorIngestion.java
annakocheshkova: :+1:

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/Ingestion.java
```diff
@@ -35,4 +37,17 @@
      * Make ingestion active again after closing.
      */
     void reopen();
+    /**
```
dhei: nit: missing a line break.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -0,0 +1,111 @@
...
+     */
+    public interface Listener {
+
+        /**
+         * Called whenever a new token is set.
+         */
+        void onNewAuthToken(String authToken);
+    }
+}
+
```
dhei: nit: remove extra line break.

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/AppCenterIngestion.java
```diff
@@ -93,11 +106,24 @@ public void setLogUrl(@NonNull String logUrl) {
...
+    public String getAuthToken() {
+        return mAuthToken;
+    }
+
     @Override
     public ServiceCall sendAsync(String appSecret, UUID installId, LogContainer logContainer, final ServiceCallback serviceCallback) throws IllegalArgumentException {
         Map<String, String> headers = new HashMap<>();
         headers.put(INSTALL_ID, installId.toString());
         headers.put(APP_SECRET, appSecret);
+        if (mAuthToken != null && mAuthToken.length() > 0) {
```
achocron: This block (or just the method) along with the getter and setters for auth token, should be synchronized because otherwise the auth token can become invalid inside this block from another thread

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/ITokenStorage.java
```diff
@@ -0,0 +1,26 @@
+package com.microsoft.appcenter.utils.context;
```
achocron: Can you move the token storage classes to the Identity module? It doesn't need to exist in core

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -0,0 +1,111 @@
...
+        return sInstance;
+    }
+
+    /**
+     * A private constructor for the class.
+     *
+     * @param context application context instance.
+     */
+    private AuthTokenContext(Context context) {
+        mTokenStorage = TokenStorageFactory.getTokenStorage(context);
```
achocron: Follow up to my comment below about moving this to Identity module: if we move the token storage from here into Identity, then identity can be the one to save the token in storage when there is a login. And if we want to be really really careful (which might not be necessary), Identity could even be a listener and update the storage when the auth token changes

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/ITokenStorage.java
achocron: We try to avoid coupling the modules as much as possible, and this puts more Identity logic in the core module than necessary

guperrot: It needs to exist in core because we share it with data sync.

achocron: Doesn't data sync require identity module as a dependency? And in that case, why would data sync need to access storage, it can just use context

achocron: Identity would retrieve token from storage and populate context

guperrot: Ah so you want to move only what can be moved and keep enough code in core just for sharing the token?

achocron: Yep. I agree that context must be in core. But the token persistence/storage can be moved

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -375,9 +376,16 @@ private synchronized void loginFromUI() {
...
 
                 @Override
-                public void onSuccess(IAuthenticationResult authenticationResult) {
+                public void onSuccess(final IAuthenticationResult authenticationResult) {
                     AppCenterLog.info(LOG_TAG, "User login succeeded. id=" + authenticationResult.getIdToken());
-                    // TODO send id token (not access token) to ingestion, if sdk still enabled.
+                    getInstance().post(new Runnable() {
+                        @Override
+                        public void run() {
+                            if (isEnabled().get()) {
```
guperrot: post ensures its enabled already, please remove that check
everything is executed in a sequence so state cannot change (its the background thread looper).

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -741,6 +752,13 @@ public synchronized void shutdown() {
         suspend(false, new CancellationException());
     }
 
+    @Override
+    public void onNewAuthToken(String authToken) {
```
guperrot: double check threading here and possibly add synchronized if needed

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/AppCenterIngestion.java
guperrot: If you add synchronized here, must also add to get and set.

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/OneCollectorIngestion.java
```diff
@@ -197,6 +193,11 @@ public void close() throws IOException {
         mHttpClient.close();
     }
 
+    public void setAuthToken(@NonNull String authToken) { }
```
guperrot: is `@Override` missing here?

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -134,9 +141,10 @@ public DefaultChannel(@NonNull Context context, String appSecret, @NonNull LogSe
      * @param persistence      Persistence object for dependency injection.
      * @param ingestion        Ingestion object for dependency injection.
      * @param appCenterHandler App Center looper thread handler.
+     * @param authTokenContext Authorization token context.
      */
     @VisibleForTesting
-    DefaultChannel(@NonNull Context context, String appSecret, @NonNull Persistence persistence, @NonNull Ingestion ingestion, @NonNull Handler appCenterHandler) {
+    DefaultChannel(@NonNull Context context, String appSecret, @NonNull Persistence persistence, @NonNull Ingestion ingestion, @NonNull Handler appCenterHandler, AuthTokenContext authTokenContext) {
```
guperrot: no need to be a dependency, context is already a dependency so you can init from context in constructor

our mocking framework is powerful enough in java to avoid boilerplate if that was the reason

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/ITokenStorage.java
guperrot: Sounds good.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
achocron: I don't think it needs to be synchronized unless `mIngestion` can disappear on another thread somehow

achocron: Which may be possible, I'm not sure

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/ITokenStorage.java
```diff
@@ -0,0 +1,26 @@
...
+     * Retrieves token value.
+     *
+     * @return auth token.
+     */
+    String getToken();
+
+    /**
+     * Removes token value.
+     */
+    void removeToken();
```
achocron: Why do we need a `removeToken` method? I don't see it used anywhere

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/PreferenceTokenStorage.java
```diff
@@ -0,0 +1,53 @@
...
+
+import android.content.Context;
+
+import com.microsoft.appcenter.utils.crypto.CryptoUtils;
+import com.microsoft.appcenter.utils.storage.SharedPreferencesManager;
+
+/**
+ * Storage for tokens that uses {@link SharedPreferencesManager}. Handles saving and encryption.
+ */
+public class PreferenceTokenStorage implements ITokenStorage {
```
achocron: Can we call this AuthTokenStorage? I think it's a bit more clear and this is what we decided to call it in the Apple SDK

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/ITokenStorage.java
achocron: Update: I think we need it, but please call it when we disable the Identity service

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/ITokenStorage.java
annakocheshkova: Discussed this with @achocron, all clear now

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/AppCenterIngestion.java
```diff
@@ -121,8 +121,10 @@ public ServiceCall sendAsync(String appSecret, UUID installId, LogContainer logC
         Map<String, String> headers = new HashMap<>();
         headers.put(INSTALL_ID, installId.toString());
         headers.put(APP_SECRET, appSecret);
-        if (mAuthToken != null && mAuthToken.length() > 0) {
-            headers.put(AUTHORIZATION_HEADER, String.format(AUTH_TOKEN_FORMAT, mAuthToken));
+        synchronized (this) {
```
guperrot: move it to a private method, synchronized block causes code coverage issue when used as a block within a method

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -88,15 +88,24 @@ public synchronized String getAuthToken() {
      *
      * @param authToken authorization token.
      */
-    public synchronized void setAuthToken(String authToken) {
-        mTokenStorage.saveToken(authToken);
+    public void setAuthToken(String authToken) {
+        synchronized (this) {
```
guperrot: same comment on synchronized

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/AppCenterIngestion.java
guperrot: that is if we still have the issue with yellow synchronized covereage on local html report, please double check

annakocheshkova: I don't see yellow coverage. All green.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -177,6 +178,7 @@ protected synchronized void applyEnabledState(boolean enabled) {
             mIdentityScope = null;
             mLoginDelayed = false;
             clearCache();
+           // AuthTokenContext.getInstance().clearData();
```
guperrot: Can we cleanup this?

sdk/appcenter-identity/src/test/java/com/microsoft/appcenter/identity/IdentityTest.java
```diff
@@ -114,6 +131,7 @@ public void setEnabled() {
         /* Disable. Testing to wait setEnabled to finish while we are at it. */
         Identity.setEnabled(false).get();
         assertFalse(Identity.isEnabled().get());
+        //verify(tokenContext).clearData();
```
guperrot: Can we cleanup this?

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -114,6 +115,11 @@
      */
     private int mCurrentState;
 
+    /**
+     * The Auth context object to retrieve auth token value.
+     */
+    private AuthTokenContext mAuthTokenContext;
```
guperrot: Keeping a reference is not needed, you access it only in the constructor.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /**
+     * Sets new authorization token.
+     *
+     * @param authToken authorization token.
+     */
+    public void setAuthToken(String authToken) {
+        mAuthToken = authToken;
+        //mLastHomeAccountId = homeAccountId;
```
guperrot: Can we cleanup this?

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -0,0 +1,120 @@
...
+        mListeners.add(listener);
+    }
+
+    /**
+     * Removes a specific listener.
+     *
+     * @param listener listener to be removed.
+     */
+    public synchronized void removeListener(Listener listener) {
+        mListeners.remove(listener);
```
guperrot: Not covered in tests

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -0,0 +1,120 @@
...
+        for (Listener listener : mListeners) {
+            listener.onNewAuthToken(authToken);
+        }
+    }
+
+    /**
+     * Clears info about the token.
+     */
+    public void clearToken() {
+        mAuthToken = null;
```
guperrot: Not covered in tests

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/AppCenterIngestion.java
guperrot: screen shot 2019-02-19 at 2 24 38 pm

The ending curly brace is yellow (not talking about the if). Might be an issue after you fixed the if. If you have an issue with that one you will need a private method.

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/AppCenterIngestion.java
```diff
@@ -93,11 +106,26 @@ public void setLogUrl(@NonNull String logUrl) {
...
+        return mAuthToken;
+    }
+
     @Override
     public ServiceCall sendAsync(String appSecret, UUID installId, LogContainer logContainer, final ServiceCallback serviceCallback) throws IllegalArgumentException {
         Map<String, String> headers = new HashMap<>();
         headers.put(INSTALL_ID, installId.toString());
         headers.put(APP_SECRET, appSecret);
+        synchronized (this) {
+            if (mAuthToken != null && mAuthToken.length() > 0) {
```
guperrot: Missing some test coverage for this if

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/AppCenterIngestion.java
annakocheshkova: Hmm on my mac it was an, i will look at it

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
annakocheshkova: @guperrot This will be covered once we merge Push PR into this branch. It uses `removeListener` and has tests to check if this listener is actually removed. But if not, I will add coverage after we merge all "child" PRs.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Sets new authorization token.
+     *
+     * @param authToken authorization token.
+     */
+    public void setAuthToken(String authToken) {
+        mAuthToken = authToken;
+
+        /* Call listeners so that they can react on new token. */
+        for (Listener listener : mListeners) {
```
guperrot: This will crash on multi threaded env

sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/OneCollectorIngestion.java
```diff
@@ -197,6 +193,12 @@ public void close() throws IOException {
         mHttpClient.close();
     }
 
+    @Override
+    public void setAuthToken(@NonNull String authToken) { }
```
guperrot: Formatting issue

PR #954, opened 2019-02-15
Add note to the java doc

MatkovIvan:
* [ ] Did you add unit tests?

## Description

Javadoc improvement

## Related PRs or issues

https://github.com/MicrosoftDocs/appcenter-docs/issues/463/

### DIFF
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
@@ -168,6 +168,8 @@ public class Analytics extends AbstractAppCenterService {
 
     /**
      * Enable or disable Analytics service.
+     * <p>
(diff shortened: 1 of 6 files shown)


MatkovIvan:
> device 'emulator(AVD)' for 'appcenter:': Unknown API Level

:flushed:

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -168,6 +168,8 @@ static synchronized void unsetInstance() {
 
     /**
      * Enable or disable Analytics service.
+     * <p>
+     * The state is stored on disk, so it won't change in following app launches.
```
guperrot: Disk is probably not the best term for phone storage.

The state is persisted if the application process restarts or something like that should do the trick without to mention disk.

PR #956, opened 2019-02-19
Identity persist token

annakocheshkova:
* [X] Did you add unit tests?
* [X] Did you check UI tests on the sample app? They are not executed on CI.

## Description

Need to persist the encrypted token. 

## Related PRs or issues

#950

### DIFF
--- a/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
+++ b/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
@@ -16,6 +16,8 @@ import com.microsoft.appcenter.http.HttpException;
 import com.microsoft.appcenter.http.HttpUtils;
 import com.microsoft.appcenter.http.ServiceCall;
 import com.microsoft.appcenter.http.ServiceCallback;
+import com.microsoft.appcenter.identity.storage.AuthTokenStorage;
+import com.microsoft.appcenter.identity.storage.TokenStorageFactory;
 import com.microsoft.appcenter.utils.AppCenterLog;
 import com.microsoft.appcenter.utils.HandlerUtils;
 import com.microsoft.appcenter.utils.async.AppCenterFuture;
@@ -102,6 +104,11 @@ public class Identity extends AbstractAppCenterService {
     private boolean mLoginDelayed;
 
     /**
+     * Instance of {@link AuthTokenStorage} to store token information.
+     */
+    private AuthTokenStorage mTokenStorage;
+
+    /**
      * Get shared instance.
      *
      * @return shared instance.
@@ -152,6 +159,7 @@ public class Identity extends AbstractAppCenterService {
     public synchronized void onStarted(@NonNull Context context, @NonNull Channel channel, String appSecret, String transmissionTargetToken, boolean startedFromApp) {
         mContext = context;
         mAppSecret = appSecret;
+        mTokenStorage = TokenStorageFactory.getTokenStorage(context);
         super.onStarted(context, channel, appSecret, transmissionTargetToken, startedFromApp);
     }
 
--- a/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
+++ b/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
@@ -175,9 +175,11 @@ public class Identity extends AbstractAppCenterService {
             /* Load cached configuration in case APIs are called early. */
             loadConfigurationFromCache();
 
+            /* Load the last stored token and cache it into token context. */
+            mTokenStorage.cacheToken();
+
             /* Download the latest configuration in background. */
             downloadConfiguration();
-            mTokenStorage.cacheToken();
         } else {
             if (mGetConfigCall != null) {
                 mGetConfigCall.cancel();
--- a/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
+++ b/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
@@ -21,10 +21,10 @@ import com.microsoft.appcenter.identity.storage.TokenStorageFactory;
 import com.microsoft.appcenter.utils.AppCenterLog;
 import com.microsoft.appcenter.utils.HandlerUtils;
 import com.microsoft.appcenter.utils.async.AppCenterFuture;
-import com.microsoft.appcenter.utils.context.AuthTokenContext;
 import com.microsoft.appcenter.utils.storage.FileManager;
 import com.microsoft.appcenter.utils.storage.SharedPreferencesManager;
 import com.microsoft.identity.client.AuthenticationCallback;
+import com.microsoft.identity.client.IAccount;
 import com.microsoft.identity.client.IAuthenticationResult;
 import com.microsoft.identity.client.PublicClientApplication;
 import com.microsoft.identity.client.exception.MsalException;
@@ -177,6 +177,7 @@ public class Identity extends AbstractAppCenterService {
 
             /* Download the latest configuration in background. */
             downloadConfiguration();
+            mTokenStorage.cacheToken();
         } else {
             if (mGetConfigCall != null) {
                 mGetConfigCall.cancel();
@@ -186,7 +187,7 @@ public class Identity extends AbstractAppCenterService {
             mIdentityScope = null;
             mLoginDelayed = false;
             clearCache();
-           // AuthTokenContext.getInstance().clearData();
+            mTokenStorage.removeToken();
         }
     }
 
@@ -390,7 +391,8 @@ public class Identity extends AbstractAppCenterService {
                     getInstance().post(new Runnable() {
                         @Override
                         public void run() {
-                            AuthTokenContext.getInstance().setAuthToken(authenticationResult.getIdToken());
+                            IAccount account = authenticationResult.getAccount();
+                            mTokenStorage.saveToken(authenticationResult.getIdToken(), account.getHomeAccountIdentifier().getIdentifier());
                         }
                     });
                 }
--- a/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
+++ b/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
@@ -389,7 +389,7 @@ public class Identity extends AbstractAppCenterService {
 
                 @Override
                 public void onSuccess(final IAuthenticationResult authenticationResult) {
-                    AppCenterLog.info(LOG_TAG, "User login succeeded. id=" + authenticationResult.getIdToken());
+                    AppCenterLog.info(LOG_TAG, "User login succeeded.");
                     getInstance().post(new Runnable() {
 
                         @Override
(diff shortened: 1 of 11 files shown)


sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
```diff
@@ -0,0 +1,83 @@
...
+     * Used for authentication requests, string field for auth token.
+     */
+    @VisibleForTesting
+    static final String PREFERENCE_KEY_AUTH_TOKEN = "AppCenter.auth_token";
+
+    /**
+     * Used for distinguishing users, string field for home account id.
+     */
+    @VisibleForTesting
+    static final String PREFERENCE_KEY_HOME_ACCOUNT_ID = "AppCenter.account_id";
```
achocron: Was it in the spec to store multiple tokens for different users? I was thinking we would just store one token at a time

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
```diff
@@ -0,0 +1,83 @@
...
+import android.support.annotation.VisibleForTesting;
+
+import com.microsoft.appcenter.utils.context.AuthTokenContext;
+import com.microsoft.appcenter.utils.crypto.CryptoUtils;
+import com.microsoft.appcenter.utils.storage.SharedPreferencesManager;
+
+/**
+ * Storage for tokens that uses {@link SharedPreferencesManager}. Handles saving and encryption.
+ */
+public class PreferenceTokenStorage implements AuthTokenStorage {
```
jwallra: If this class is also storing account ID and not just auth token I think that we could do some renaming.  Perhaps `PreferenceAccountStorage` or `PreferenceAuthStorage`?  I'm not in love with either of those, does anyone else have thoughts/ideas?  If we change it, the methods (`getToken`, `saveToken`, etc.) should be changed to align.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
annakocheshkova: I didn't implement multiple tokens. Just storing the last account id to distinguish new users

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
jwallra: Or maybe this is an OK convention... I don't know.  Does anyone have another opinion?

annakocheshkova: @jwallra i think storing last account id doesn't change the main "token storage" purpose. I mean this class still focuses mainly on token and uses account id just as a piece of a helper info to tell when the user has changed. It also does not provide any publuc methods for retrieving/removing account id etc.

guperrot: userId is a claim within the token.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
```diff
@@ -0,0 +1,83 @@
...
+    private String getHomeAccountId() {
+        return SharedPreferencesManager.getString(PREFERENCE_KEY_HOME_ACCOUNT_ID, null);
+    }
+
+    @Override
+    public void cacheToken() {
+        String tokenFromStorage = getToken();
+        String accountId = getHomeAccountId();
+
+        // We need to update Token context here.
```
guperrot: Code convention is to use `/*` comments only for Java.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -169,6 +177,7 @@ protected synchronized void applyEnabledState(boolean enabled) {
 
             /* Download the latest configuration in background. */
             downloadConfiguration();
+            mTokenStorage.cacheToken();
```
guperrot: This is not downloading the latest config, need a new comment before. Also should be done before downloading.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/TokenStorageFactory.java
```diff
@@ -0,0 +1,19 @@
...
+public class TokenStorageFactory {
+
+    /**
+     * Retrieves current implementation of {@link AuthTokenStorage}.
+     *
+     * @param context application context.
+     * @return instance of {@link AuthTokenStorage}.
+     */
+    public static AuthTokenStorage getTokenStorage(Context context) {
+        return new PreferenceTokenStorage(context);
```
guperrot: make that a singleton

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    private final Context mContext;
+
+    /**
+     * Default constructor.
+     *
+     * @param context {@link Context} instance.
+     */
+    PreferenceTokenStorage(Context context) {
+        mContext = context;
```
guperrot: mContext = context.getApplicationContext()

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
```diff
@@ -0,0 +1,83 @@
...
+     * {@link Context} instance.
+     */
+    private final Context mContext;
+
+    /**
+     * Default constructor.
+     *
+     * @param context {@link Context} instance.
+     */
+    PreferenceTokenStorage(Context context) {
```
guperrot: @NonNull

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
achocron: ah it's account id. got it.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
annakocheshkova: @guperrot this is already done in CryptoUtils so it will crash if do this twice

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/AuthTokenStorage.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public interface AuthTokenStorage {
+
+    /**
+     * Stores token value.
+     *
+     * @param token auth token.
+     * @param homeAccountId unique identifier of user.
+     */
+    void saveToken(String token, String homeAccountId);
```
dhei: Similarly to @jwallra's [comment](https://github.com/Microsoft/AppCenter-SDK-Android/pull/956/files#r258206045), I think the `saveToken` naming is confusing because it takes 2 parameters: token value and account id value.

dhei: Same confusion for `cacheToken` naming. It caches token value and account id value.

dhei: Ignore my first 2 comments. I realized account id is actually a claim from the auth token, which means `homeAccountId` is a part of the `token`, therefore `AuthTokenStorage` is still the right naming.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/AuthTokenStorage.java
```diff
@@ -0,0 +1,32 @@
+package com.microsoft.appcenter.identity.storage;
+
+/**
+ * Interface for storage that works with token.
+ */
+public interface AuthTokenStorage {
+
+    /**
+     * Stores token value.
```
dhei: Please update the comment to include home account id.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/AuthTokenStorage.java
```diff
@@ -0,0 +1,32 @@
...
+     */
+    String getToken();
+
+    /**
+     * Removes token value.
+     */
+    void removeToken();
+
+    /**
+     * Gets token from storage and caches it.
```
dhei: Please update the comment to include home account id.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
guperrot: oh, weird I believe it should just return self if already app context

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/TokenStorageFactory.java
```diff
@@ -7,13 +7,21 @@
...
+
     /**
      * Retrieves current implementation of {@link AuthTokenStorage}.
      *
      * @param context application context.
      * @return instance of {@link AuthTokenStorage}.
      */
-    public static AuthTokenStorage getTokenStorage(Context context) {
-        return new PreferenceTokenStorage(context);
+    public AuthTokenStorage getTokenStorage(Context context) {
```
guperrot: why removing static?

guperrot: nvm I was looking at changes and seems it was reverted back after.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/TokenStorageFactory.java
```diff
@@ -0,0 +1,27 @@
...
+     */
+    private static AuthTokenStorage sTokenStorageInstance;
+
+    /**
+     * Retrieves current implementation of {@link AuthTokenStorage}.
+     *
+     * @param context application context.
+     * @return instance of {@link AuthTokenStorage}.
+     */
+    public static AuthTokenStorage getTokenStorage(Context context) {
```
guperrot: missing synchronized

guperrot: should be fine if always accessing from worker thread since its storage actually, just adding annotation then

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
guperrot: it works in sasquatch, I just tested

PR #963, opened 2019-02-26
Added signOut method

dtrabo:

* [ ] Did you add unit tests?

## Description

Added signOut method 

## Related PRs or issues

WorkItem: https://msmobilecenter.visualstudio.com/Mobile-Center/_boards/board/t/Doug-Team/Backlog%20Items

### DIFF
--- a/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
+++ b/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
@@ -396,7 +396,7 @@ public class Identity extends AbstractAppCenterService {
 
     private synchronized void instanceSignOut() {
         if (mTokenStorage.getToken() == null) {
-            AppCenterLog.info(LOG_TAG, "Couldn't sign-out: authToken doesn't exist.");
+            AppCenterLog.warn( LOG_TAG, "Couldn't sign-out: authToken doesn't exist.");
             return;
         }
         mSignInDelayed = false;
@@ -413,8 +413,8 @@ public class Identity extends AbstractAppCenterService {
 
             @Override
             public void onAccountsLoaded(List<IAccount> accounts) {
-                for (IAccount account : accounts){
-                    if (account.getHomeAccountIdentifier().getIdentifier().equals(homeAccountIdentifier)){
+                for (IAccount account : accounts) {
+                    if (account.getHomeAccountIdentifier().getIdentifier().equals(homeAccountIdentifier)) {
                         mAuthenticationClient.removeAccount(account);
                     }
                 }
--- a/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
+++ b/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
@@ -363,8 +363,15 @@ public class Identity extends AbstractAppCenterService {
         }
     }
 
-    public synchronized static void signOut() {
-        getInstance().mTokenStorage.removeToken();
+    /**
+     * Sign out user and invalidate a user's token.
+     */
+    public static void signOut() {
+        getInstance().instanceSignOut();
+    }
+
+    private void instanceSignOut() {
+
     }
 
     @WorkerThread
--- a/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
+++ b/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
@@ -37,6 +37,7 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URL;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import static android.util.Log.VERBOSE;
@@ -159,7 +160,13 @@ public class Identity extends AbstractAppCenterService {
      * Sign out user and invalidate a user's token.
      */
     public static void signOut() {
-        getInstance().instanceSignOut();
+        getInstance().post(new Runnable() {
+
+            @Override
+            public void run() {
+                getInstance().instanceSignOut();
+            }
+        });
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -125,6 +125,7 @@ public void clearToken() {
         for (Listener listener : mListeners) {
             listener.onNewAuthToken(null);
             listener.onNewUser(null);
+            listener.onSignOut();
```
MatkovIvan: It will be raised on disable Identity module

MatkovIvan: Why do you need the separate method then? Maybe we can just re-use the previous one with `null` in parameters?

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -363,6 +363,10 @@ private void saveConfigFile(String payload, String eTag) {
         }
     }
 
+    public synchronized static void signOut() {
```
guperrot: remove synchronized on the static version, add Javadoc

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -363,6 +363,10 @@ private void saveConfigFile(String payload, String eTag) {
...
+        getInstance().mTokenStorage.removeToken();
```
guperrot: move this code inside a private method to avoid getInstance().mField

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
guperrot: Disables already uses null indeed.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -380,6 +387,10 @@ public void run() {
         });
     }
 
+    private void instanceSignOut() {
+        mTokenStorage.removeToken();
```
MatkovIvan: @dtrabo Please also reset the `mLoginDelayed` flag

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -380,6 +387,16 @@ public void run() {
         });
     }
 
+    private void instanceSignOut() {
+        post(new Runnable() {
+
+            @Override
+            public void run() {
+                mTokenStorage.removeToken();
```
MatkovIvan: Also, we need to synchronize it to avoid thread issues

guperrot: Actually not needed here as it's always accessed from the background looper.

We usually don't add synchronized to things accessed solely on background thread, especially after https://github.com/Microsoft/AppCenter-SDK-Android/pull/665 happened.

The code is not harmonized.

sdk/appcenter-identity/src/test/java/com/microsoft/appcenter/identity/IdentityTest.java
```diff
@@ -461,6 +461,16 @@ public void readCacheAndRefreshNotModified() throws Exception {
         verifyNew(PublicClientApplication.class, times(1));
     }
 
+    @Test
+    public void testSignOut() {
```
dhei: You don't need to prefix `test` to Java tests as opposed to Objective-C tests, though we're not consistent. What about name it `signOutRemovesToken` for clarity?

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
MatkovIvan: @guperrot `mSignInDelayed` is accessed from UI thread too. So, we need to synchronize it.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -380,6 +394,34 @@ public void run() {
         });
     }
 
+    private synchronized void instanceSignOut() {
+        if (mTokenStorage.getToken() == null) {
+            AppCenterLog.warn( LOG_TAG, "Couldn't sign-out: authToken doesn't exist.");
```
MatkovIvan:
```suggestion
            AppCenterLog.warn(LOG_TAG, "Couldn't sign-out: authToken doesn't exist.");
```

jwallra: also should be `sign out` since it's a verb - `sign-out` is when it's used as a noun

sdk/appcenter-identity/src/test/java/com/microsoft/appcenter/identity/IdentityTest.java
```diff
@@ -461,6 +462,27 @@ public void readCacheAndRefreshNotModified() throws Exception {
...
+        /* Sign out should clear token. */
+        Identity.signOut();
+        verify(mPreferenceTokenStorage).removeToken();
+    }
+
+    @Test
+    public void verifyTokenStorageIsEmpty() {
+        Identity identity = Identity.getInstance();
+        start(identity);
+
```
jwallra: remove extra blank line

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -380,6 +394,34 @@ public void run() {
...
+        if (mAuthenticationClient == null || homeAccountIdentifier == null) {
+            return;
+        }
+        mAuthenticationClient.getAccounts(new PublicClientApplication.AccountsLoadedListener() {
+
+            @Override
+            public void onAccountsLoaded(List<IAccount> accounts) {
+                for (IAccount account : accounts) {
+                    if (account.getHomeAccountIdentifier().getIdentifier().equals(homeAccountIdentifier)) {
+                        mAuthenticationClient.removeAccount(account);
```
guperrot: threading unsafe here, mAuthenticationClient should be accessed from worker thread (it mutates in worker thread) AND checking we are still enabled.

To do that simply wrap the entire code of the callback into
```
post(new Runnable() {

     @Override
     public void run() {
            for (IAcccont....

     }
}
```

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -380,6 +394,34 @@ public void run() {
         });
     }
 
+    private synchronized void instanceSignOut() {
+        if (mTokenStorage.getToken() == null) {
+            AppCenterLog.warn( LOG_TAG, "Couldn't sign-out: authToken doesn't exist.");
+            return;
+        }
+        mSignInDelayed = false;
```
guperrot: Those 3 lines 402-404 should be shared with disabled code as disables does a sign out as part of disabling the module.

MatkovIvan: @guperrot Do we need to remove an account on disabling the module?

guperrot: signOut is a sub-set of disabling

PR #965, opened 2019-02-26
Refactored Token Exchange and Cosmos DB service calls. Added future management.

nikitag-ms:
- Refactored Token Exchange and Cosmos DB service calls. 
- Added future management.

### DIFF
--- a/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
+++ b/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
@@ -5,20 +5,21 @@ import android.content.Context;
 import android.support.annotation.NonNull;
 import android.support.annotation.VisibleForTesting;
 
-import com.google.gson.Gson;
 import com.microsoft.appcenter.AbstractAppCenterService;
 import com.microsoft.appcenter.channel.Channel;
+import com.microsoft.appcenter.http.HttpClient;
+import com.microsoft.appcenter.http.ServiceCall;
 import com.microsoft.appcenter.http.ServiceCallback;
 import com.microsoft.appcenter.storage.client.CosmosDb;
 import com.microsoft.appcenter.storage.client.TokenExchange;
 import com.microsoft.appcenter.storage.models.Document;
 import com.microsoft.appcenter.storage.models.Documents;
 import com.microsoft.appcenter.storage.models.TokenResult;
-import com.microsoft.appcenter.storage.models.TokensResponse;
 import com.microsoft.appcenter.utils.AppCenterLog;
 import com.microsoft.appcenter.utils.async.AppCenterFuture;
 import com.microsoft.appcenter.utils.async.DefaultAppCenterFuture;
 
+import java.util.HashMap;
 import java.util.Map;
 
 import static com.microsoft.appcenter.Constants.DEFAULT_API_URL;
@@ -54,11 +55,9 @@ public class Storage extends AbstractAppCenterService {
      */
     private String mApiUrl = DEFAULT_API_URL;
 
-    private final Gson gson;
+    private Map<DefaultAppCenterFuture<?>, ServiceCall> mPendingCalls = new HashMap<>();
 
-    private Storage() {
-        gson = new Gson();
-    }
+    private HttpClient mHttpClient;
 
     /**
      * Get shared instance.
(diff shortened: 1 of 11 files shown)


sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
```diff
@@ -151,72 +156,70 @@ protected String getLoggerTag() {
...
+        // https://docs.microsoft.com/en-us/rest/api/cosmos-db/get-a-document
+        ServiceCall cosmosDbCall = CosmosDb.callCosmosDbApi(
+                tokenResult,
+                documentId,
+                mHttpClient,
+                METHOD_GET,
+                "",
+                new ServiceCallback() {
+                    @Override
+                    public void onCallSucceeded(final String payload, Map<String, String> headers) {
```
guperrot: final not needed here, we use final on local variables or parameters only when needed.
While member/static variables must have final if they can be.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
```diff
@@ -299,46 +307,68 @@ public void completeFuture(Exception e) {
...
+        return result;
+    }
+
+    private synchronized void callCosmosDbDeleteApi(TokenResult tokenResult, String documentId, final DefaultAppCenterFuture<Document<Void>> result) {
+        CosmosDb.callCosmosDbApi(
+                tokenResult,
+                documentId,
+                mHttpClient,
+                METHOD_DELETE,
+                "",
```
guperrot: null

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Utils.java
```diff
@@ -0,0 +1,23 @@
+package com.microsoft.appcenter.storage;
+
+import com.google.gson.Gson;
+import com.microsoft.appcenter.http.HttpException;
+import com.microsoft.appcenter.http.HttpUtils;
+import com.microsoft.appcenter.utils.AppCenterLog;
+
+public final class Utils {
+    public static final Gson sGson = new Gson();
```
guperrot: add a blank line before the static field

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Utils.java
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * Handle API call failure.
```
guperrot: When a method is public, we will get a warning if we don't document parameter when using Javadoc syntax.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
```diff
@@ -151,72 +156,70 @@ protected String getLoggerTag() {
...
-
-
     // Read a document
     // The document type (T) must be JSON deserializable
     public static <T> AppCenterFuture<Document<T>> read(String partition, String documentId){
         AppCenterLog.debug(LOG_TAG, "Read started");
         return getInstance().instanceRead(partition, documentId);
     }
 
+    //region Read implementation
```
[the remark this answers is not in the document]
nikitag-ms: Yeah, works exactly the same way as C# `#region` in Android Studio

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
```diff
@@ -151,194 +156,216 @@ protected String getLoggerTag() {
...
+                        completeFutureAndRemovePendingCall(e, result);
+                    }});
+        mPendingCalls.put(result, cosmosDbCall);
     }
 
+    //endregion
+
+    //region List implementation
+
     // List (need optional signature to configure page size)
```
guperrot: all comments I meant for `/*`, I commented only in 1 place but was global feedback.

Also do region comments have to be `//` to work?

nikitag-ms: Yeah, it's an IDE thing. I don't think it would work with /* comments. It's not a comment, technically, but an instruction to IDE to be able to collapse and expand sections of code.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
```diff
@@ -151,194 +156,216 @@ protected String getLoggerTag() {
...
     // The document type (T) must be JSON deserializable
```
guperrot: My comment applied to Javadoc though not region. Use `/**` for Javadoc like everywhere else we have Javadoc.

PR #979, opened 2019-03-05
Feature/identity acquire token silently

t-nafle:

* [x] Did you add unit tests?

## Description

Adding the ability to sign in silently if the token is not expired.
Adding tests for cases where sign in is silent.


sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -231,7 +236,7 @@ protected String getLoggerTag() {
     public synchronized void onActivityResumed(Activity activity) {
         mActivity = activity;
         if (mSignInDelayed) {
-            signInFromUI();
+            signInInteractively();
```
guperrot: This is actually wrong as it does not try to use silent sign-in but we are going to remove delayed sign-in very soon so we can let this go I guess.

guperrot: Expecting fix for this in #982 that should remove delayed sign-in.

Issue #975, opened 2019-03-03
ANR because of deadlock in the SDK


techee:
I started receiving the following ANRs in the Play Console after I started using the 1.11.2 and 1.11.3 releases:

```
"main" prio=5 tid=1 Blocked
  | group="main" sCount=1 dsCount=0 flags=1 obj=0x72b35c20 self=0x727bcc3e00
  | sysTid=17629 nice=0 cgrp=default sched=0/0 handle=0x73018aa9a8
  | state=S schedstat=( 8674365711 917464738 12512 ) utm=673 stm=196 core=0 HZ=100
  | stack=0x7fe7f1a000-0x7fe7f1c000 stackSize=8MB
  | held mutexes=
  at com.microsoft.appcenter.utils.NetworkStateHelper.a (NetworkStateHelper.java)
- waiting to lock <0x095ef410> (a o.aqu) held by thread 16
  at com.microsoft.appcenter.http.HttpClientNetworkStateHandler.callAsync (HttpClientNetworkStateHandler.java:43)
- locked <0x018348ba> (a o.aot)
  at com.microsoft.appcenter.http.HttpClientCallDecorator.run (HttpClientCallDecorator.java:49)
- locked <0x002afb09> (a o.aou$a)
  at android.os.Handler.handleCallback (Handler.java:790)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loop (Looper.java:192)
  at android.app.ActivityThread.main (ActivityThread.java:6703)
  at java.lang.reflect.Method.invoke (Method.java)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:438)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:826)
```
```
"ConnectivityThread" prio=5 tid=16 Blocked
  | group="main" sCount=1 dsCount=0 flags=1 obj=0x13880860 self=0x726ffa8400
  | sysTid=17656 nice=0 cgrp=default sched=0/0 handle=0x726230e4f0
  | state=S schedstat=( 15532083 13352710 50 ) utm=0 stm=1 core=0 HZ=100
  | stack=0x726220c000-0x726220e000 stackSize=1037KB
  | held mutexes=
  at o.aot.a (SourceFile)
- waiting to lock <0x018348ba> (a o.aot) held by thread 1
  at o.aqu.a (SourceFile:237)
[...]

PR #983, opened 2019-03-05
Fixed deadlock in NetworkStateHelper

MatkovIvan:
* [x] Did you add unit tests?

## Related PRs or issues

#975

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
@@ -17,8 +17,8 @@ import android.support.annotation.VisibleForTesting;
 
 import java.io.Closeable;
 import java.util.Arrays;
-import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import static android.content.Context.CONNECTIVITY_SERVICE;
@@ -48,7 +48,7 @@ public class NetworkStateHelper implements Closeable {
     /**
      * Network state listeners that will subscribe to us.
      */
-    private final Set<Listener> mListeners = new HashSet<>();
+    private final Set<Listener> mListeners = new CopyOnWriteArraySet<>();
 
     /**
      * Network callback, null on API level < 21.
@@ -93,7 +93,7 @@ public class NetworkStateHelper implements Closeable {
     /**
      * Make this helper active again after closing.
      */
-    public synchronized void reopen() {
+    public void reopen() {
         try {
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
 
@@ -189,7 +189,7 @@ public class NetworkStateHelper implements Closeable {
      * Handle network available update on API level >= 21.
      */
     @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
-    private synchronized void onNetworkAvailable(Network network) {
+    private void onNetworkAvailable(Network network) {
         AppCenterLog.debug(LOG_TAG, "Network "+ network + " is available.");
         if (mConnected.compareAndSet(false, true)) {
             notifyNetworkStateUpdated(true);
@@ -200,7 +200,7 @@ public class NetworkStateHelper implements Closeable {
      * Handle network lost update on API level >= 21.
      */
     @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
-    private synchronized void onNetworkLost(Network network) {
+    private void onNetworkLost(Network network) {
         AppCenterLog.debug(LOG_TAG, "Network "+ network + " is lost.");
         Network[] networks = mConnectivityManager.getAllNetworks();
         boolean noNetwork = networks == null || networks.length == 0 ||
@@ -213,7 +213,7 @@ public class NetworkStateHelper implements Closeable {
     /**
      * Handle network state update on API level < 21.
      */
-    private synchronized void handleNetworkStateUpdate() {
+    private void handleNetworkStateUpdate() {
         boolean connected = isAnyNetworkConnected();
         if (mConnected.compareAndSet(!connected, connected)) {
             notifyNetworkStateUpdated(connected);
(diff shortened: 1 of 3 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -190,8 +191,7 @@ private boolean isAnyNetworkConnected() {
     @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
     private synchronized void onNetworkAvailable(Network network) {
```
guperrot: This synchronized can be removed.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -205,8 +205,7 @@ private synchronized void onNetworkLost(Network network) {
         Network[] networks = mConnectivityManager.getAllNetworks();
```
guperrot: No need to use synchronized anymore in that method as well.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -216,8 +215,7 @@ private synchronized void onNetworkLost(Network network) {
      */
     private synchronized void handleNetworkStateUpdate() {
```
guperrot: remove synchronized as well

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -236,7 +234,7 @@ private void notifyNetworkStateUpdated(boolean connected) {
 
     @Override
     public synchronized void close() {
```
guperrot: remove synchronized

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -216,8 +215,7 @@ private synchronized void onNetworkLost(Network network) {
      */
     private synchronized void handleNetworkStateUpdate() {
         boolean connected = isAnyNetworkConnected();
-        if (connected != mConnected) {
-            mConnected = connected;
+        if (mConnected.compareAndSet(!connected, connected)) {
             notifyNetworkStateUpdated(connected);
         }
     }
```
guperrot: Use a concurrent map for listeners and remove synchronized everywhere.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
```diff
@@ -62,15 +63,15 @@
...
+    private final AtomicBoolean mConnected = new AtomicBoolean();
 
     /**
      * Init.
      *
      * @param context any Android context.
      */
     @VisibleForTesting
-    NetworkStateHelper(Context context) {
+    public NetworkStateHelper(Context context) {
```
jaelim-ms: Since this becomes a public method, you can remove the annotation for `VisibleForTesting`.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
jaelim-ms: Shouldn't we keep it? It calls `notifyNetworkStateUpdated(true);` in the method.

guperrot:
Please take this opportunity to fix:
```
> Task :sdk:appcenter:compileDebugJavaWithJavac
/Users/guperrot/git/appcenter-sdk-android/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java:174: warning: [deprecation] getAllNetworkInfo() in ConnectivityManager has been deprecated
            NetworkInfo[] networks = mConnectivityManager.getAllNetworkInfo();
                                                         ^
1 warning
```

Isolating method and using annotation, comments don't work at javac level, they work only in IDE.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
guperrot: See https://github.com/Microsoft/AppCenter-SDK-Android/pull/983/files#r262664710

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
guperrot: It's public only because of tests I think so it's actually what this annotation is used for.

jaelim-ms: Ah.. It is used in unit tests in different packages. I see.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/NetworkStateHelper.java
jaelim-ms: Didn't hear any answers for my previous comment, @guperrot can you explain why we don't need this `synchronized`?

jaelim-ms:
I'm not blocking this PR since it fixes deadlock issue but it has another issue that network state order could be wrong since we are removing `synchronized`.As the result of this behavior the SDK can have opposite network state, I already verified the behavior by calling `onNetworkAvailable` and `onNetworkLost` many times in different threads. We need to fix this issue but it could be in another PR.

guperrot:
> I'm not blocking this PR since it fixes deadlock issue but it has another issue that network state order could be wrong since we are removing `synchronized`.As the result of this behavior the SDK can have opposite network state, I already verified the behavior by calling `onNetworkAvailable` and `onNetworkLost` many times in different threads. We need to fix this issue but it could be in another PR.

Updates are always delivered by Android from the same thread so we don't have ordering issues.

guperrot:
nit: `Collections.newSetFromMap(new ConcurrentHashMap<Listener, Boolean>());`

PR #986, opened 2019-03-06
Fix in-app update for MIUI 10

maximkornilov:

* [x] Did you add unit tests?

## Description

From MIUI 10, it started to return `com.miui.packageinstaller` as application package name instead of null for local installed applications. It causes the problem that in-app update stops working because it treats application as installed from a store. So, 'com.miui.packageinstaller` should be added as know local stores to identify correctly that applicaiton is not installed from a store.

## Misc
Output of the app center logger with Verbose log level enabled (before fix): 
```
D/AppCenterDistribute: InstallerPackageName=com.miui.packageinstaller
I/AppCenterDistribute: Not checking in app updates as installed from a store.
```

### DIFF
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/InstallerUtils.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/InstallerUtils.java
@@ -38,6 +38,7 @@ class InstallerUtils {
         LOCAL_STORES.add("com.android.packageinstaller");
         LOCAL_STORES.add("com.google.android.packageinstaller");
         LOCAL_STORES.add("com.android.managedprovisioning");
+        LOCAL_STORES.add("com.miui.packageinstaller");
     }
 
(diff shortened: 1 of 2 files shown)


guperrot:
Hi, thanks for your patch, just for more context did you install from ADB or from a downloaded file?

maximkornilov:
The application was installed using HockeyApp. I also tried to install using the downloaded file. In both cases, the application installed package name is set to `com.miui.packageinstaller`.

guperrot:
Thanks for the additional context and for having signed CLA, I can merge the PR now, we'll take care of the changelog.

guperrot:
Hi, we have a fix for that bug.

The SDK is not released yet (this should happen before end of month) so if it's urgent you can pull a pre-release by adding the pre-release repository:

```
repositories {
    maven {
        url  "https://dl.bintray.com/vsappcenter/appcenter-snapshot" 
    }
}
```

And use version `1.11.4-5+b15effb0` of the SDK modules.

PR #997, opened 2019-03-12
Implement Identity.setConfigUrl and change default to prod

guperrot:

* [x] Did you add unit tests?

## Description

* Add Identity.setConfigUrl to change remote configuration endpoint.
* Change default value for prod.
* AB#58365


guperrot:
> I wonder if config needs to be redownloaded if the new value is different. Or log a warning 🤔

It's not going to be used by customers most likely, we just call it before start in test apps.
We can revisit if there is an actual need.

PR #1006, opened 2019-03-18
Add history to token store

MatkovIvan:
* [x] Did you add unit tests?

## Description

Sending logs is an async operation and there is a chance that token will be updated multiple times during sending.
We need to store information when each token was valid.

## Related PRs or issues

[AB#58502](https://msmobilecenter.visualstudio.com/web/wi.aspx?pcguid=27bf89de-9697-418a-accf-1aa125b22914&id=58502)


guperrot:
No need for in-progress label, this was introduced for advertising progress on issues labeled for feature-request.

For PRs, draft is all we need (or dont-merge when forgetting to use draft).

sdk/appcenter-identity/src/androidTest/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorageAndroidTest.java
```diff
@@ -0,0 +1,183 @@
...
+        /* Remove the token from storage. */
+        tokenStorage.saveToken(null, null, null);
+
+        /* Assert that there's no token in storage. */
+        assertNull(tokenStorage.getToken());
+        assertNull(tokenStorage.getHomeAccountId());
+
+        /* The same token should't be in history twice in a row. */
+        tokenStorage.saveToken(AUTH_TOKEN, ACCOUNT_ID, new Date());
+        tokenStorage.saveToken(AUTH_TOKEN, ACCOUNT_ID, new Date());
```
guperrot: We should use 2 fixed different dates as currently those 2 dates will likely be the same millisecond and thus equal so not really proving the point that we are trying to test (even if dates different if same value we don't store twice).

sdk/appcenter-identity/src/androidTest/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorageAndroidTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+        /* Check history. */
+        assertEquals(4, tokenStorage.loadTokenHistory().size());
```
guperrot: wait, why 4?

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -124,11 +125,6 @@
      */
     private DefaultAppCenterFuture<SignInResult> mPendingSignInFuture;
 
-    /**
-     * Instance of {@link AuthTokenStorage} to store token information.
-     */
-    private AuthTokenStorage mTokenStorage;
```
guperrot: I kinda liked having a filed instead of having to use getInstance() all the time. Any reason for not having a field anymore?

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/TokenStorageFactory.java
```diff
@@ -7,6 +7,8 @@
 
 import android.content.Context;
 
+import com.microsoft.appcenter.utils.storage.AuthTokenStorage;
```
guperrot: was that to fix a Javadoc warning?

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/utils/storage/AuthTokenTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.Date;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+
+@SmallTest
+@RunWith(AndroidJUnit4.class)
```
guperrot: Why do we need those 2 annotations?

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -83,23 +91,41 @@ public synchronized String getAuthToken() {
...
-    public synchronized void setAuthToken(String authToken, String homeAccountId) {
+    public synchronized void setAuthToken(String authToken, String homeAccountId, Date expiresOn) {
+        if (mStorage != null) {
+            mStorage.saveToken(authToken, homeAccountId, expiresOn);
+        }
+        updateAuthToken(authToken, homeAccountId);
+    }
+
+    private void updateAuthToken(String authToken, String homeAccountId) {
+        final boolean isNewUser = isNewUser(homeAccountId);
```
guperrot:
```suggestion
        boolean isNewUser = isNewUser(homeAccountId);
```

guperrot: but why do we need a local variable now? It's used only once.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/AuthTokenStorage.java
```diff
@@ -33,12 +38,16 @@
     String getHomeAccountId();
 
     /**
-     * Removes token value.
+     * Gets the oldest token info from history.
+     *
+     * @return auth token info.
      */
-    void removeToken();
+    AuthTokenInfo getOldestToken();
```
guperrot: Why do we need a getOldestToken?

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
```diff
@@ -8,48 +8,103 @@
...
+
+        /* Limit history size. */
+        if (history.size() > TOKEN_HISTORY_LIMIT) {
+            AppCenterLog.debug(LOG_TAG, "Size of the token history is exceeded. The oldest token has been removed.");
+            history.remove(0);
+        }
+
+        /* Update history and current token. */
+        saveTokenHistory(history);
+        if (token != null) {
```
guperrot: Why do we need to persist the current value on disk if history is on disk? Current can be in memory since all history is on disk.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
MatkovIvan: 1. It's used in the loop
2. We need to check in before changing `mHomeAccountId`

I'll add comment

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/AuthTokenStorage.java
MatkovIvan: To send the logs with the oldest token first. During sending we're interesting only for the oldest one, no need to have a full history outside of this class.

I'll add more description in Javadoc.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/TokenStorageFactory.java
MatkovIvan: No, this class was moved into another package (to core instead of identity module)

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
```diff
@@ -32,61 +69,153 @@
...
+            AppCenterLog.warn(LOG_TAG, "Failed to deserialize auth token history.", e);
+        }
+        mHistory = entities != null ? new ArrayList<>(Arrays.asList(entities)) : new ArrayList<TokenStoreEntity>();
+        return mHistory;
+    }
+
+    @VisibleForTesting
+    void setTokenHistory(List<TokenStoreEntity> history) {
+        mHistory = history;
+        String json = new Gson().toJson(history.toArray());
```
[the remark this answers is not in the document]
annakocheshkova: uh-oh I mentioned someone

sdk/appcenter-identity/src/test/java/com/microsoft/appcenter/identity/IdentityTest.java
```diff
@@ -1152,6 +1156,36 @@ public Void answer(InvocationOnMock invocationOnMock) {
...
+                .thenReturn(file);
+        when(file.exists()).thenReturn(true);
+        String config = jsonConfig.toString();
+        when(FileManager.read(file)).thenReturn(config);
+
+        /* Mock MSAL client. */
+        PublicClientApplication publicClientApplication = mock(PublicClientApplication.class);
+        whenNew(PublicClientApplication.class).withAnyArguments().thenReturn(publicClientApplication);
+        when(publicClientApplication.getAccount(anyString(), anyString())).thenReturn(null);
+
```
annakocheshkova:
```suggestion
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -109,33 +137,56 @@ public synchronized void setAuthToken(String authToken, String homeAccountId) {
...
-        mAuthToken = null;
-        mLastHomeAccountId = null;
-        for (Listener listener : mListeners) {
-            listener.onNewAuthToken(null);
-            listener.onNewUser(null);
+    public synchronized void clearAuthToken() {
+        setAuthToken(null, null, null);
+    }
+
+    public synchronized void cacheAuthToken() {
```
annakocheshkova: Comment :)

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -574,6 +581,9 @@ private synchronized void handleSendingSuccess(@NonNull GroupState groupState, @
                     groupListener.onSuccess(log);
                 }
             }
+            /* Removing token after successful sent logs related to this token */
```
annakocheshkova: New line

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -454,10 +459,13 @@ private synchronized void triggerIngestion(final @NonNull GroupState groupState)
             return;
         }
 
+        /* Get oldest token to getting batch by endTime of this token*      */
```
annakocheshkova:
```suggestion
        /* Get oldest token to getting batch by endTime of this token. */
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -120,6 +120,11 @@
      */
     private int mCurrentState;
 
+    /****
+     *
```
annakocheshkova: What is that? :D

sdk/appcenter-identity/src/androidTest/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorageAndroidTest.java
```diff
@@ -0,0 +1,183 @@
...
+        /* Removing does nothing. */
+        tokenStorage.removeToken(null);
+
+        /* Save the token into storage. */
+        tokenStorage.saveToken(AUTH_TOKEN, ACCOUNT_ID, new Date());
+        tokenStorage.saveToken("test", ACCOUNT_ID, new Date());
+        assertEquals(3, tokenStorage.loadTokenHistory().size());
+
+        /* Remove last one. */
+        tokenStorage.removeToken("test");
```
annakocheshkova: Please add a string with "test" and use it as parameter

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -203,7 +199,8 @@ public static void signOut() {
     public synchronized void onStarted(@NonNull Context context, @NonNull Channel channel, String appSecret, String transmissionTargetToken, boolean startedFromApp) {
         mContext = context;
         mAppSecret = appSecret;
-        mTokenStorage = TokenStorageFactory.getTokenStorage(context);
+        AuthTokenContext authTokenContext = AuthTokenContext.getInstance();
+        authTokenContext.setStorage(TokenStorageFactory.getTokenStorage(context));
```
annakocheshkova: I don't like this implicit logic very much.. What will happen if a developer unfamiliar with this logic forgets to setStorage?
Maybe we revert to how it was where we just passed context to  AuthTokenContext.getInstance()?

MatkovIvan: @annakocheshkova Token storage implemented in the identity module, context in core.
It was a conscious decision. No, getInstance didn't have context parameter

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
```diff
@@ -32,61 +69,153 @@
...
+            json = decryptedData.getDecryptedData();
+        }
+        if (json == null || json.isEmpty()) {
+            return null;
+        }
+        TokenStoreEntity[] entities = null;
+        try {
+            entities = new Gson().fromJson(json, TokenStoreEntity[].class);
+        } catch (JsonParseException e) {
+            AppCenterLog.warn(LOG_TAG, "Failed to deserialize auth token history.", e);
```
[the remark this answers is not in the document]
MatkovIvan: Why? It can affect behavior, but still recoverable. It's a warning

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
```diff
@@ -32,61 +69,153 @@
...
-        String tokenFromStorage = getToken();
-        String homeAccountId = getHomeAccountId();
-
-        /* We need to update Token context here if the values are not null. */
-        if (tokenFromStorage != null && homeAccountId != null) {
-            AuthTokenContext.getInstance().setAuthToken(tokenFromStorage, homeAccountId);
+    public synchronized AuthTokenInfo getOldestToken() {
+        List<TokenStoreEntity> history = getTokenHistory();
+        if (history == null || history.size() == 0) {
+            return new AuthTokenInfo(null, null, null);
```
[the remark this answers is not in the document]
MatkovIvan: The meaning of it: anonymous that valid from the beginning of time and still valid.
It's commented inside `AuthTokenInfo`

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
MatkovIvan: It's used inside a context now, not directly

sdk/appcenter-identity/src/androidTest/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorageAndroidTest.java
MatkovIvan: I've added the comments about it

https://github.com/Microsoft/AppCenter-SDK-Android/blob/8104843362e714a879e9bc693a1048592237304f/sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java#L78-L83

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
```diff
@@ -32,61 +69,161 @@
...
+        }
     }
 
     @Override
-    public String getToken() {
-        String encryptedToken = SharedPreferencesManager.getString(PREFERENCE_KEY_AUTH_TOKEN, null);
-        if (encryptedToken == null || encryptedToken.length() == 0) {
-            return null;
+    public synchronized String getToken() {
+        List<TokenStoreEntity> history = getTokenHistory();
```
guperrot: I think there was a misunderstanding on my last comment about in memory.
We call this method often and this decrypts the full history every time.

What I said is that instead of having a persistent key just for latest token, we can cache in memory the latest value since its frequently used.

Is the reason we are afraid of memory dumps of some kind? If there is a security concern then might as well roll back that part to a single key for current, otherwise in memory.

The token will live in memory for offline calls anyway in http network state decorator, or all pending calls in general.

MatkovIvan: Not anymore. See https://github.com/Microsoft/AppCenter-SDK-Android/pull/1006/files#diff-7ef491a2f4f5c31272e9f26cd400e0e8R61

guperrot: The part I missed is actually https://github.com/Microsoft/AppCenter-SDK-Android/pull/1006/files#diff-7ef491a2f4f5c31272e9f26cd400e0e8R165

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/storage/PreferenceTokenStorage.java
```diff
@@ -32,61 +69,161 @@
...
+        if (json == null || json.isEmpty()) {
+            return null;
+        }
+        TokenStoreEntity[] entities = null;
+        try {
+            entities = new Gson().fromJson(json, TokenStoreEntity[].class);
+        } catch (JsonParseException e) {
+            AppCenterLog.warn(LOG_TAG, "Failed to deserialize auth token history.", e);
+        }
+        mHistory = entities != null ? new ArrayList<>(Arrays.asList(entities)) : new ArrayList<TokenStoreEntity>();
```
guperrot: Gson knows list and generics too, no need to use an intermediate array. Need to use the TypeToken syntax though.

MatkovIvan: 1. `TypeToken` 10x slower
2. List instead of array not work in test env (spent some time on it have no idea why exactly)

PR #1008, opened 2019-03-18
add token to the request db header and account id into the response

YiMysty:

* [x] Did you add unit tests?

## Description

Implement token exchange API, using user identity when available.

## Related PRs or issues

[AB#58606](https://msmobilecenter.visualstudio.com/web/wi.aspx?pcguid=27bf89de-9697-418a-accf-1aa125b22914&id=58606)

### DIFF
--- a/sdk/appcenter-storage/build.gradle
+++ b/sdk/appcenter-storage/build.gradle
@@ -10,6 +10,7 @@ dependencies {
     api project(':sdk:appcenter')
 
     implementation 'com.google.code.gson:gson:2.8.5'
+    compile project(path: ':sdk:appcenter-identity')
 
 }
 
--- a/sdk/appcenter-storage/build.gradle
+++ b/sdk/appcenter-storage/build.gradle
@@ -10,7 +10,6 @@ dependencies {
     api project(':sdk:appcenter')
 
     implementation 'com.google.code.gson:gson:2.8.5'
-    compile project(path: ':sdk:appcenter-identity')
 
 }
 
--- a/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
+++ b/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
@@ -68,6 +68,8 @@ public class Storage extends AbstractAppCenterService {
 
     private HttpClient mHttpClient;
 
+    private Context mConntext;
+
     private DocumentCache mDocumentCache;
 
     /**
@@ -203,6 +205,7 @@ public class Storage extends AbstractAppCenterService {
 
     @Override
     public synchronized void onStarted(@NonNull Context context, @NonNull Channel channel, String appSecret, String transmissionTargetToken, boolean startedFromApp) {
+        mConntext = context;
         mNetworkStateHelper = NetworkStateHelper.getSharedInstance(context);
         mHttpClient = createHttpClient(context);
         mAppSecret = appSecret;
@@ -481,6 +484,7 @@ public class Storage extends AbstractAppCenterService {
                     TokenExchange.getDbToken(
                             partition,
                             mHttpClient,
+                            mConntext,
                             mApiUrl,
                             mAppSecret,
                             callback);
--- a/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
+++ b/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
@@ -68,8 +68,6 @@ public class Storage extends AbstractAppCenterService {
 
     private HttpClient mHttpClient;
 
-    private Context mConntext;
-
     private DocumentCache mDocumentCache;
 
     /**
@@ -134,6 +132,7 @@ public class Storage extends AbstractAppCenterService {
      * Read a document.
      * The document type (T) must be JSON deserializable.
      */
+    @SuppressWarnings("WeakerAccess") // TODO remove warning suppress after release.
     public static <T> AppCenterFuture<Document<T>> read(String partition, String documentId, Class<T> documentType) {
         return read(partition, documentId, documentType, new ReadOptions());
     }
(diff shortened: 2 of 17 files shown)


sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
```diff
@@ -68,6 +68,8 @@
 
     private HttpClient mHttpClient;
 
+    private Context mConntext;
```
YiMysty: do not know why other property name start with `m`... thus add m here. I thought `m` is short for mock?

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/client/TokenExchange.java
```diff
@@ -65,17 +70,22 @@ public static String buildAppCenterGetDbTokenBodyPayload(String partition) {
     public static synchronized ServiceCall getDbToken(
             final String partition,
             HttpClient httpClient,
+            Context context,
             String apiUrl,
             final String appSecret,
             TokenExchangeServiceCallback serviceCallback) {
         AppCenterLog.debug(LOG_TAG, "Getting a resource token from App Center...");
         String url = apiUrl + GET_TOKEN_PATH_FORMAT;
+        final String userToken = TokenStorageFactory.getTokenStorage(context).getToken();
```
YiMysty: This context is get when the Storage service initialized, is it using the same storage as identity module?

sdk/appcenter-storage/build.gradle
```diff
@@ -10,6 +10,7 @@ dependencies {
     api project(':sdk:appcenter')
 
     implementation 'com.google.code.gson:gson:2.8.5'
+    compile project(path: ':sdk:appcenter-identity')
```
guperrot: Remove that, we must not depend on Identity directly, everything you need is on the core.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
guperrot: m is android convention for member (variable) as opposed to s for static.

guperrot: typo: mContext

jwallra: Here's the convention: https://source.android.com/setup/contribute/code-style#follow-field-naming-conventions

```suggestion
    private Context mContext;
```

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/client/TokenExchange.java
guperrot: This pattern is the right one to use.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/models/TokenResult.java
```diff
@@ -123,6 +130,17 @@ public TokenResult withExpirationTime(Date expiresOn) {
         return this;
     }
 
+    /**
+     * Set the account id.
+     *
+     * @param accountId Account id value to be set.
+     * @return The TokenResult object itself.
+     */
+    public TokenResult withAccountId(String accountId) {
```
jwallra: The "with" convention is only used in apple SDK.

```suggestion
    public TokenResult setAccountId(String accountId) {
```

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
guperrot: We use that convention everywhere except for JSON models.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/models/TokenResult.java
```diff
@@ -212,6 +230,14 @@ public String status() {
         return this.status;
     }
 
+    /**
+     * Get account id.
+     * @return The account id value.
+     */
+    public String accountId() {
```
jwallra:
```suggestion
    public String getAccountId() {
```

sdk/appcenter-storage/build.gradle
guperrot: You're using the wrong class actually.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/client/TokenExchange.java
guperrot: You need to use `AuthTokenContext` instead.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/models/TokenResult.java
YiMysty: sure, will fix it latter :) overall files, currently the naming is in all models

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/models/TokenResult.java
YiMysty: will fix latter in another pr~

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/models/TokenResult.java
guperrot: Let's get that merged, then change everywhere when noone is modifying it.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/models/TokenResult.java
guperrot: Let's get that merged, then change everywhere when noone is modifying it.

PR #1011, opened 2019-03-19
Support opting in to using the Distribute sdk features in DEBUGGABLE builds

RyanThomas73:
* [x]  Did you add unit tests?

## Description

In select cases, consumers may wish to use the Distribute feature in DEBUGGABLE builds. This is a feature that we want support for so we can complete our migration from the hockey sdk to the app center sdk. 

In lieu of always preventing the Distribute feature from working in DEBUGGABLE builds, this PR prevents it by default but allows consumers to explicitly opt in for debuggable build support by calling `Distribute.setEnabledForDebuggableBuild(true);`

## Related PRs or issues

NONE

## Misc

In addition to the feature change, I split the unit checking debuggable builds took no action into three unit test cases.
* Testing that no action is taken for a DEBUGGABLE build when the default value is used.
* Testing that no action is taken for a DEBUGGABLE build when the consumer explicitly calls `setEnabledForDebuggableBuild(false)`.
* Testing that the Distribute feature continues normally for a DEBUGGABLE build if the consumer explicitly opts in via calling `setEnabledForDebuggableBuild(true)`.


guperrot:
Thanks for your contribution, every convention we have was followed so no comment on the code itself.

However, we need to discuss with the team as on iOS (and thus Xamarin), the logic is opposite: `Distribute.DontCheckForUpdatesInDebug();` that is called before start.

The reason for that is that on iOS we can only detect debugger attached and not the debug build itself.

If we change the default on Android and call it the same as iOS/Xamarin then that would allow us to make `Distribute.DontCheckForUpdatesInDebug();` portable and move it to the portable library on Xamarin/Unity and have a consistent experience there. (Xamarin/Unity use bindings to the iOS and Android SDKs and then expose a portable layer on APIs when possible).

guperrot:
Just to clarify, I am not asking to do any change at this change, will need to wait for team discussion.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -638,8 +663,8 @@ private synchronized void resumeDistributeWorkflow() {
         if (mPackageInfo != null && mForegroundActivity != null && !mWorkflowCompleted && isInstanceEnabled()) {
 
             /* Don't go any further it this is a debug app. */
-            if ((mContext.getApplicationInfo().flags & FLAG_DEBUGGABLE) == FLAG_DEBUGGABLE) {
-                AppCenterLog.info(LOG_TAG, "Not checking in app updates in debug.");
+            if ((mContext.getApplicationInfo().flags & FLAG_DEBUGGABLE) == FLAG_DEBUGGABLE && !mEnabledForDebuggableBuild) {
+                AppCenterLog.info(LOG_TAG, "Not checking for in app updates in debuggable build.");
```
jwallra: Small grammar correction

```suggestion
                AppCenterLog.info(LOG_TAG, "Not checking for in-app updates in debuggable build.");
```

RyanThomas73:
> DontCheckForUpdatesInDebug

RyanThomas73:
@guperrot To clarify the core iOS sdk doesn't implement `DontCheckForUpdatesInDebug( )`. The apple sdk uses [this logic](https://github.com/Microsoft/AppCenter-SDK-Apple/blob/199cc9fc2ffe9a4e2295956ecea1152d5da3daae/AppCenterDistribute/AppCenterDistribute/MSDistribute.m#L734).

The method you mentioned is implemented in the appcenter .net sdk but it doesn't bind directly to an iOS method, it sets the _internalBindingType to null:  [See .net SDK DontCheckForUpdatesInDebug( ) Method](https://github.com/Microsoft/AppCenter-SDK-DotNet/blob/26c80e05efaed11f0b3ee2ec79d3b6f0485bd2c1/SDK/AppCenterDistribute/Microsoft.AppCenter.Distribute.iOS/Distribute.cs#L66)

Because the core apple sdk does not implement the behavior your described and because the `setEnabled` / `isEnabled` naming pattern is more common, imo, my recommendation for your team is that we implement the following changes:

* Add a `setEnabledForDebuggableBuild(Boolean)` method to the apple sdk.
* If desired, add accompanying `isEnabledForDebuggableBuild( )` methods to the android/apple sdks.
* Updated the .net sdk to include bindings + implementations for the above method(s).
* For backwards compatibility, keep the `DontCheckForUpdatesInDebug( )` method but deprecate it in favor of the new set method. Update the logic of `DontCheckForUpdatesInDebug( )` to call the new method `setEnabledForDebuggableBuild(false);`

guperrot:
@RyanThomas73 I would agree if we could actually implement `setEnabledForDebuggableBuild` in Objective C but it's not possible to detect debug vs release on iOS from a library.

We would still need the old method for Xamarin.iOS if we go that route to disable checks on debug completely (whether debugger attached or not, if you relaunch the app from home screen it's still a debug build but no debugger attached), so with that solution we could not deprecate that method if `setEnabledForDebuggableBuild` would just check for debugger attached.

Also we actually document how to disable debug checks completely on iOS:
On Native iOS: https://docs.microsoft.com/en-us/appcenter/sdk/distribute/ios#dont-initialize-app-center-distribute-during-development
On Xamarin iOS it's just a helper method to avoid using a pragma but is equivalent (different implementation but equivalent as a feature).

If we introduce `setEnabledForDebuggableBuild` for iOS then the behavior would be different from Android which could be fine but in that case we cannot deprecate Xamarin.iOS `DontCheckForUpdatesInDebug` and would have both APIs side by side.

I will forward the feedback to the team.

RyanThomas73:
@guperrot Thanks much.

I'd like to submit a similar PR for the Apple sdk so we can use the Distribute sdk with a debugger attached. Does your team have a preference on the method name for that scenario since the behavior is not identical?

e.g. `setEnabledWithDebuggerAttached(...)`

guperrot:
For that specific use case, we could actually use the same method name I think since the goal is to disable the default behavior. A debuggable build could mean debugger is attached in that platform. If the iOS experts think differently they will comment on the PR.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -584,6 +602,13 @@ private synchronized void setInstanceListener(DistributeListener listener) {
         mListener = listener;
     }
 
+    /**
+     * Implements {@link #setEnabledForDebuggableBuild(Boolean)}.
```
```suggestion
     * Implements {@link #setEnabledForDebuggableBuild(boolean)}.
```

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -368,6 +376,16 @@ public static void setListener(DistributeListener listener) {
         getInstance().setInstanceListener(listener);
     }
 
+    /**
+     * Change whether the distribute service can be used within a DEBUGGABLE build.
```
```suggestion
     * Change whether the distribute service can be used within a debuggable build.
```

PR #1012, opened 2019-03-21
Add offline mode simulation to Storage

nikitag-ms:
* [x] Did you add unit tests?

## Description

Implemented a version of `HttpDecorator` that handles offline mode simulation flag and used it at the top of the existing decorator hierarchy.

### DIFF
--- a/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
+++ b/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
@@ -11,6 +11,7 @@ import android.support.annotation.NonNull;
 import android.support.annotation.VisibleForTesting;
 
 import com.microsoft.appcenter.AbstractAppCenterService;
+import com.microsoft.appcenter.AppCenter;
 import com.microsoft.appcenter.channel.Channel;
 import com.microsoft.appcenter.http.ServiceCall;
 import com.microsoft.appcenter.http.ServiceCallback;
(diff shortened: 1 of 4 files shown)


sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
```diff
@@ -128,6 +128,54 @@ public static void setApiUrl(String apiUrl) {
...
     }
 
+    /**
+     * Check whether offline simulation is enabled or not.
+     *
+     * @return result being <code>true</code> if enabled, <code>false</code> otherwise.
+     * @see AppCenterFuture
+     */
+    @SuppressWarnings({"unused", "WeakerAccess"}) // TODO Remove warning suppress after release.
+    public static boolean isSimulateOffline() {
```
guperrot: It should be simulating right?

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
```diff
@@ -128,6 +128,54 @@ public static void setApiUrl(String apiUrl) {
...
+
+    /**
+     * Enable or disable offline simulation.
+     *
+     * @param offline <code>true</code> to simulate device being offline, <code>false</code> to go back to the original network state of the device.
+     */
+    @SuppressWarnings({"unused", "WeakerAccess"}) // TODO Remove warning suppress after release.
+    public void setSimulateOfflineInstance(boolean offline) {
+        if (mHttpClient != null) {
+            mHttpClient.setSimulateOffline(offline);
```
guperrot: so we don't resume operations when coming back online?
this is different than what network state handler is doing that will resume calls when network back. I think we need to talk more about this feature.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
nikitag-ms: `isNetworkOfflineSimulationEnabled`

PR #1019, opened 2019-03-21
Update draft documents storage to use new SQLite schema

nikitag-ms:
* [ ] Did you add unit tests?

## Description

Update draft documents storage to use new SQLite schema:

- Added extra fields to the schema and renamed non-matching existing ones
- Renamed `DocumentCache` to `LocalDocumentStorage`

## Related PRs or issues

Did not change any of the behaviors yet, which still do not match the design in full


sdk/appcenter-storage/src/androidTest/java/com/microsoft/appcenter/storage/LocalDocumentStorageAndroidTest.java
```diff
@@ -25,7 +25,7 @@
 import static org.junit.Assert.assertTrue;
 
 @SuppressWarnings("unused")
-public class DocumentCacheAndroidTest {
+public class LocalDocumentStorageAndroidTest {
```
nikitag-ms:
```suggestion
public class LocalDocumentStorageAndroidTest2 {
```

PR #1032, opened 2019-04-01
Add @NonNull annotation to DocumentError's constructor

dhei:

* [ ] Did you add unit tests?

## Description

1. Add @NonNull annotation to DocumentError's constructor.
2. Simplified a parameter in Storage that only used once. No functional change.

### DIFF
--- a/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/models/DocumentError.java
+++ b/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/models/DocumentError.java
@@ -5,6 +5,8 @@
 
 package com.microsoft.appcenter.storage.models;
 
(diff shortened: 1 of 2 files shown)


sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/models/DocumentError.java
```diff
@@ -16,7 +18,7 @@
      * @param exception that occurred during the network state change.
      */
     @SuppressWarnings("WeakerAccess")
-    public DocumentError(Throwable exception) {
+    public DocumentError(@NonNull Throwable exception) {
```
guperrot: Do not use support annotations in models or public APIs. We don't depend on those in jCenter artifacts and it can cause weird warnings for customers not having the dependency.

PR #1036, opened 2019-04-02
Add samsung installer to local stores

MatkovIvan:
* [x] Did you add unit tests?

## Description

In-app updates don't work on Samsung S8 and S9


sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/InstallerUtils.java
```diff
@@ -44,6 +44,7 @@
         LOCAL_STORES.add("com.google.android.packageinstaller");
         LOCAL_STORES.add("com.android.managedprovisioning");
         LOCAL_STORES.add("com.miui.packageinstaller");
+        LOCAL_STORES.add("com.samsung.android.packageinstaller");
```
guperrot: I wonder if we should consider reversing the logic and check for stores instead, this logic comes from HockeyApp SDK but it does not seem to be scalable these days 🤔

smyshine: Maybe we can also add a config setting for client to enable store check or not.

PR #1038, opened 2019-04-02
Fix sending logs after refreshing

MatkovIvan:
* [x] Did you add unit tests?

## Description

Fix sending logs after refreshing

## Related PRs or issues

[AB#59543](https://msmobilecenter.visualstudio.com/web/wi.aspx?pcguid=27bf89de-9697-418a-accf-1aa125b22914&id=59543)

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
@@ -216,7 +216,7 @@ public class DefaultChannel implements Channel {
         /* Count pending logs. */
         groupState.mPendingLogCount = mPersistence.countLogs(groupName);
 
-        /* */
+        /* Listen for token refreshed to unblock sending logs after waiting for the token update. */
         AuthTokenContext.getInstance().addListener(groupState);
 
         /*
@@ -534,7 +534,7 @@ public class DefaultChannel implements Channel {
             return;
         }
 
-        /* Nothing to send. Some corrupted entries may be deleted, reset the counter. */
+        /* Some corrupted entries may be deleted, reset the counter to actual amount of logs pending to be sent. */
         groupState.mPendingLogCount = mPersistence.countLogs(groupState.mName);
     }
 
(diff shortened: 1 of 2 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -530,7 +535,7 @@ public void run() {
         }
 
         /* Nothing to send. Some corrupted entries may be deleted, reset the counter. */
-        groupState.mPendingLogCount = 0;
+        groupState.mPendingLogCount = mPersistence.countLogs(groupState.mName);
```
guperrot: Can you explain?
Also the comment should be updated.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -215,6 +216,9 @@ public synchronized void addGroup(final String groupName, int maxLogsPerBatch, l
         /* Count pending logs. */
         groupState.mPendingLogCount = mPersistence.countLogs(groupName);
 
+        /* */
```
guperrot:
```suggestion
        /* Listen for token refreshed to unblock sending logs after waiting for the token update. */
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
Jamminroot: Its not that we get fitting logs for existing authTokens, its the opposite - we iterate through tokens, and send logs that fit time bounds. Since we are adding a callback on token refresh (when we find expired token in iteration), we need to make sure we sent all the logs available - so we need to get that count and put it to mPendingLogCount.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -530,7 +535,7 @@ public void run() {
         }
 
         /* Nothing to send. Some corrupted entries may be deleted, reset the counter. */
```
Jamminroot:
```suggestion
        /* Some corrupted entries may be deleted, reset the counter to actual amount of logs pending to be sent. */
```

Jamminroot: Those are @guperrot suggested changes

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
MatkovIvan: @Jamminroot Thanks

PR #1040, opened 2019-04-03
Adding improved logging

jwallra:

* [x] Did you add unit tests?

## Description

~Previously cached tokens with failure status would be retrieved and attempted to be serialized.  Now we check for the failed status before attempting to serialize to avoid strange errors.~

After further inspection this check is handled in parseToken.  Just leaving the improved logging bits.

### DIFF
--- a/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
+++ b/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
@@ -52,6 +52,7 @@ import static com.microsoft.appcenter.storage.Constants.PENDING_OPERATION_DELETE
 import static com.microsoft.appcenter.storage.Constants.PENDING_OPERATION_REPLACE_VALUE;
 import static com.microsoft.appcenter.storage.Constants.SERVICE_NAME;
 import static com.microsoft.appcenter.storage.Constants.STORAGE_GROUP;
+import static com.microsoft.appcenter.storage.Constants.TOKEN_RESULT_SUCCEED;
 
 /**
  * Storage service.
@@ -677,12 +678,10 @@ public class Storage extends AbstractAppCenterService implements NetworkStateHel
             TokenExchangeServiceCallback callback) {
         TokenResult cachedTokenResult = TokenManager.getInstance().getCachedToken(partition);
         if (cachedTokenResult != null) {
-            int cachedTokenResultStatusCode = Integer.parseInt(cachedTokenResult.status());
-            if (cachedTokenResultStatusCode == 200) {
+            if (cachedTokenResult.status().equals(TOKEN_RESULT_SUCCEED)) {
                 callback.callCosmosDb(cachedTokenResult);
             } else {
-                AppCenterLog.error(LOG_TAG, String.format("The token store returned %d", cachedTokenResultStatusCode));
-                callback.onCallFailed(new HttpException(cachedTokenResultStatusCode));
+                callback.onCallFailed(new StorageException(String.format("The token status result was '%s'", cachedTokenResult.status())));
             }
         } else {
             ServiceCall tokenExchangeServiceCall =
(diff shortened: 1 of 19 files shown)


sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/TokenManager.java
```diff
@@ -59,13 +60,23 @@ public TokenResult getCachedToken(String partitionName) {
         if (token != null) {
             Calendar utcCalendar = Calendar.getInstance(TimeZone.getTimeZone("GMT"));
 
-            /* The token is considered expired. */
+            /* Token result was cached but status was not 'succeed'. */
+            if (!token.status().equals(Constants.TOKEN_RESULT_SUCCEED)) {
+                AppCenterLog.warn(Constants.LOG_TAG, String.format("Cached token result was failed for partition '%s'", partitionName));
```
guperrot: Import static for LOG_TAG usually.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/TokenManager.java
```diff
@@ -59,13 +60,23 @@ public TokenResult getCachedToken(String partitionName) {
         if (token != null) {
             Calendar utcCalendar = Calendar.getInstance(TimeZone.getTimeZone("GMT"));
 
-            /* The token is considered expired. */
+            /* Token result was cached but status was not 'succeed'. */
+            if (!token.status().equals(Constants.TOKEN_RESULT_SUCCEED)) {
```
YiMysty: I do not think we should cache the failed token... I would rather not add here, if we cache the failed token.

jwallra: @yochaude You had said you were getting strange errors because we were trying to serialize failed cached tokens for iOS.  Do you know how to repro this issue on a device?

jwallra:
@guperrot Thanks for the suggestion, found some other failing tests and fixed them :)

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
```diff
@@ -732,7 +734,11 @@ synchronized void getTokenAndCallCosmosDbApi(
             TokenExchangeServiceCallback callback) {
         TokenResult cachedTokenResult = TokenManager.getInstance().getCachedToken(partition);
         if (cachedTokenResult != null) {
-            callback.callCosmosDb(cachedTokenResult);
+            if (cachedTokenResult.status().equals(TOKEN_RESULT_SUCCEED)) {
```
YiMysty: yeah... is there status check required?

jwallra: Yes - if the token result is not 'succeed', that means that there was some error fetching the token.  You can look at the setup in the test we wrote, let us know if you think there is a problem or this situation is impossible.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
```diff
@@ -676,7 +677,13 @@ synchronized void getTokenAndCallCosmosDbApi(
             TokenExchangeServiceCallback callback) {
         TokenResult cachedTokenResult = TokenManager.getInstance().getCachedToken(partition);
         if (cachedTokenResult != null) {
-            callback.callCosmosDb(cachedTokenResult);
+            int cachedTokenResultStatusCode = Integer.parseInt(cachedTokenResult.status());
+            if (cachedTokenResultStatusCode == 200) {
```
guperrot: Why are we changing back to hardcoded 200 int now? I thought I saw something cleaner in previous iterations.

guperrot:
Not outdated: https://github.com/Microsoft/AppCenter-SDK-Android/pull/1040#discussion_r271978453

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
jwallra: I don't know what you're looking at, this is outdated...

guperrot: No idea, thought I clicked on files and just looking at diffs.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
guperrot: Was already done in token http callback.

PR #1043, opened 2019-04-03
Storage/cache for delete

anjbur:
* [x] Did you add unit tests?

## Description

Add caching on delete api.

## Related PRs or issues

## Misc

### DIFF
--- a/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/LocalDocumentStorage.java
+++ b/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/LocalDocumentStorage.java
@@ -195,12 +195,25 @@ class LocalDocumentStorage {
         return write(document, writeOptions, Constants.PENDING_OPERATION_REPLACE_VALUE);
     }
 
-    boolean writeDelete(String partition, String documentId) {
+    /**
+     * Creates or overwrites specified document entry in the cache. Sets the de-serialized value of
+     * the document to null and the pending operation to DELETE.
+     *
+     * @param partition  Partition key.
+     * @param documentId Document id.
+     * @return True if cache was successfully written to, false otherwise.
+     */
+    boolean markForDeletion(String partition, String documentId) {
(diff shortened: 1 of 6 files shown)


sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/LocalDocumentStorage.java
```diff
@@ -195,6 +195,12 @@ private static SQLiteQueryBuilder getPartitionAndDocumentIdQueryBuilder() {
         return write(document, writeOptions, Constants.PENDING_OPERATION_REPLACE_VALUE);
     }
 
+    boolean writeDelete(String partition, String documentId) {
```
guperrot: What's the difference between writeDelete and delete in the same class? This class methods really need some Javadoc as this is very confusing when we don't know the code base here.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
```diff
@@ -617,21 +617,30 @@ public void onCallFailed(Exception e) {
...
-                        Storage.this.completeFuture(e, result);
-                    }
-                });
+                        @Override
+                        public void completeFuture(Exception e) {
+                            Storage.this.completeFuture(e, result);
+                        }
+                    });
+        } else {
+            boolean isWriteSucceed = mLocalDocumentStorage.writeDelete(partition, documentId);
```
guperrot: This causes strict mode issues to access disk on the calling thread which is most of the time the UI thread.

Please catch up with develop and look at how we use async here in similar methods.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/LocalDocumentStorage.java
anjbur: Delete removes the document from the cache, writeDelete adds DELETE to the pending operations column. I’ll add javadoc to clarify the difference.

PR #1048, opened 2019-04-04
Catch all exceptions from reading throwable

MatkovIvan:
* [ ] Did you add unit tests?

## Related PRs or issues

https://github.com/Microsoft/AppCenter-SDK-React-Native/issues/568/
[AB#59615](https://msmobilecenter.visualstudio.com/web/wi.aspx?pcguid=27bf89de-9697-418a-accf-1aa125b22914&id=59615)

### DIFF
--- a/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
+++ b/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
@@ -776,7 +776,7 @@ public class Crashes extends AbstractAppCenterService {
                 if (file.length() > 0) {
                     try {
                         throwable = FileManager.readObject(file);
-                    } catch (java.lang.Exception | StackOverflowError e) {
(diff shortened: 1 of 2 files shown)


sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -776,7 +776,7 @@ ErrorReport buildErrorReport(ManagedErrorLog log) {
                 if (file.length() > 0) {
                     try {
                         throwable = FileManager.readObject(file);
-                    } catch (IOException | ClassNotFoundException | StackOverflowError e) {
+                    } catch (java.lang.Exception | StackOverflowError e) {
```
guperrot:
```suggestion
                    } catch (IOException | RuntimeException | StackOverflowError e) {
```

The only checked exception is IOException so that means unchecked exceptions can only be RuntimeException or a sub-class.

PR #1051, opened 2019-04-04
Handle multiple tables in DatabaseManager (creation and deletion)

nikitag-ms:
* [x] Did you add unit tests?

## Description

Add `create` and `drop` table operations to `DatabaseManager`

### DIFF
--- a/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/utils/storage/DatabaseManagerAndroidTest.java
+++ b/sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/utils/storage/DatabaseManagerAndroidTest.java
@@ -83,6 +83,7 @@ public class DatabaseManagerAndroidTest {
         sContext.deleteDatabase("test-setMaximumSize");
         sContext.deleteDatabase("test-replace");
         sContext.deleteDatabase("test-replace-by-multiple-columns");
+        sContext.deleteDatabase("test-multiple-tables");
     }
 
     @SuppressWarnings("TryFinallyCanBeTryWithResources")
@@ -488,5 +489,41 @@ public class DatabaseManagerAndroidTest {
             databaseManager.close();
         }
         verify(listener).onCreate(any(SQLiteDatabase.class));
+    }
+
+    @Test
+    public void testMultipleTables() {
+        final String firstTable = "firstTable";
+        final String secondTable = "secondTable";
+
+        /* Get instance to access database. */
+        DatabaseManager.Listener listener = mock(DatabaseManager.Listener.class);
+        DatabaseManager databaseManager = new DatabaseManager(sContext, "test-multiple-tables", firstTable, 1, mSchema, listener);
+        assertTrue(isTableExists(databaseManager.getDatabase(), firstTable));
+        assertFalse(isTableExists(databaseManager.getDatabase(), secondTable));
+
+        databaseManager.createTable(secondTable);
+        assertTrue(isTableExists(databaseManager.getDatabase(), firstTable));
+        assertTrue(isTableExists(databaseManager.getDatabase(), secondTable));
+
+        databaseManager.dropTable(firstTable);
+        assertFalse(isTableExists(databaseManager.getDatabase(), firstTable));
+        assertTrue(isTableExists(databaseManager.getDatabase(), secondTable));
+
+        databaseManager.dropTable(secondTable);
+        assertFalse(isTableExists(databaseManager.getDatabase(), firstTable));
+        assertFalse(isTableExists(databaseManager.getDatabase(), secondTable));
+    }
+
+    private boolean isTableExists(SQLiteDatabase db, String tableName) {
+        Cursor cursor = db.rawQuery("select DISTINCT tbl_name from sqlite_master where tbl_name = '" + tableName + "'", null);
+        if(cursor!=null) {
+            if(cursor.getCount()>0) {
+                cursor.close();
+                return true;
+            }
+            cursor.close();
+        }
+        return false;
     }
 }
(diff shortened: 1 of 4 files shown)


sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/utils/storage/DatabaseManagerAndroidTest.java
```diff
@@ -489,4 +490,40 @@ public void replaceByMultipleColumns() {
...
+        assertFalse(isTableExists(databaseManager.getDatabase(), firstTable));
+        assertTrue(isTableExists(databaseManager.getDatabase(), secondTable));
+
+        databaseManager.dropTable(secondTable);
+        assertFalse(isTableExists(databaseManager.getDatabase(), firstTable));
+        assertFalse(isTableExists(databaseManager.getDatabase(), secondTable));
+    }
+
+    private boolean isTableExists(SQLiteDatabase db, String tableName) {
+        Cursor cursor = db.rawQuery("select DISTINCT tbl_name from sqlite_master where tbl_name = '" + tableName + "'", null);
```
guperrot: Would be better to just put a value and read it. With different schemas.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -514,6 +512,28 @@ void setSQLiteOpenHelper(@NonNull SQLiteOpenHelper helper) {
         mSQLiteOpenHelper = helper;
     }
 
+    private void createTable(SQLiteDatabase db, String table) {
+        /* Generate a schema from specimen. */
```
guperrot:
```suggestion
        
        /* Generate a schema from specimen. */
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -170,6 +151,23 @@ private static ContentValues buildValues(Cursor cursor, ContentValues schema) {
         return values;
     }
 
+    /**
+     * Creates a new table in the database.
+     * @param table name.
+     */
+    public void createTable(String table) {
```
guperrot: Use schema as parameter here.

guperrot:
```suggestion
    public void createTable(@NonNull String table) {
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -170,6 +151,23 @@ private static ContentValues buildValues(Cursor cursor, ContentValues schema) {
...
+        createTable(getDatabase(), table);
+    }
+
+    /**
+     * Deletes a table in the database.
+     * @param table name.
+     */
+    public void dropTable(String table) {
```
guperrot:
```suggestion
    public void dropTable(@NonNull String table) {
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -170,6 +151,23 @@ private static ContentValues buildValues(Cursor cursor, ContentValues schema) {
...
+        SQLiteDatabase db = getDatabase();
+        db.execSQL(String.format("DROP TABLE %s", table));
```
guperrot:
```suggestion
        db.execSQL(String.format("DROP TABLE `%s`", table));
```

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/utils/storage/DatabaseManagerAndroidTest.java
```diff
@@ -489,4 +490,40 @@ public void replaceByMultipleColumns() {
         }
         verify(listener).onCreate(any(SQLiteDatabase.class));
     }
+
+    @Test
+    public void testMultipleTables() {
+        final String firstTable = "firstTable";
```
guperrot:
```suggestion
       String firstTable = "firstTable";
```

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/utils/storage/DatabaseManagerAndroidTest.java
```diff
@@ -489,4 +490,40 @@ public void replaceByMultipleColumns() {
...
+        final String secondTable = "secondTable";
```
guperrot:
```suggestion
       String secondTable = "secondTable";
```

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/utils/storage/DatabaseManagerAndroidTest.java
nikitag-ms: Sure, I'll add write-read tests as well, but I still think it's valuable to check that there are indeed tables underneath.

nikitag-ms: Still working on that new test. For some reason, the write to the second table doesn't actually add a record in there (though no exception and I checked in the DB browser, it's actually not there).

nikitag-ms: Ok, that was a typo. Fixed it.

guperrot: If put/read works then the table has to exist. We can remove this code which is basically using unsecure query that can be picked up by a code scanner.

PR #1052, opened 2019-04-05
Fix the scenario when user first time use the storage sdk and network…

YiMysty:
… is off.

Fix tiny test case

* [x] Did you add unit tests?

## Description

## Related PRs or issues

## Misc


YiMysty:
Handle the case when user first time uses the storage SDK and the network switched to offline.
In this case, since we cannot get the auth token information (because never talked to token-exchange service before, and cannot verify the validation of the user.), we do not do anything in local storage.

Also little fix to some test case and a little clean

PR #1055, opened 2019-04-08
Send push install log again if userId changes

MatkovIvan:
* [x] Did you add unit tests?

## Description

We must re-register the device after the developer sets the User ID (via SDK) so the device ID gets associated with the user ID internally.

## Related PRs or issues

[AB#59454](https://msmobilecenter.visualstudio.com/web/wi.aspx?pcguid=27bf89de-9697-418a-accf-1aa125b22914&id=59454)


sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -313,12 +320,24 @@ public synchronized void onStarted(@NonNull Context context, @NonNull Channel ch
         mAuthListener = new AbstractTokenContextListener() {
 
             @Override
+            // TODO check synchronized
+            // TODO post to background thread?
```
guperrot: The listener is already called from background thread in practice.
mChannel is non null after start so no need for synchronized.

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -313,12 +320,24 @@ public synchronized void onStarted(@NonNull Context context, @NonNull Channel ch
...
                 if (mLatestPushToken != null) {
                     enqueuePushInstallationLog(mLatestPushToken);
                 }
             }
         };
+        mUserListener = new UserIdContext.Listener() {
+
+            @Override
+            public void onNewUserId(String userId) {
+                // TODO use userId
```
guperrot: Probably pass it as param to enqueue to avoid querying context again (could have race condition).

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/UserIdContext.java
```diff
@@ -126,5 +155,20 @@ public synchronized String getUserId() {
...
+        /* Call listeners so that they can react on new token. */
+        for (Listener listener : mListeners) {
+            listener.onNewUserId(mUserId);
+        }
+    }
+
+    public interface Listener {
+
+        /**
+         *
```
guperrot: // TODO

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/UserIdContext.java
```diff
@@ -153,21 +148,35 @@ public synchronized String getUserId() {
...
         }
     }
 
+    /**
+     * Check user identifier.
+     *
+     * @param userId user identifier.
+     * @return true if equals.
+     */
+    private synchronized boolean checkEqualsUserId(String userId) {
```
MatkovIvan: Reconsider the name

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/UserIdContext.java
```diff
@@ -153,21 +148,35 @@ public synchronized String getUserId() {
...
+        if (mUserId == userId && userId != null) {
```
MatkovIvan: Strings should be compared via `equals`.
Why `!= null`?

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/UserIdContext.java
```diff
@@ -153,21 +148,35 @@ public synchronized String getUserId() {
...
+        mUserId = userId;
+        return false;
+    }
+
     public interface Listener {
 
         /**
-         *
+         * Send new user id.
          */
```
MatkovIvan: Add `@param`

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/UserIdContext.java
Jamminroot: `equals`, too? I don't think we need this one

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -241,12 +247,37 @@ private synchronized void setFirebaseAnalyticsEnabled(@NonNull Context context,
      * Enqueue a push installation log.
      *
      * @param pushToken the push token value
+     * @param userId the user identifier value
```
MatkovIvan: Dot at the end

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -277,9 +308,11 @@ public void run() {
     protected synchronized void applyEnabledState(boolean enabled) {
         if (enabled) {
             AuthTokenContext.getInstance().addListener(mAuthListener);
+            UserIdContext.getInstance().addListener(mUserListener);
             registerPushToken();
         } else {
             AuthTokenContext.getInstance().removeListener(mAuthListener);
+            UserIdContext.getInstance().removeListener(mUserListener);
```
MatkovIvan: Please write a test for it

sdk/appcenter-push/src/test/java/com/microsoft/appcenter/push/PushTest.java
```diff
@@ -761,6 +763,45 @@ public void verifyEnqueueCalledOnNewAuthToken() {
         verify(channel, times(2)).enqueue(any(Log.class), anyString(), anyInt());
     }
 
+    @Test
+    public void verifyEnqueueCalledOnNewUserId() {
+        String mockUserId1 = "userId1";
+        String mockUserId2 = "userId2";
+        Push push = Push.getInstance();
+        Channel channel = mock(Channel.class);
+        doNothing().when(channel).enqueue(any(Log.class), anyString(), anyInt());
```
guperrot: Unnecessary since its not a spy.

sdk/appcenter-push/src/test/java/com/microsoft/appcenter/push/PushTest.java
```diff
@@ -761,6 +763,45 @@ public void verifyEnqueueCalledOnNewAuthToken() {
...
+        start(push, channel);
+        push.onTokenRefresh("push-token");
+        ArgumentCaptor<PushInstallationLog> log = ArgumentCaptor.forClass(PushInstallationLog.class);
+        UserIdContext.getInstance().setUserId(mockUserId1);
+        verify(channel, times(2)).enqueue(log.capture(), anyString(), anyInt());
```
guperrot: we need to put the first verify at the right place otherwise we don't know why we have the second call.

sdk/appcenter-push/src/test/java/com/microsoft/appcenter/push/PushTest.java
```diff
@@ -761,6 +763,45 @@ public void verifyEnqueueCalledOnNewAuthToken() {
...
+        assertNull(log.getValue().getUserId());
+    }
+
+    @Test
+    public void verifyEnqueueNotCalledWhenTokenNull() {
+        String mockUserId = "userId";
+        Push push = Push.getInstance();
+        push.onTokenRefresh("push-token");
+        Channel channel = mock(Channel.class);
+        doNothing().when(channel).enqueue(any(Log.class), anyString(), anyInt());
```
guperrot: Should not be necessary

sdk/appcenter-push/src/test/java/com/microsoft/appcenter/push/PushTest.java
```diff
@@ -824,9 +868,44 @@ public void verifyEnqueueNotCalledOnPushDisabled() {
         String mockToken = UUIDUtils.randomUUID().toString();
         String mockHomeId = UUIDUtils.randomUUID().toString();
         AuthTokenContext.getInstance().setAuthToken(mockToken, mockHomeId, mock(Date.class));
+        UserIdContext.getInstance().setUserId(mockUserId);
         verify(channel, never()).enqueue(any(Log.class), anyString(), anyInt());
     }
 
+    @Test
+    public void verifyCalledListenerAfterStart() {
+        ArgumentCaptor<UserIdContext.Listener> userIdContextArgument = ArgumentCaptor.forClass(UserIdContext.Listener.class);
```
guperrot: We need comments to the tests.

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -240,13 +246,38 @@ private synchronized void setFirebaseAnalyticsEnabled(@NonNull Context context,
...
+    }
+
+    /**
+     * Enqueue a push installation log.
+     *
+     * @param pushToken the push token value.
+     */
+    @WorkerThread
     private void enqueuePushInstallationLog(@NonNull String pushToken) {
+        PushInstallationLog log = buildPushInstallationLog(pushToken, UserIdContext.getInstance().getUserId());
```
guperrot: this is duplicate code, we can share the code another way.

sdk/appcenter-push/src/test/java/com/microsoft/appcenter/push/PushTest.java
```diff
@@ -761,11 +761,61 @@ public void verifyEnqueueCalledOnNewAuthToken() {
...
+
+        /* When we remove userId, we re-send log without userId. */
+        UserIdContext.getInstance().setUserId(null);
+        verify(channel, times(4)).enqueue(log.capture(), anyString(), anyInt());
+        assertNull(log.getValue().getUserId());
+    }
+
+    @Test
+    public void verifyEnqueueNotCalledWhenTokenNull() {
+        String mockUserId = "userId";
```
guperrot: Test not being real enough. We should test what happens if et userId while token is null which is better semantics.

PR #1080, opened 2019-04-15
Add the pending operation when activity resume

YiMysty:

* [x] Did you add unit tests?

## Description

Add the pending operation when activity resume

## Related PRs or issues

## Misc

### DIFF
--- a/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
+++ b/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
@@ -6,6 +6,7 @@
 package com.microsoft.appcenter.storage;
 
 import android.annotation.SuppressLint;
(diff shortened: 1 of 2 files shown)


sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
```diff
@@ -281,21 +282,32 @@ public void run() {
...
-                    }
+                    processPendingOperations();
                 }
             }
         });
     }
 
+    @Override
+    public void onActivityResumed(Activity activity) {
+        if (mNetworkStateHelper.isNetworkConnected()) {
```
guperrot: Move to applyEnabledState(true) to do it on only start as this tries to sync operations every time you navigate to a page in the app.

Also we need to manage duplicate http calls between that and network listener.

PR #1086, opened 2019-04-17
Fix conventions for object naming / use gson for expiry date parsing

guperrot:

* [ ] Did you add unit tests?

## Description

* Fix conventions for object naming.
* Just use Gson for date parsing in token.

### DIFF
--- a/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/LocalDocumentStorage.java
+++ b/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/LocalDocumentStorage.java
@@ -131,6 +131,18 @@ class LocalDocumentStorage {
         write(table, document, writeOptions, null);
     }
 
+    private static ContentValues getContentValues(PendingOperation operation, long now) {
+        return getContentValues(
+                operation.getPartition(),
+                operation.getDocumentId(),
+                operation.getDocument(),
+                operation.getETag(),
+                operation.getExpirationTime(),
+                now,
+                now,
+                operation.getOperation());
+    }
+
     private <T> long write(String table, Document<T> document, WriteOptions writeOptions, String pendingOperationValue) {
         if (writeOptions.getDeviceTimeToLive() == WriteOptions.NO_CACHE) {
             return 0;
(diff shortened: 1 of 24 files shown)


sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
```diff
@@ -81,7 +81,7 @@
 
     private Map<DefaultAppCenterFuture<?>, ServiceCall> mPendingCalls = new HashMap<>();
 
-    private final Map<String, ServiceCall> mOutgoingPendingOperationCalls = new HashMap<>();
+    private final HashMap<String, ServiceCall> mOutgoingPendingOperationCalls = new HashMap<>();
```
guperrot: should not put null value on Map as its implementation dependent so we get a lint warning. By telling we depend on this implementation we can depend on ability to set null values

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/LocalDocumentStorage.java
```diff
@@ -338,15 +306,47 @@ void updatePendingOperation(PendingOperation operation) {
...
-        return getContentValues(
-                operation.getPartition(),
-                operation.getDocumentId(),
-                operation.getDocument(),
-                operation.getEtag(),
-                operation.getExpirationTime(),
-                now,
-                now,
-                operation.getOperation());
+    @NonNull
```
guperrot: Someone didn't run the formatter, the formatter can reorder methods based on static or visiblity

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
dhei: +1.

Plus `mOutgoingPendingOperationCalls` is only a private property so it's not leaking anything and there's no point to use `Map` here.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/LocalDocumentStorage.java
```diff
@@ -141,7 +153,7 @@ void resetDatabase() {
                 document.getPartition(),
                 document.getId(),
                 document.toString(),
-                document.getEtag(),
+                document.getETag(),
```
yochaude: I updated the specs in the wiki to match the casing.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/Storage.java
guperrot: That was my feedback to use Map as this is the pattern to use interface whenever possible. This is an exception.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/models/Page.java
```diff
@@ -16,18 +16,18 @@
...
      */
     @SerializedName(value = Constants.DOCUMENTS_FIELD_NAME)
-    private List<Document<T>> documents;
+    private List<Document<T>> mItems;
 
     /**
      * Document error.
      */
-    private DocumentError error;
+    private transient DocumentError mError;
```
dhei: Good catch 👍

PR #1087, opened 2019-04-17
Refresh token after coming back online

MatkovIvan:
* [x] Did you add unit tests?

## Description

When no network and we try to silent sign in (which happens automatically on refresh token), it fails and we are signed out as a result, and offline storage stops working altogether (everything is cleared).

If we need to silent sign in for an automatic SDK behavior, we need to delay it until network comes back online and then submit the call to msal.

## Related PRs or issues

[AB#60235](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/60235)


sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -237,13 +246,22 @@ protected synchronized void applyEnabledState(boolean enabled) {
...
             AuthTokenContext.getInstance().removeListener(mAuthTokenContextListener);
+            NetworkStateHelper.getSharedInstance(mContext).removeListener(this);
             if (mGetConfigCall != null) {
                 mGetConfigCall.cancel();
                 mGetConfigCall = null;
             }
             mAuthenticationClient = null;
             mIdentityScope = null;
-            completeSignIn(null, new IllegalStateException("Identity is disabled."));
+            if (mLastSignInFuture != null && !mLastSignInFuture.isDone()) {
```
guperrot: Please share code here, something like:
```
if (completeOnDisabled(mLastSignInFuture)) {
   mLastSignInFuture = null;
}
if (completeOnDisabled(mLastRefreshFuture)) {
   mLastRefreshFuture = null;
}
```

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -492,89 +518,114 @@ private IAccount retrieveAccount(String id) {
...
+            return;
+        }
+        if (!networkConnected) {
+            mHomeAccountIdToRefresh = homeAccountId;
+            AppCenterLog.debug(LOG_TAG, "Network not connected. The token will be refreshed after coming back online.");
+            return;
+        }
+        final DefaultAppCenterFuture<SignInResult> future = new DefaultAppCenterFuture<>();
+        mLastRefreshFuture = future;
+        IAccount account = retrieveAccount(homeAccountId);
```
guperrot: Before lollipop, onNetworkState is from UI thread.
This method accesses disk.
Need to post() from onNetworkStateUpdated and review synchronized again.

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -429,23 +469,26 @@ private void clearCache() {
...
         final DefaultAppCenterFuture<SignInResult> future = new DefaultAppCenterFuture<>();
-        if (mPendingSignInFuture != null) {
-            future.complete(new SignInResult(null, new IllegalStateException("signIn already in progress.")));
+        if (isFutureInProgress(mLastSignInFuture)) {
+            future.complete(new SignInResult(null, new IllegalStateException("Sign-in already in progress.")));
             return future;
         }
-        mPendingSignInFuture = future;
+        if (isFutureInProgress(mLastRefreshFuture)) {
+            mLastRefreshFuture.complete(new SignInResult(null, new CancellationException()));
```
guperrot: That should be something like
```suggestion
            mLastRefreshFuture.complete(new SignInResult(null, new IllegalStateException("Silent sign-in already in progress.")));
```
And come to think of it, interactive UI login should take precedence. I think we should discuss that later (just please accept suggestion if it makes sense).

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -492,93 +537,122 @@ private IAccount retrieveAccount(String id) {
...
 
-    private void handleSignInSuccess(final IAuthenticationResult authenticationResult) {
+    @WorkerThread
+    private synchronized void refreshToken(String homeAccountId, boolean networkConnected) {
+        if (isFutureInProgress(mLastSignInFuture)) {
+            AppCenterLog.debug(LOG_TAG, "Failed to refresh token: sign-in already in progress.");
+            return;
+        }
+        if (isFutureInProgress(mLastRefreshFuture)) {
+            AppCenterLog.verbose(LOG_TAG, "Token refresh already in progress. Skip this refresh request.");
```
guperrot:
```suggestion
            AppCenterLog.debug(LOG_TAG, "Token refresh already in progress. Skip this refresh request.");
```

MatkovIvan:
I've marked it as ready for review but a few lines still need to be covered

sdk/appcenter-identity/src/main/java/com/microsoft/appcenter/identity/Identity.java
```diff
@@ -492,93 +537,122 @@ private IAccount retrieveAccount(String id) {
...
+            AuthTokenContext.getInstance().setAuthToken(null, null, null);
+        }
+    }
+
+    private void handleSignInSuccess(@NonNull final DefaultAppCenterFuture<SignInResult> future, final IAuthenticationResult authenticationResult) {
         post(new Runnable() {
 
             @Override
             public void run() {
+                if (future.isDone()) {
```
guperrot: Comment why we need that

sdk/appcenter-identity/src/test/java/com/microsoft/appcenter/identity/IdentityTest.java
```diff
@@ -276,6 +277,63 @@ public void signInThenDownloadValidConfiguration() throws Exception {
...
+
+        /* Start identity service. */
+        Identity identity = Identity.getInstance();
+        start(identity);
+
+        /* Mock foreground then background again. */
+        identity.onActivityResumed(mock(Activity.class));
+
+        /* Sign in, will be delayed until configuration ready. */
+        Identity.signIn();
```
guperrot: We don't have delayed sign in anymore, comment is probably wrong.

guperrot: Screen Shot 2019-04-19 at 11 28 34 AM

Thus this test setup seems unreal to me, can you fix the semantics of the test?

guperrot: Fixing some tests right now that were not updated after the error on background/not configured change.

PR #1090, opened 2019-04-17
Adding uriEncoding for CosmosDB calls.

hovsepm:
Following scenario test fails:
1. Create a document with document id that contains spaces (e.g. "Test Document").
2. Retrieve all documents (Storage.list) - make sure the new document is present and its id is "Test Document"
3. Call Storage.read with that document id results in 400 error.

The issue is that we construct DocumentDB Uri without escaping strings that can contain unencoded URI values (e.g. spaces or slashes). 

* [ ] Did you add unit tests?

## Description

## Related PRs or issues

## Misc

### DIFF
--- a/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/client/CosmosDb.java
+++ b/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/client/CosmosDb.java
@@ -92,9 +92,9 @@ public class CosmosDb {
 
 
     public static String getDocumentBaseUrl(String databaseName, String collectionName, String documentId) {
-        return String.format(DOCUMENT_DB_DATABASE_URL_SUFFIX, databaseName) + "/" +
-                String.format(DOCUMENT_DB_COLLECTION_URL_SUFFIX, collectionName) + "/" +
-                DOCUMENT_DB_DOCUMENT_URL_PREFIX + (documentId == null ? "" : '/' + documentId);
+        return String.format(DOCUMENT_DB_DATABASE_URL_SUFFIX, urlEncode(databaseName)) + "/" +
+                String.format(DOCUMENT_DB_COLLECTION_URL_SUFFIX, urlEncode(collectionName)) + "/" +
+                DOCUMENT_DB_DOCUMENT_URL_PREFIX + (documentId == null ? "" : '/' + urlEncode(documentId));
     }
 
     private static String getDocumentUrl(TokenResult tokenResult, String documentId) {
--- a/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/client/CosmosDb.java
+++ b/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/client/CosmosDb.java
@@ -62,7 +62,10 @@ public class CosmosDb {
     }
 
     private static String urlEncode(String url) {
-        return urlEncode(url, "UTF-8");
+        if(url != null) {
+            return urlEncode(url, "UTF-8");
+        }
+        return "null";
     }
 
     public static String urlEncode(String url, String enc) {
--- a/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/client/CosmosDb.java
+++ b/sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/client/CosmosDb.java
@@ -5,12 +5,15 @@
 
 package com.microsoft.appcenter.storage.client;
 
+import android.support.annotation.VisibleForTesting;
+
 import com.microsoft.appcenter.http.HttpClient;
 import com.microsoft.appcenter.http.ServiceCall;
 import com.microsoft.appcenter.http.ServiceCallback;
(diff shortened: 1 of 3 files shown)


YiMysty:
a little question, why do we not just urlencode the whole url directly?

hovsepm:
> a little question, why do we not just urlencode the whole url directly?

@YiMysty  because the entire Uri contains slashes `/` but we want to encode slashes in the parts of uri such as documentId. This code is a mirror reflection of the official CosmosDB C# client that I extracted from their nuget package library.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/client/CosmosDb.java
```diff
@@ -62,7 +62,10 @@ private static String nowAsRFC1123() {
     }
 
     private static String urlEncode(String url) {
-        return urlEncode(url, "UTF-8");
+        if(url != null) {
```
guperrot: The formatter was not run.

sdk/appcenter-storage/src/test/java/com/microsoft/appcenter/storage/StorageTest.java
```diff
@@ -553,6 +555,32 @@ public void readFailedCosmosDbCallFailed() throws JSONException {
...
     }
 
+    @Test
+    public void readCosmosDbCallEncodeDocumentId() throws JSONException, UnsupportedEncodingException {
+        String documentID = "Test Document";
+        AppCenterFuture<Document<TestDocument>> doc = Storage.read(USER, documentID, TestDocument.class);
+
+        verifyTokenExchangeFlow(TOKEN_EXCHANGE_USER_PAYLOAD, null);
+
+        // verify that document base Uri is properly constructed by CosmosDb.getDocumentBaseUrl method
```
guperrot: We use `/*` for comments in Android

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/client/CosmosDb.java
```diff
@@ -62,7 +62,10 @@ private static String nowAsRFC1123() {
...
+            return urlEncode(url, "UTF-8");
+        }
+        return "null";
```
guperrot: What is that return value? What do we do with it? It looks to me we need to return a document error instead if we get an error for url encoding.

sdk/appcenter-storage/src/test/java/com/microsoft/appcenter/storage/StorageTest.java
```diff
@@ -553,6 +555,32 @@ public void readFailedCosmosDbCallFailed() throws JSONException {
                 CoreMatchers.containsString("Cosmos db exception."));
     }
 
+    @Test
+    public void readCosmosDbCallEncodeDocumentId() throws JSONException, UnsupportedEncodingException {
+        String documentID = "Test Document";
+        AppCenterFuture<Document<TestDocument>> doc = Storage.read(USER, documentID, TestDocument.class);
```
guperrot: It's not allowed to commit with Android Studio warnings. Normally we use Android Studio commit dialog to prevent that.

sdk/appcenter-storage/src/test/java/com/microsoft/appcenter/storage/StorageTest.java
```diff
@@ -553,6 +555,32 @@ public void readFailedCosmosDbCallFailed() throws JSONException {
...
+        // verify that document base Uri is properly constructed by CosmosDb.getDocumentBaseUrl method
+        String expectedUri = String.format("dbs/%s", DATABASE_NAME) + "/" +
+                String.format("colls/%s", COLLECTION_NAME) + "/" +
+                "docs" + '/' + URLEncoder.encode(documentID, "UTF-8");
+        assertEquals(expectedUri, CosmosDb.getDocumentBaseUrl(DATABASE_NAME, COLLECTION_NAME, documentID));
+
+        // now verify that actual call was properly encoded
+        ArgumentCaptor<HttpClient.CallTemplate> cosmosDbCallTemplateCallbackArgumentCaptor =
+                ArgumentCaptor.forClass(HttpClient.CallTemplate.class);
+        ArgumentCaptor<ServiceCallback> cosmosDbServiceCallbackArgumentCaptor =
```
guperrot: Argument captured and nothing is done on it.

guperrot:
@hovsepm addressed feedbacks myself.
Please take some time to look at the fixes I pushed and configure commit dialog with

Screen Shot 2019-04-17 at 7 42 26 PM

Before merging.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/client/CosmosDb.java
guperrot: Ok that's because we mock token without fields that should be required such as db name or collection name and then url encoding on it. Added a TODO as we should have a bug to fix the root cause.

PR #1091, opened 2019-04-17
Fix multiple bugs on pending operation success and failure scenarios

dhei:

* [x] Did you add unit tests?

## Description

Adds 2 properties to `PendingOperation` model to map operations columns to save one extra database read when updating the operation in cache.
- `download_time`
- `operations_time`

Fixes multiple issues for processing pending offline operations when network is back:

1. If pending operation has been processed successfully, we don't set the `pending_operation` column of the operation to `null` in cache. This PR fixes it.
2. If we get 500 errors from cosmos Db on processing pending operation, we should leave the operation in cache (for retry later) rather than delete it. This PR fixes it.
3. Some unit tests mistakenly use 24 hours time interval as expiration timestamp, causing false positive on those tests.

https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/60065


sdk/appcenter-storage/src/test/java/com/microsoft/appcenter/storage/AbstractStorageTest.java
```diff
@@ -91,6 +92,12 @@
 
     static final String ETAG = "06000da6-0000-0000-0000-5c7093c30000";
 
+    static final long TIMESTAMP_TODAY = System.currentTimeMillis();
+
+    static final long TIMESTAMP_TOMORROW = System.currentTimeMillis() + BaseOptions.DEFAULT_EXPIRATION_IN_SECONDS;
```
guperrot: If it was SDK I would tell to use Calendar for adding or removing days, but given it's a unit test it's ok to be inaccurate for that purpose. Just afraid it will encourage that in other places. Maybe rename to past and future?

dhei: Good point, fixed.

sdk/appcenter-storage/src/main/java/com/microsoft/appcenter/storage/LocalDocumentStorage.java
```diff
@@ -244,23 +232,16 @@ boolean deleteOnline(String table, String partition, String documentId) {
                         values.getAsString(PARTITION_COLUMN_NAME),
                         values.getAsString(DOCUMENT_ID_COLUMN_NAME),
                         values.getAsString(DOCUMENT_COLUMN_NAME),
-                        values.getAsLong(EXPIRATION_TIME_COLUMN_NAME)));
+                        values.getAsLong(EXPIRATION_TIME_COLUMN_NAME),
+                        values.getAsLong(DOWNLOAD_TIME_COLUMN_NAME),
```
guperrot: I can't find where it is used for the initial put.

guperrot: Also it's read but never used.

dhei: The sqlite database schema contains `download_time` and `operations_time`:
```
INTEGER oid (primary key auto increment)
   TEXT partition
   TEXT document_id
   TEXT document
   TEXT etag
INTEGER expiration_time
INTEGER download_time
INTEGER operation_time
   TEXT pending_operation (NULL, CREATE, REPLACE, DELETE)
```

The current `PendingOperation` model contains all the columns except `download_time` and `operations_time` columns. As part of this PR, we need to call [`mLocalDocumentStorage.updatePendingOperation(pendingOperation)`](https://github.com/Microsoft/AppCenter-SDK-Android/pull/1091/files#diff-9a1784d794428d5c4bddb2690a20dcb5R906) which internally calls `mDatabaseManager.replace` API. That is why we need the 1:1 mapping of database columns to `PendingOperation` properties to be able to replace an row in database. Without such mapping, I believe it would set `download_time` and `operations_time` columns to NULL.

PR #1120, opened 2019-04-29
Sign in always fails on first launch fixed

russelarms:

* [x] Did you add unit tests?

## Description

The first sign in always fails for a fresh Android app because the config is not downloaded in time and the app thinks that configuration failed.

Steps to reproduce:
1)Create a new Android app and integrate auth
2)Sign in (you will need to close the app and reopen to get sign in to actually happen)

Expected behavior:
1)Auth sign in method waits asynchronously for the HTTP response to get the remote config if the network is up and if there is no cached config.
2)If the HTTP call fails the completion handler of the sign in API is called with SignInResult with an exception.
3)Sign in immediately fails if no network and if the application is in background. (This matches the existing behavior)

## Related PRs or issues

[AB#60930](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/60930)

## Misc

### DIFF
--- a/sdk/appcenter-auth/src/main/java/com/microsoft/appcenter/auth/Auth.java
+++ b/sdk/appcenter-auth/src/main/java/com/microsoft/appcenter/auth/Auth.java
@@ -380,12 +380,11 @@ public class Auth extends AbstractAppCenterService implements NetworkStateHelper
 
     @WorkerThread
     private synchronized void processDownloadedConfig(String payload, String eTag) {
-        boolean continueSignIn = mAuthenticationClient == null && mLastSignInFuture != null;
         mGetConfigCall = null;
         saveConfigFile(payload, eTag);
         AppCenterLog.info(LOG_TAG, "Configure auth from downloaded configuration.");
         initAuthenticationClient(payload);
-        if (continueSignIn) {
+        if (isFirstSignInPending()) {
             selectSignInTypeAndSignIn(mLastSignInFuture);
         }
     }
@@ -393,7 +392,7 @@ public class Auth extends AbstractAppCenterService implements NetworkStateHelper
     private synchronized void processDownloadNotModified() {
         mGetConfigCall = null;
         AppCenterLog.info(LOG_TAG, "Auth configuration didn't change.");
-        if (mAuthenticationClient == null && mLastSignInFuture != null) {
+        if (isFirstSignInPending()) {
             mLastSignInFuture.complete(new SignInResult(null, new IllegalStateException("Cannot load auth configuration from the server.")));
         }
     }
@@ -410,7 +409,7 @@ public class Auth extends AbstractAppCenterService implements NetworkStateHelper
     private synchronized void processDownloadError(Exception e) {
         mGetConfigCall = null;
         AppCenterLog.error(LOG_TAG, "Cannot load auth configuration from the server.", e);
-        if (mAuthenticationClient == null && mLastSignInFuture != null) {
+        if (isFirstSignInPending()) {
             mLastSignInFuture.complete(new SignInResult(null, new IllegalStateException("Cannot load auth configuration from the server.")));
         }
     }
@@ -727,6 +726,10 @@ public class Auth extends AbstractAppCenterService implements NetworkStateHelper
                 future.complete(new SignInResult(null, new CancellationException("User cancelled sign-in.")));
             }
         });
+    }
+
+    private boolean isFirstSignInPending() {
+        return mAuthenticationClient == null && mLastSignInFuture != null;
     }
 
     private boolean isFutureInProgress(AppCenterFuture<SignInResult> future) {
--- a/sdk/appcenter-auth/src/main/java/com/microsoft/appcenter/auth/Auth.java
+++ b/sdk/appcenter-auth/src/main/java/com/microsoft/appcenter/auth/Auth.java
@@ -556,7 +556,7 @@ public class Auth extends AbstractAppCenterService implements NetworkStateHelper
 
             /* Check if getting the config in process. */
             if (mGetConfigCall != null) {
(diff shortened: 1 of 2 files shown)


sdk/appcenter-auth/src/main/java/com/microsoft/appcenter/auth/Auth.java
```diff
@@ -714,6 +728,10 @@ public void run() {
         });
     }
 
+    private boolean isFirstSignInPending() {
```
MatkovIvan: Why "first"?

sdk/appcenter-auth/src/main/java/com/microsoft/appcenter/auth/Auth.java
```diff
@@ -384,12 +384,17 @@ private synchronized void processDownloadedConfig(String payload, String eTag) {
         saveConfigFile(payload, eTag);
         AppCenterLog.info(LOG_TAG, "Configure auth from downloaded configuration.");
         initAuthenticationClient(payload);
+        if (isFirstSignInPending()) {
```
MatkovIvan: It's called after `initAuthenticationClient `, so it's always `false`

sdk/appcenter-auth/src/main/java/com/microsoft/appcenter/auth/Auth.java
```diff
@@ -380,20 +380,19 @@ public void run() {
 
     @WorkerThread
     private synchronized void processDownloadedConfig(String payload, String eTag) {
-        boolean continueSignIn = mAuthenticationClient == null && mLastSignInFuture != null;
```
MatkovIvan: Incorrect change. You can't check it after `initAuthenticationClient`. Add comment about that.

sdk/appcenter-auth/src/main/java/com/microsoft/appcenter/auth/Auth.java
```diff
@@ -380,16 +380,22 @@ public void run() {
...
+            selectSignInTypeAndSignIn(mLastSignInFuture);
+        }
     }
 
-
     private synchronized void processDownloadNotModified() {
         mGetConfigCall = null;
         AppCenterLog.info(LOG_TAG, "Auth configuration didn't change.");
+        if (isPendingSignInWaitingForConfiguration()) {
+            mLastSignInFuture.complete(new SignInResult(null, new IllegalStateException("Cannot load auth configuration from the server.")));
```
jwallra: Should this be an `InternalServerErrorException` or similar?  Theoretically this is a problem with the CDN/server providing the config, so `IllegalStateException` doesn't seem quite right to me

sdk/appcenter-auth/src/main/java/com/microsoft/appcenter/auth/Auth.java
```diff
@@ -544,7 +553,13 @@ private synchronized void selectSignInTypeAndSignIn(final DefaultAppCenterFuture
             return;
         }
         if (mAuthenticationClient == null) {
-            future.complete(new SignInResult(null, new IllegalStateException("signIn is called while it's not configured.")));
+
+            /* Check if getting the config in process. */
+            if (mGetConfigCall != null) {
+                AppCenterLog.debug(LOG_TAG, "Downloading configuration in process. Waiting it before sign-in.");
```
jwallra:
```suggestion
                AppCenterLog.debug(LOG_TAG, "Downloading configuration in process. Waiting for it before sign-in.");
```

sdk/appcenter-auth/src/test/java/com/microsoft/appcenter/auth/AuthTest.java
```diff
@@ -1751,6 +1732,121 @@ public void removeAccountWithWrongIdentifier() throws Exception {
...
+        AppCenterFuture<SignInResult> future = Auth.signIn();
+
+        /* Verify that future is not completed, while config is being downloaded. */
+        assertNotNull(future);
+        assertFalse(future.isDone());
+        verify(publicClientApplication, never()).acquireTokenSilentAsync(any(String[].class), notNull(IAccount.class), isNull(String.class), eq(true), signInCallbackCaptor.capture());
+
+        /* Simulate download configuration response. */
+        mockHttpCallSuccess(mockValidForAppCenterConfig(), serviceCallback);
+
```
jwallra:
```suggestion
```

sdk/appcenter-auth/src/main/java/com/microsoft/appcenter/auth/Auth.java
guperrot: This is not an error case, and this cannot really happen to be delayed for that case.
If we ever get 304 that means we send an etag, meaning we had a config in the first place and thus we already allowed signIn.

I guess this is just a defensive check in case the impossible happens @jwallra.

sdk/appcenter-auth/src/main/java/com/microsoft/appcenter/auth/Auth.java
```diff
@@ -544,7 +553,13 @@ private synchronized void selectSignInTypeAndSignIn(final DefaultAppCenterFuture
             return;
         }
         if (mAuthenticationClient == null) {
-            future.complete(new SignInResult(null, new IllegalStateException("signIn is called while it's not configured.")));
+
+            /* Check if getting the config in process. */
+            if (mGetConfigCall != null) {
+                AppCenterLog.debug(LOG_TAG, "Downloading configuration in process. Waiting for it before sign-in.");
```
guperrot:
```suggestion
                AppCenterLog.info(LOG_TAG, "Downloading configuration in process. Waiting for it before sign-in.");
```

PR #1124, opened 2019-04-30
[From Intercom] Crash when stacktrace is null

AnastasiaKubova:

* [x] Did you add unit tests?

## Description

```
java.io.File.<init> File.java:282
com.microsoft.appcenter.crashes.Crashes$9.run Crashes.java:837
com.microsoft.appcenter.AbstractAppCenterService$4.run AbstractAppCenterService.java:301
com.microsoft.appcenter.AppCenter$7.run AppCenter.java:704
android.os.Handler.handleCallback Handler.java:789
```

Should check StackTrace on null.

## Related PRs or issues

based on intercom report
[AB#61633](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/61633)


sdk/appcenter-crashes/src/test/java/com/microsoft/appcenter/crashes/CrashesTest.java
```diff
@@ -1489,6 +1492,47 @@ public void minidumpAppLaunchTimestampFromSessionContextNotFound() throws Except
         assertEquals(new Date(crashTime), crashLog.getAppLaunchTimestamp());
     }
 
+    @Test
+    public void stacktraceNull() throws Exception {
+        long appStartTime = 99L;
+        long crashTime = 123L;
+
```
MatkovIvan:
```suggestion
```

sdk/appcenter-crashes/src/test/java/com/microsoft/appcenter/crashes/CrashesTest.java
```diff
@@ -1489,6 +1492,47 @@ public void minidumpAppLaunchTimestampFromSessionContextNotFound() throws Except
         assertEquals(new Date(crashTime), crashLog.getAppLaunchTimestamp());
     }
 
+    @Test
+    public void stacktraceNull() throws Exception {
```
MatkovIvan: Please add some comments about what's happening here

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -838,10 +838,12 @@ public void run() {
                         ErrorLogReport errorLogReport = unprocessedEntry.getValue();
                         if (errorLogReport.report.getThrowable() instanceof NativeException) {
                             Exception exception = errorLogReport.log.getException();
-                            dumpFile = new File(exception.getStackTrace());
-                            exception.setStackTrace(null);
-                            byte[] logfileContents = FileManager.readBytes(dumpFile);
-                            dumpAttachment = ErrorAttachmentLog.attachmentWithBinary(logfileContents, "minidump.dmp", "application/octet-stream");
+                            if (exception.getStackTrace() != null) {
```
guperrot: That does not explain how stack trace is null and seems like a work around.
What is the cause?

guperrot: Also that means we won't send minidump for that crash and thus it will not show up on portal.

jaelim-ms: This will just hide the issue. We should investigate and find root causes of the issue.

guperrot: If we really cannot find the root cause then yes we could avoid the crash, but since we are not yet releasing the SDK we still have time to find why.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -837,10 +837,12 @@ public void run() {
...
+                                dumpFile = new File(exception.getStackTrace());
+                                exception.setStackTrace(null);
```
guperrot: This is the only place that set it to null.
As in line 583 the API we use is `@NonNull`. 
This suggests some kind of double processing even though we remove the entry in the iterator.

guperrot: If we can't find why, I would rather we remove `exception.setStackTrace(null)` rather than checking for null as that is only here to cleanup the json we send. That way we can check if that is the cause of the null pointer (double processing) and reopen investigation if crash reported again. Just adding the if will hide problems that will be even harder to investigate.

guperrot:
So far, only way I can reproduce is using the internal `NativeException` of the SDK to throw it as a test crash: `throw new NativeException()`. This type is not meant to be used in public, will ask if that was a test.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/ingestion/models/Exception.java
```diff
@@ -43,9 +43,9 @@
     private String message;
 
     /**
-     * Raw stack trace. Sent when the frames property is either missing or unreliable.
+     * The path to the minidump file. Used for reports from the NDK.
      */
-    private String stackTrace;
```
guperrot: You cannot rename stacktrace, it is used for legit Java exceptions.
You can use a new field that is not serialized to JSON instead. We were using stacktrace as a temporary field just to hold file path, so if you want a clean solution you need to restore stacktrace as before and use a separate field that is just in memory.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -836,11 +836,14 @@ public void run() {
                         Map.Entry<UUID, ErrorLogReport> unprocessedEntry = unprocessedIterator.next();
                         ErrorLogReport errorLogReport = unprocessedEntry.getValue();
                         if (errorLogReport.report.getThrowable() instanceof NativeException) {
-                            Exception exception = errorLogReport.log.getException();
-                            dumpFile = new File(exception.getStackTrace());
-                            exception.setStackTrace(null);
```
guperrot: Don't remove setting back to null now that you have the check.
We highly suspect double processing as the customer confirmed it never tried to throw `NativeException`.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/ingestion/models/Exception.java
```diff
@@ -42,6 +42,11 @@
      */
     private String message;
 
+    /**
+     * Raw stack trace. Sent when the frames property is either missing or unreliable.
+     */
```
MatkovIvan: Add note about usage in Xamarin SDK

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/ingestion/models/Exception.java
```diff
@@ -177,6 +200,7 @@ public void read(JSONObject object) throws JSONException {
         setType(object.optString(TYPE, null));
         setMessage(object.optString(MESSAGE, null));
         setMinidumpFilePath(object.optString(STACK_TRACE, null));
```
MatkovIvan: Use different key

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -836,11 +837,16 @@ public void run() {
...
+                            String minidumpFilePath = errorLogReport.log.getException().getMinidumpFilePath();
+
+                            /* It can be null when NativeException is incorrectly used or there is already stored invalid data. */
+                            if (minidumpFilePath != null) {
+                                dumpFile = new File(minidumpFilePath);
+                                byte[] logfileContents = FileManager.readBytes(dumpFile);
+                                dumpAttachment = ErrorAttachmentLog.attachmentWithBinary(logfileContents, "minidump.dmp", "application/octet-stream");
+                            }
+                            errorLogReport.log.getException().setMinidumpFilePath(null);
+                            errorLogReport.log.getException().setStackTrace(null);
```
guperrot: This line is not needed, we are using minidump field now.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -836,11 +837,16 @@ public void run() {
                         Map.Entry<UUID, ErrorLogReport> unprocessedEntry = unprocessedIterator.next();
                         ErrorLogReport errorLogReport = unprocessedEntry.getValue();
                         if (errorLogReport.report.getThrowable() instanceof NativeException) {
-                            Exception exception = errorLogReport.log.getException();
-                            dumpFile = new File(exception.getStackTrace());
-                            exception.setStackTrace(null);
-                            byte[] logfileContents = FileManager.readBytes(dumpFile);
-                            dumpAttachment = ErrorAttachmentLog.attachmentWithBinary(logfileContents, "minidump.dmp", "application/octet-stream");
+                            String minidumpFilePath = errorLogReport.log.getException().getMinidumpFilePath();
```
guperrot: `errorLogReport.log.getException()` is used a lot, can be refactored to a local variable.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/ingestion/models/Exception.java
```diff
@@ -28,6 +28,8 @@
 
     private static final String STACK_TRACE = "stackTrace";
 
+    private static final String MINIDUMP_FILE_PATH = "MinidumpFilePath";
```
guperrot:
```suggestion
    private static final String MINIDUMP_FILE_PATH = "minidumpFilePath";
```

guperrot: Also move it after last field, we order fields consistently for json models.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/ingestion/models/Exception.java
```diff
@@ -176,6 +201,7 @@ public void setWrapperSdkName(String wrapperSdkName) {
     public void read(JSONObject object) throws JSONException {
         setType(object.optString(TYPE, null));
         setMessage(object.optString(MESSAGE, null));
+        setMinidumpFilePath(object.optString(MINIDUMP_FILE_PATH, null));
```
guperrot: move it last

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/ingestion/models/Exception.java
```diff
@@ -186,6 +212,7 @@ public void read(JSONObject object) throws JSONException {
     public void write(JSONStringer writer) throws JSONException {
         JSONUtils.write(writer, TYPE, getType());
         JSONUtils.write(writer, MESSAGE, getMessage());
+        JSONUtils.write(writer, STACK_TRACE, getMinidumpFilePath());
```
guperrot: This using the wrong constant here and should be moved to last line

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/ingestion/models/Exception.java
```diff
@@ -224,6 +254,7 @@ public boolean equals(Object o) {
     public int hashCode() {
         int result = type != null ? type.hashCode() : 0;
         result = 31 * result + (message != null ? message.hashCode() : 0);
+        result = 31 * result + (minidumpFilePath != null ? minidumpFilePath.hashCode() : 0);
```
guperrot: regenerate after moving code

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/ingestion/models/Exception.java
```diff
@@ -208,6 +235,9 @@ public boolean equals(Object o) {
         if (message != null ? !message.equals(exception.message) : exception.message != null) {
             return false;
         }
+        if (minidumpFilePath != null ? !minidumpFilePath.equals(exception.minidumpFilePath) : exception.minidumpFilePath != null) {
```
guperrot: regenerate after moving code

sdk/appcenter-crashes/src/androidTest/java/com/microsoft/appcenter/crashes/ingestion/models/ErrorModelTest.java
```diff
@@ -301,13 +302,16 @@ public void managedErrorLog() throws JSONException {
                 checkEquals(exception1, exception2);
             }
             {
-                exception1.setStackTrace("1");
+                exception1.setMinidumpFilePath("1");
```
guperrot: do not test 2 fields at once, please keep the existing pattern (and test the new field last)

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
achocron: I think it's good to have both since that's an implementation detail and if that changed in the future we could regress

guperrot: It's dead code, stack trace will always be null here.

PR #1125, opened 2019-04-30
Make the postAsyncGetter() call return a proper document error instead of a null

nikitag-ms:

* [ ] Did you add unit tests?

## Description

Make the postAsyncGetter() call return a proper document error instead of a null


nikitag-ms:
> Please test the new result in unit tests.

I couldn't find a way to do it because each call to `postAsyncGetter` is always preceded by `isInvalidStateOrParameters`, which checks whether the module has started or not and returns before that return value can be verified.

guperrot:
> > Please test the new result in unit tests.
> 
> I couldn't find a way to do it because each call to `postAsyncGetter` is always preceded by `isInvalidStateOrParameters`, which checks whether the module has started or not and returns before that return value can be verified.

You can check started bu SDK disabled.

PR #1145, opened 2019-05-07
Add set send events latency api for Analytics 

Zongrenfeng:

* [x] Did you add unit tests?

## Description
Add setTransmissionInterval API that allows normal events send period to be set.


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -74,6 +74,12 @@
     @SuppressLint("StaticFieldLeak")
     private static Analytics sInstance;
 
+    /**
+     * Transmission interval range values
+     */
+    private static final int TRANSMISSIONINTERVAL_MINMAL = 3;
+    private static final int TRANSMISSIONINTERVAL_MAXMIUM = 24*60;
```
Jamminroot: This won't ne enough :)

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -74,6 +74,12 @@
     @SuppressLint("StaticFieldLeak")
     private static Analytics sInstance;
 
+    /**
+     * Transmission interval range values
+     */
+    private static final int TRANSMISSIONINTERVAL_MINMAL = 3;
```
Jamminroot:
```suggestion
    private static final int TRANSMISSIONINTERVAL_MINIMAL = 3;
```

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
```diff
@@ -62,6 +62,18 @@
      */
     private final PropertyConfigurator mPropertyConfigurator;
 
+    /**
+     * Transmission interval range values
+     */
+    private static final int TRANSMISSIONINTERVAL_MINMAL = 3;
```
Jamminroot:
```suggestion
    private static final int TRANSMISSIONINTERVAL_MINIMAL = 3;
```

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
```diff
@@ -62,6 +62,18 @@
...
+    private static final int TRANSMISSIONINTERVAL_MAXMIUM = 24*60;
```
Jamminroot: 24 * 60 * 60

Zongrenfeng: Oh, sorry, I will correct this in soon, thanks,

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/AnalyticsTransmissionTarget.java
```diff
@@ -72,6 +84,26 @@
...
+    /**
+     * set Transmission interval
+     * @param seconds Set latency of sending events to Analytics
+     * @return succeed <code>true</code> if set succeed, <code>false</code> otherwise;
+     */
+    public boolean setTransmissionInterval(int seconds)
+    {
+        if (seconds < TRANSMISSIONINTERVAL_MINMAL || seconds > TRANSMISSIONINTERVAL_MAXMIUM)
+        {
+            AppCenterLog.error(LOG_TAG, "Valid Transmission Interval value should be a int represented by seconds ranger 3 to 1 day (24*60)");
```
Jamminroot:
```suggestion
            AppCenterLog.error(LOG_TAG, "Valid transmission interval value should be between 3 and 24*60*60");
```

Zongrenfeng: updated.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -125,6 +136,12 @@
      */
     private AnalyticsListener mAnalyticsListener;
 
+    /**
+     * Transmission Interval
+     * valid value range from 3 seconds to 1 day
```
MatkovIvan: Dot at the end

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -183,6 +200,21 @@ static synchronized void unsetInstance() {
         return getInstance().setInstanceEnabledAsync(enabled);
     }
 
+    /**
+     * set Transmission interval
```
MatkovIvan:
```suggestion
     * Set transmission interval.
```

MatkovIvan: Add note about requirements and restrictions

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -183,6 +200,21 @@ static synchronized void unsetInstance() {
...
+     *
+     * @param seconds Set latency of sending events to Analytics
```
MatkovIvan: Please format. Add dot at the end

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -183,6 +200,21 @@ static synchronized void unsetInstance() {
...
+     * @return succeed <code>true</code> if set succeed, <code>false</code> otherwise;
+     */
+    public static boolean setTransmissionInterval(int seconds) {
+        if (seconds < TRANSMISSION_INTERVAL_MINIMAL || seconds > TRANSMISSION_INTERVAL_MAXIMUM) {
+            AppCenterLog.error(LOG_TAG, "Valid transmission interval value should be between 3 and 24*60*60");
```
MatkovIvan:
```suggestion
            AppCenterLog.error(LOG_TAG, "The transmission interval is invalid. The value should be between 3 seconds and 1 day");
```

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -183,6 +200,21 @@ static synchronized void unsetInstance() {
...
+            return false;
+        }
+        getInstance().mTransmissionInterval = (int) TimeUnit.SECONDS.toMillis(seconds);
```
MatkovIvan: Please move logic out from static method

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/OneCollectorChannelListener.java
```diff
@@ -109,12 +109,12 @@ public void setLogUrl(@NonNull String logUrl) {
     }
 
     @Override
-    public void onGroupAdded(@NonNull String groupName, Channel.GroupListener groupListener) {
+    public void onGroupAdded(@NonNull String groupName, Channel.GroupListener groupListener, long batchTimeInterval) {
         if (isOneCollectorGroup(groupName)) {
             return;
         }
         String oneCollectorGroupName = getOneCollectorGroupName(groupName);
-        mChannel.addGroup(oneCollectorGroupName, ONE_COLLECTOR_TRIGGER_COUNT, ONE_COLLECTOR_TRIGGER_INTERVAL, ONE_COLLECTOR_TRIGGER_MAX_PARALLEL_REQUESTS, mIngestion, groupListener);
```
MatkovIvan: `ONE_COLLECTOR_TRIGGER_INTERVAL` is unused now

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -74,6 +75,16 @@
     @SuppressLint("StaticFieldLeak")
     private static Analytics sInstance;
 
+    /**
+     * Transmission interval minimal value.
+     */
+    private static final int TRANSMISSION_INTERVAL_MINIMAL = 3;
```
MatkovIvan:
```suggestion
    private static final int TRANSMISSION_INTERVAL_MINIMUM = 3;
```

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -867,6 +899,21 @@ private void setDefaultTransmissionTarget(String transmissionTargetToken) {
...
+     *
+     * @param seconds Set latency of sending events to Analytics.
+     * @return succeed <code>true</code> if set succeed, <code>false</code> otherwise.
+     */
+    private boolean setTriggerInterval(int seconds) {
+        if (seconds < TRANSMISSION_INTERVAL_MINIMAL || seconds > TRANSMISSION_INTERVAL_MAXIMUM) {
+            AppCenterLog.error(LOG_TAG, "The transmission interval is invalid. The value should be between 3 seconds and 1 day");
+            return false;
+        }
+        mTransmissionInterval = (int) TimeUnit.SECONDS.toMillis(seconds);
```
MatkovIvan: It's only valid if called before start. Add check and comment

sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/OneCollectorChannelListenerTest.java
```diff
@@ -63,7 +62,7 @@ public void addCorrespondingGroup() {
         listener.onGroupAdded(TEST_GROUP, groupListener);
 
         /* Verify one collector group added. */
-        verify(channel).addGroup(eq(TEST_GROUP + ONE_COLLECTOR_GROUP_NAME_SUFFIX), eq(ONE_COLLECTOR_TRIGGER_COUNT), eq(ONE_COLLECTOR_TRIGGER_INTERVAL), eq(ONE_COLLECTOR_TRIGGER_MAX_PARALLEL_REQUESTS), argThat(new ArgumentMatcher<Ingestion>() {
+        verify(channel).addGroup(eq(TEST_GROUP + ONE_COLLECTOR_GROUP_NAME_SUFFIX), eq(ONE_COLLECTOR_TRIGGER_COUNT), anyInt(), eq(ONE_COLLECTOR_TRIGGER_MAX_PARALLEL_REQUESTS), argThat(new ArgumentMatcher<Ingestion>() {
```
MatkovIvan: Not any but the same with the original channel

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -875,6 +908,26 @@ private void setDefaultTransmissionTarget(String transmissionTargetToken) {
         }
     }
 
+    /**
+     * Set trigger interval. The interval should be between 3 seconds and 1 day.
+     * Should be called before the service is started.
```
annakocheshkova: We should explicitly document the unit of measurement. Parameter name implies that this is seconds, but parameter name is not javadocs.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -80,6 +81,18 @@
     @SuppressLint("StaticFieldLeak")
     private static Analytics sInstance;
 
+    /**
+     * Transmission interval minimal value.
+     */
+    @VisibleForTesting
+    static final int TRANSMISSION_INTERVAL_MINIMUM = 3;
```
dhei: The convention is to suffix with `IN_SECONDS`.

TRANSMISSION_INTERVAL_MINIMUM -> MINIMUM_TRANSMISSION_INTERVAL_IN_SECONDS

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -80,6 +81,18 @@
...
+
+    /**
+     * Transmission interval maximum value.
+     */
+    @VisibleForTesting
+    static final int TRANSMISSION_INTERVAL_MAXIMUM = 24 * 60 * 60;
```
dhei: Same as my previous comment.

TRANSMISSION_INTERVAL_MAXIMUM -> MAXIMUM_TRANSMISSION_INTERVAL_IN_SECONDS

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -875,6 +911,26 @@ private void setDefaultTransmissionTarget(String transmissionTargetToken) {
...
+     * @param seconds Set latency of sending events to Analytics.
+     * @return succeed <code>true</code> if set succeed, <code>false</code> otherwise.
+     */
+    private boolean setInstanceTransmissionInterval(int seconds) {
+        if (mChannel != null) {
+            AppCenterLog.error(LOG_TAG, "Transmission interval should be set before the service is started.");
+            return false;
+        }
+        if (seconds < TRANSMISSION_INTERVAL_MINIMUM || seconds > TRANSMISSION_INTERVAL_MAXIMUM) {
+            AppCenterLog.error(LOG_TAG, "The transmission interval is invalid. The value should be between 3 seconds and 1 day");
```
dhei: To be future proof in case we change the min/max range but forget to modify the hardcoded value in this error message, let's use `TRANSMISSION_INTERVAL_MINIMUM` and `TRANSMISSION_INTERVAL_MAXIMUM` to format the string.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -80,6 +82,18 @@
     @SuppressLint("StaticFieldLeak")
     private static Analytics sInstance;
 
+    /**
+     * Transmission interval minimal value.
```
jaelim-ms:
```suggestion
     * Transmission interval minimum value.
```

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -131,6 +145,12 @@
      */
     private AnalyticsListener mAnalyticsListener;
 
+    /**
+     * Transmission interval.
+     * Valid value ranges from 3 seconds to 1 day.
```
jaelim-ms: I don't think we need this in the doc for the variable. This is business logic so it should be enough to say it in the method doc.
```suggestion
```

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -189,6 +210,17 @@ static synchronized void unsetInstance() {
         return getInstance().setInstanceEnabledAsync(enabled);
     }
 
+    /**
+     * Set transmission interval. The transmission interval should be between 3 seconds and 1 day.
+     * Should be called before the service is started.
+     *
+     * @param seconds Set latency of sending events to Analytics in seconds.
+     * @return succeed <code>true</code> if set succeed, <code>false</code> otherwise.
```
jaelim-ms:
```suggestion
     * @return <code>true</code> if the interval is set, <code>false</code> otherwise.
```

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -131,6 +145,12 @@
      */
     private AnalyticsListener mAnalyticsListener;
 
+    /**
+     * Transmission interval.
+     * Valid value ranges from 3 seconds to 1 day.
+     */
+    private int mTransmissionInterval;
```
jaelim-ms: The value itself isn't actually transmission interval but trigger interval. Should we just say `mTriggerInterval` for this private variable? Maybe we can also say `setTriggerInterval` for the instance method and only say "transmission interval" for public facing API but it isn't consistent as well. @guperrot What's your opinion?

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -875,6 +912,30 @@ private void setDefaultTransmissionTarget(String transmissionTargetToken) {
         }
     }
 
+    /**
+     * Set trigger interval. The interval should be between 3 seconds and 1 day.
```
jaelim-ms: This doc is saying "trigger interval" which isn't consistent with the member variable name. I think we can keep it if the variable name is changed.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -875,6 +912,30 @@ private void setDefaultTransmissionTarget(String transmissionTargetToken) {
...
+     * Should be called before the service is started.
+     *
+     * @param seconds Set latency of sending events to Analytics.
+     * @return succeed <code>true</code> if set succeed, <code>false</code> otherwise.
```
jaelim-ms:
```suggestion
     * @return <code>true</code> if the interval is set, <code>false</code> otherwise.
```

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -875,6 +912,30 @@ private void setDefaultTransmissionTarget(String transmissionTargetToken) {
...
+     */
+    private boolean setInstanceTransmissionInterval(int seconds) {
+        if (mChannel != null) {
+            AppCenterLog.error(LOG_TAG, "Transmission interval should be set before the service is started.");
+            return false;
+        }
+        if (seconds < MINIMUM_TRANSMISSION_INTERVAL_IN_SECONDS || seconds > MAXIMUM_TRANSMISSION_INTERVAL_IN_SECONDS) {
+            AppCenterLog.error(LOG_TAG, String.format(Locale.getDefault(),
+                    "The transmission interval is invalid. The value should be between %d seconds and %d seconds (%d day)",
```
jaelim-ms: The doc says the range of valid value. I don't think we need to print a log with details.

jaelim-ms: I'd prefer to say "is not valid" instead of "is invalid" but not a big deal.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -589,6 +621,11 @@ public void run() {
         }, updateCurrentActivityRunnable, updateCurrentActivityRunnable);
     }
 
+    @Override
+    protected int getTriggerInterval() {
```
jaelim-ms: It returns an interval in millisecond. I wonder if we need to change the return type to long.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
guperrot: Trigger interval is not in any specification, I think this is fine to use the new name here.

guperrot: Use long and document we store in millis here

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -147,6 +167,7 @@ private Analytics() {
         mFactories.put(EventLog.TYPE, new EventLogFactory());
         mFactories.put(CommonSchemaEventLog.TYPE, new CommonSchemaEventLogFactory());
         mTransmissionTargets = new HashMap<>();
+        mTransmissionInterval = (int) TimeUnit.SECONDS.toMillis(MINIMUM_TRANSMISSION_INTERVAL_IN_SECONDS);
```
guperrot: remove `(int)` after addressing the comment to change the member variable to long

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
guperrot: Internally we use millis, so need to use long indeed

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
guperrot: I think the specification of the new API is transmission interval, use the new name here.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -875,6 +911,26 @@ private void setDefaultTransmissionTarget(String transmissionTargetToken) {
...
+     * @param seconds Set latency of sending events to Analytics.
+     * @return <code>true</code> if the interval is set, <code>false</code> otherwise.
+     */
+    private boolean setInstanceTransmissionInterval(int seconds) {
+        if (mChannel != null) {
+            AppCenterLog.error(LOG_TAG, "Transmission interval should be set before the service is started.");
+            return false;
+        }
+        if (seconds < MINIMUM_TRANSMISSION_INTERVAL_IN_SECONDS || seconds > MAXIMUM_TRANSMISSION_INTERVAL_IN_SECONDS) {
+            AppCenterLog.error(LOG_TAG,"The transmission interval is not valid. ");
```
jaelim-ms: Please reformat all files.

jaelim-ms: No space after the period.

Zongrenfeng: Search through entire project, only found this one (the space after period at the end of a message) need to be fix.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -924,7 +924,7 @@ private boolean setInstanceTransmissionInterval(int seconds) {
             return false;
         }
         if (seconds < MINIMUM_TRANSMISSION_INTERVAL_IN_SECONDS || seconds > MAXIMUM_TRANSMISSION_INTERVAL_IN_SECONDS) {
-            AppCenterLog.error(LOG_TAG,"The transmission interval is not valid. ");
+            AppCenterLog.error(LOG_TAG,"The transmission interval is not valid.");
```
guperrot: Still not formatted using Android Studio.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -589,6 +619,11 @@ public void run() {
         }, updateCurrentActivityRunnable, updateCurrentActivityRunnable);
     }
 
+    @Override
+    protected int getTriggerInterval() {
+        return  (int) mTransmissionInterval;
```
guperrot: Not formatted correctly

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -589,6 +619,11 @@ public void run() {
         }, updateCurrentActivityRunnable, updateCurrentActivityRunnable);
     }
 
+    @Override
+    protected int getTriggerInterval() {
```
guperrot: Change trigger interval (using refactor options) to return a long and remove the cast.
The channel group already accepts a long.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -877,6 +912,26 @@ private void setDefaultTransmissionTarget(String transmissionTargetToken) {
...
     }
 
+    /**
+     * Set transmission interval. The interval should be between 3 seconds and 1 day.
+     * Should be called before the service is started.
+     *
+     * @param seconds Set latency of sending events to Analytics.
+     * @return <code>true</code> if the interval is set, <code>false</code> otherwise.
+     */
+    private boolean setInstanceTransmissionInterval(int seconds) {
```
jaelim-ms: Will it be remaining as an int value?

jaelim-ms: Will it be remaining as an int value?

guperrot: its passed in seconds, stored in milliseconds.

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTest.java
```diff
@@ -863,6 +865,24 @@ public void createTransmissionTargetBeforeStart() {
         AppCenterLog.error(anyString(), contains("AppCenter is not configured"));
     }
 
+    @Test
+    public void triggerIntervalReturnFalse() {
+        Analytics analytics = Analytics.getInstance();
```
guperrot: This test is flawed in the logic.

You need to start before checking for min and max value.
As for checking for not started you need to pass a valid value to know the reason it returns false is because not started.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
dhei: Any reason not to use `long` type?

guperrot: We pass in int seconds, not need for long given the allowed values.
Long is used for internals of the SDK that uses milliseconds.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -877,6 +912,26 @@ private void setDefaultTransmissionTarget(String transmissionTargetToken) {
...
+        if (mChannel != null) {
+            AppCenterLog.error(LOG_TAG, "Transmission interval should be set before the service is started.");
+            return false;
+        }
+        if (seconds < MINIMUM_TRANSMISSION_INTERVAL_IN_SECONDS || seconds > MAXIMUM_TRANSMISSION_INTERVAL_IN_SECONDS) {
```
jwallra: I prefer for this to be split into separate blocks for "greater than" and "less than" intervals so we can supply a more informative message.  Otherwise the message should read "The transmission interval is out of range (3s-86400s): %d" and insert the provided range.

Zongrenfeng: Jae ever suggested in previous comments to remove the range:
https://github.com/microsoft/appcenter-sdk-android/pull/1145#discussion_r282239157

sdk/appcenter-analytics/src/test/java/com/microsoft/appcenter/analytics/AnalyticsTest.java
```diff
@@ -863,6 +865,24 @@ public void createTransmissionTargetBeforeStart() {
         AppCenterLog.error(anyString(), contains("AppCenter is not configured"));
     }
 
+    @Test
+    public void triggerInterval() {
+        Analytics analytics = Analytics.getInstance();
+        boolean result = Analytics.setTransmissionInterval(1);
+        assertFalse(result);
```
Jamminroot: I think we should split this test into separate.
In this test we are checking:
**On disabled module**
1) Setting less than min value (false)
2) Setting more than max value (false)
3) Setting value (true)
**On enabled module**
4) Setting less than min value (false)
It would be great if each test had one and only reason to fail, separate for each test. And, the latter (4th in a list), as pointed out by @guperrot, is not quite a valid test.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -131,6 +144,11 @@
      */
     private AnalyticsListener mAnalyticsListener;
 
+    /**
+     * Transmission interval.
```
MatkovIvan:
```suggestion
     * Transmission interval in milliseconds.
```

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -189,6 +208,17 @@ static synchronized void unsetInstance() {
         return getInstance().setInstanceEnabledAsync(enabled);
     }
 
+    /**
+     * Set transmission interval. The transmission interval should be between 3 seconds and 1 day.
```
MatkovIvan:
```suggestion
     * Set transmission interval. The transmission interval should be between 3 seconds and 86400 seconds (1 day).
```

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -189,6 +208,17 @@ static synchronized void unsetInstance() {
...
+     * Should be called before the service is started.
+     *
+     * @param seconds Set latency of sending events to Analytics in seconds.
```
MatkovIvan:
```suggestion
     * @param seconds the latency of sending events to Analytics in seconds.
```

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -877,6 +912,26 @@ private void setDefaultTransmissionTarget(String transmissionTargetToken) {
         }
     }
 
+    /**
+     * Set transmission interval. The interval should be between 3 seconds and 1 day.
+     * Should be called before the service is started.
+     *
+     * @param seconds Set latency of sending events to Analytics.
```
MatkovIvan:
```suggestion
     * @param seconds the latency of sending events to Analytics.
```

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -877,6 +912,26 @@ private void setDefaultTransmissionTarget(String transmissionTargetToken) {
         }
     }
 
+    /**
+     * Set transmission interval. The interval should be between 3 seconds and 1 day.
```
MatkovIvan:
```suggestion
     * Set transmission interval. The interval should be between 3 seconds and 86400 seconds (1 day).
```

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -877,6 +912,26 @@ private void setDefaultTransmissionTarget(String transmissionTargetToken) {
...
+     * @param seconds the latency of sending events to Analytics.
+     * @return <code>true</code> if the interval is set, <code>false</code> otherwise.
+     */
+    private boolean setInstanceTransmissionInterval(int seconds) {
+        if (mChannel != null) {
+            AppCenterLog.error(LOG_TAG, "Transmission interval should be set before the service is started.");
+            return false;
+        }
+        if (seconds < MINIMUM_TRANSMISSION_INTERVAL_IN_SECONDS || seconds > MAXIMUM_TRANSMISSION_INTERVAL_IN_SECONDS) {
+            AppCenterLog.error(LOG_TAG, "The transmission interval is not valid, it should be between 3 seconds and 1 day (86400 seconds).");
```
annakocheshkova: Use constants instead of 3 seconds and 1 day in logs. There was a comment about that some time ago but I guess it was overwritten? https://github.com/microsoft/appcenter-sdk-android/pull/1145#discussion_r282121071

MatkovIvan:
@jwallra @guperrot @jaelim-ms It's time for opening common PR for the whole feature. I'm merging this one

Issue #1141, opened 2019-05-06
Potential crash in Distribute.java


dajarami:
https://github.com/microsoft/AppCenter-SDK-Android/blob/2117635bb2ad1b030f15df944b6814e61e7e9c3b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java#L1619

Experienced a rare crash around isMandatoryUpdate().
Here's the stack trace:
 java.lang.RuntimeException: An error occurred while executing doInBackground() at android.os.AsyncTask$3.done(AsyncTask.java:354) at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:383) at java.util.concurrent.FutureTask.setException(FutureTask.java:252) at java.util.concurrent.FutureTask.run(FutureTask.java:271) at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167) at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641) at java.lang.Thread.run(Thread.java:764) Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.microsoft.appcenter.distribute.ReleaseDetails.isMandatoryUpdate()' on a null object reference at com.microsoft.appcenter.distribute.Distribute.storeDownloadRequestId(Distribute.java:1584) at com.microsoft.appcenter.distribute.DownloadTask.doInBackground(DownloadTask.java:59) at com.microsoft.appcenter.distribute.DownloadTask.doInBackground(DownloadTask.java:17) at android.os.AsyncTask$2.call(AsyncTask.java:333) at java.util.concurrent.FutureTask.run(FutureTask.java:266) ... 3 more

This was with version 1.11.2

Was this addressed in 2.0.0?

MatkovIvan:
Hi @dajarami,

Thanks for the report. It doesn't look like a known issue. I'll let you know when this will be fixed.

guperrot:
Hi @dajarami does your app have calls to `Distribute.notifyUpdateAction` and if yes can you describe how you implemented that?

dajarami:
Hey @guperrot, we're not explicitly calling `Distribute.notifyUpdateAction`, we're using the default, bare-bones implementation of:
`AppCenter.start(MainApplication.this, <appSecret>, Distribute.class);`
during our `onCreate()` of our `MainApplication.java` class of our app.

[...]

PR #1146, opened 2019-05-07
Fix NPE in distribution

MatkovIvan:
* [x] Did you add unit tests?

## Description

Fix NPE in distribution

## Related PRs or issues

#1141
[AB#61358](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/61358)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 
 ### AppCenterDistribute
 
-* **[Fix]** Fix duplicate in-app update dialog when restarting activity quickly after clicking download. Also fixes a crash when choosing "Ask me in a day" in that duplicate dialog.
+* **[Fix]** Fix duplicate in-app update dialog when restarting (or switching) activity quickly after clicking download. Also fixes a crash when choosing "Ask me in a day" in the duplicate dialog.
 * **[Fix]** Fix a crash that could occur when downloading the update with a customized dialog and then calling `Distribute.notifyUserConfirmation(UpdateAction.POSTPONE)` right after calling `Distribute.notifyUserConfirmation(UpdateAction.UPDATE)`.
 
 ## Version 2.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@
 
 ### AppCenterDistribute
 
-* **[Fix]** Fix a crash that could occur when downloading the update and calling Distribute APIs such as `notifyUserConfirmation`.
+* **[Fix]** Fix duplicate in-app update dialog when restarting activity quickly after clicking download. Also fixes a crash when choosing "Ask me in a day" in that duplicate dialog.
+* **[Fix]** Fix a crash that could occur when downloading the update with a customized dialog and then calling `Distribute.notifyUserConfirmation(UpdateAction.POSTPONE)` right after calling `Distribute.notifyUserConfirmation(UpdateAction.UPDATE)`.
 
 ## Version 2.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,13 @@
 
 ## Version 2.0.1 (Under development)
 
-### App Center Data
+### AppCenterData
 
 * **[Fix]** Fix an issue where invalid characters in the document ID are accepted at creation time but causing errors while trying to read or delete the document. The characters are `#`, `\`, `/`, `?`, and all whitespaces.
+
+### AppCenterDistribute
+
+* **[Fix]** Fix a rare crash on downloading the update.
 
 ## Version 2.0.0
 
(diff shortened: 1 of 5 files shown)


sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1601,7 +1601,7 @@ private void showDisabledToast() {
     synchronized void storeDownloadRequestId(DownloadManager downloadManager, DownloadTask task, long downloadId, long enqueueTime) {
 
         /* Check for if state changed and task not canceled in time. */
-        if (mDownloadTask == task) {
+        if (mDownloadTask == task && mReleaseDetails != null) {
```
achocron: Should the `else` that goes with this `if` be changed to an `else if`? I guess before it would execute regardless of whether `mReleaseDetails == null`

guperrot: We are preventing a crash but the root cause is unknown, how can we have a download task still pending but losing the release details object? We need to understand the root cause otherwise we will have downloads not working and we won't know why.

CHANGELOG.md
```diff
@@ -2,10 +2,14 @@
...
 ## Version 2.0.1 (Under development)
 
-### App Center Data
+### AppCenterData
 
 * **[Fix]** Fix an issue where invalid characters in the document ID are accepted at creation time but causing errors while trying to read or delete the document. The characters are `#`, `\`, `/`, `?`, and all whitespaces.
 
+### AppCenterDistribute
+
+* **[Fix]** Fix a rare crash on downloading the update.
```
jaelim-ms: I think you can give a bit more details here instead of "a rare crash".

guperrot: Yes since you have a repro now that can explain the crash, please update the description.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDownloadTest.java
```diff
@@ -786,4 +786,17 @@ public Object answer(InvocationOnMock invocation) {
         /* Verify new release checked (for example what we installed was something else than the upgrade. */
         verify(mDialog, times(2)).show();
     }
+
+    @Test
+    public void completeWorkflowOnStartDownloading() {
```
guperrot: The unit test does not match the repro scenario.

Also should we just verify notifyUserConfirmation calls are correct instead?

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeCustomizationTest.java
```diff
@@ -6,20 +6,25 @@
...
+import com.microsoft.appcenter.utils.AsyncTaskUtils;
 import com.microsoft.appcenter.utils.storage.SharedPreferencesManager;
 
 import org.junit.Test;
+import org.mockito.Mockito;
 import org.mockito.invocation.InvocationOnMock;
 import org.mockito.stubbing.Answer;
 import org.powermock.core.classloader.annotations.PrepareForTest;
 
+import static android.os.AsyncTask.Status.FINISHED;
```
jwallra: sort imports - these should be up with the android imports

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1601,7 +1602,7 @@ private void showDisabledToast() {
...
```
guperrot: I'd like to remove that check to prove it was the only reason of failing. Also will ask the customer to confirm he's calling notify. Otherwise we would need to root cause again.

guperrot: Confirmed that it's unrelated to user confirmation: https://github.com/microsoft/AppCenter-SDK-Android/issues/1141#issuecomment-491973260

We really need to understand what is going on as hiding a crash will make in-app update silently fail and we'll be really unable to debug it later.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeCustomizationTest.java
guperrot: Actually unused, I cannot stress enough that if commit dialog is not used, the cleanup actions need to be performed manually.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
guperrot: So the other scenario is click download, click back and restart activity and click postpone. If async task pool is saturated by other app tasks, you have time to do it. It also explains why it's rare.

Reviewed all code paths and cannot find anything else so this is probably the right one.

MatkovIvan:
@guperrot Thanks!

PR #1151, opened 2019-05-10
Fix more Android Studio 3.4 warnings

guperrot:

* [ ] Did you add unit tests?

## Description

Spring cleanup. Some are true positives in Crashes that would need a fix so I kept the warnings there.

### DIFF
--- a/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
+++ b/sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
@@ -148,7 +148,7 @@ public class Analytics extends AbstractAppCenterService {
      *
      * @return shared instance.
      */
-    @SuppressWarnings("WeakerAccess")
+    @SuppressWarnings({"WeakerAccess", "RedundantSuppression"})
     public static synchronized Analytics getInstance() {
         if (sInstance == null) {
             sInstance = new Analytics();
@@ -211,6 +211,7 @@ public class Analytics extends AbstractAppCenterService {
      *
      * @param listener The custom analytics listener.
      */
+    @SuppressWarnings({"WeakerAccess", "RedundantSuppression"})
     @VisibleForTesting
     protected static void setListener(AnalyticsListener listener) {
         getInstance().setInstanceListener(listener);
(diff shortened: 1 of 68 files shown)


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -148,7 +148,7 @@ private Analytics() {
      *
      * @return shared instance.
      */
-    @SuppressWarnings("WeakerAccess")
+    @SuppressWarnings({"WeakerAccess", "RedundantSuppression"})
```
guperrot: So we need this because we call it via reflection.
The batch code analysis will say it can be package-private so we suppress with "WeakerAccess", the in file edit code analysis will disagree with the batch analysis... so we need to add the second suppress to reconcile them...

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/CrashesListener.java
```diff
@@ -28,6 +28,7 @@
      * @return <code>true</code> if it requires to be confirmed by a user, otherwise <code>false</code>.
      * If this method returns <code>true</code>, {@link Crashes#notifyUserConfirmation(int)} must be called by yourself.
      */
+    @SuppressWarnings({"JavadocReference", "RedundantSuppression"})
```
guperrot: The javadoc links actually work but batch analysis complains. Then in file edit analysis disagrees so we need the second suppress to reconcile them...

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/CheckDownloadTask.java
```diff
@@ -70,7 +70,7 @@
         mReleaseDetails = releaseDetails;
     }
 
-    @SuppressWarnings("deprecation")
+    @SuppressWarnings({"deprecation", "RedundantSuppression"})
```
guperrot: javac and in file analysis do not agree with the depreaction warnings, so the second suppress is needed to reconcile them

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/DatabaseManager.java
```diff
@@ -581,7 +581,8 @@ public long getMaxSize() {
          * @param newVersion new version of the schema.
          * @return true if upgrade was managed, false to drop/create table.
          */
-        boolean onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion);
+        @SuppressWarnings({"BooleanMethodIsAlwaysInverted", "RedundantSuppression"})
```
guperrot: that's again batch analysis and in file analysis not agreeing with each other

PR #1153, opened 2019-05-13
Fix log url after changing start type

zhangcc01:

* [ ] Did you add unit tests?

## Description

Fix log url after changing start type.

## Related PRs or issues

## Misc

### DIFF
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
@@ -73,6 +73,12 @@ public class SettingsActivity extends AppCompatActivity {
 
     private static boolean sAnalyticsPaused;
 
+    private static String initialStartType;
+
+    private static String oneCollectorLogUrl = "https://mobile.events.data.microsoft.com/OneCollector/1.0";
+
+    private static String otherLogUrl = "https://in-integration.dev.avalanch.es";
+
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
@@ -80,6 +86,14 @@ public class SettingsActivity extends AppCompatActivity {
         getFragmentManager().beginTransaction()
                 .replace(android.R.id.content, new SettingsFragment())
                 .commit();
+    }
+
+    public static String defaultLogUrlFunction(String startType){
+        if(StartType.valueOf(startType) == StartType.TARGET){
+            return oneCollectorLogUrl;
+        }else{
+            return otherLogUrl;
+        }
     }
 
     public static class SettingsFragment extends android.preference.PreferenceFragment implements SharedPreferences.OnSharedPreferenceChangeListener {
@@ -91,6 +105,7 @@ public class SettingsActivity extends AppCompatActivity {
         @Override
         public void onCreate(Bundle savedInstanceState) {
             super.onCreate(savedInstanceState);
+
             addPreferencesFromResource(R.xml.settings);
             initCheckBoxSetting(R.string.appcenter_state_key, R.string.appcenter_state_summary_enabled, R.string.appcenter_state_summary_disabled, new HasEnabled() {
 
@@ -105,6 +120,7 @@ public class SettingsActivity extends AppCompatActivity {
                     return AppCenter.isEnabled().get();
                 }
             });
+
             initClickableSetting(R.string.storage_size_key, Formatter.formatFileSize(getActivity(), MainActivity.sSharedPreferences.getLong(MAX_STORAGE_SIZE_KEY, DEFAULT_MAX_STORAGE_SIZE)), new Preference.OnPreferenceClickListener() {
 
                 @Override
@@ -453,7 +469,7 @@ public class SettingsActivity extends AppCompatActivity {
             });
 
             /* Miscellaneous. */
-            String initialStartType = MainActivity.sSharedPreferences.getString(APPCENTER_START_TYPE, StartType.APP_SECRET.toString());
+            initialStartType = MainActivity.sSharedPreferences.getString(APPCENTER_START_TYPE, StartType.APP_SECRET.toString());
             initChangeableSetting(R.string.appcenter_start_type_key, initialStartType, new Preference.OnPreferenceChangeListener() {
 
                 @Override
@@ -521,7 +537,9 @@ public class SettingsActivity extends AppCompatActivity {
                     return true;
                 }
             });
-            String defaultLogUrl = getString(R.string.log_url);
+
+//            String defaultLogUrl = getString(R.string.log_url);
+            String defaultLogUrl = defaultLogUrlFunction(initialStartType);
             final String defaultLogUrlDisplay = TextUtils.isEmpty(defaultLogUrl) ? getString(R.string.log_url_set_to_production) : defaultLogUrl;
             initClickableSetting(R.string.log_url_key, MainActivity.sSharedPreferences.getString(LOG_URL_KEY, defaultLogUrlDisplay), new Preference.OnPreferenceClickListener() {
 
@@ -803,5 +821,13 @@ public class SettingsActivity extends AppCompatActivity {
 
             void onReset();
         }
+    }
+
+    public enum StartType {
+        APP_SECRET,
+        TARGET,
+        BOTH,
+        NO_SECRET,
+        SKIP_START
     }
 }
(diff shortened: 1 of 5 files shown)


apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -82,6 +88,14 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
                 .commit();
     }
 
+    public static String defaultLogUrlFunction(String startType){
+        if(StartType.valueOf(startType) == StartType.TARGET){
+            return oneCollectorLogUrl;
+        }else{
```
MatkovIvan: Please format

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -804,4 +819,12 @@ private String getCrashesFileAttachmentSummary() {
             void onReset();
         }
     }
+
+    public enum StartType {
```
MatkovIvan: Reuse enum from `MainActivity`

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -73,6 +73,12 @@
 
     private static boolean sAnalyticsPaused;
 
+    private static String initialStartType;
+
+    private static String oneCollectorLogUrl = "https://mobile.events.data.microsoft.com/OneCollector/1.0";
+
+    private static String otherLogUrl = "https://in-integration.dev.avalanch.es";
```
jaelim-ms: This value would be different based on build variant. If the app is built with jcenterDependency then the logUrl should be pointing to PROD ingestion.
You should be able to find URLs already defined in `env.xml`. Just read the URL by `log_url` key. It will automatically pick up a right value based on current build variant.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -73,6 +73,12 @@
 
     private static boolean sAnalyticsPaused;
 
+    private static String initialStartType;
+
+    private static String oneCollectorLogUrl = "https://mobile.events.data.microsoft.com/OneCollector/1.0";
```
jaelim-ms: Please do not add the URL as a constant in the code. You can find `env.xml` in `res` folder that can contain all variables related to an environment.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -82,6 +88,14 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
                 .commit();
     }
 
+    public static String defaultLogUrlFunction(String startType){
```
jwallra: Don't name a function "function"

```suggestion
    public static String generateLogUrl(String startType) {
```

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -73,6 +73,12 @@
 
     private static boolean sAnalyticsPaused;
 
+    private static String initialStartType;
```
guperrot: static variables that are not compile time constants must be prefixed with `s`

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -82,6 +88,14 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
                 .commit();
     }
 
+    public static String defaultLogUrlFunction(String startType){
+        if(StartType.valueOf(startType) == StartType.TARGET){
+            return oneCollectorLogUrl;
+        }else{
+            return otherLogUrl;
```
dhei: You can inlined the if-else using ternary operation `? :` to make it more readable.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
zhangcc01: Yes, resolved in new commit.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
zhangcc01: Yes, resolved in new commit.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
zhangcc01: Yes, resolved in new commit.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
zhangcc01: Yes, resolved in new commit.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
zhangcc01: Yes, I changed it in env.xml.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -521,7 +529,8 @@ public boolean onPreferenceClick(Preference preference) {
                     return true;
                 }
             });
-            String defaultLogUrl = getString(R.string.log_url);
+
```
MatkovIvan: Empty lines only before comments

zhangcc01: Thank you for you reminder.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -453,8 +461,8 @@ public void onReset() {
             });
 
             /* Miscellaneous. */
-            String initialStartType = MainActivity.sSharedPreferences.getString(APPCENTER_START_TYPE, StartType.APP_SECRET.toString());
-            initChangeableSetting(R.string.appcenter_start_type_key, initialStartType, new Preference.OnPreferenceChangeListener() {
+            sInitialStartType = MainActivity.sSharedPreferences.getString(APPCENTER_START_TYPE, StartType.APP_SECRET.toString());
```
guperrot: Missing a if null check here otherwise you overwrite the value and its not initial anymore.
That also causes a warning in Android Studio.
You need to commit within Android Studio with code analysis enabled or run code analysis manually.

guperrot: Missing a if null check here otherwise you overwrite the value and its not initial anymore.
That also causes a warning in Android Studio.
You need to commit within Android Studio with code analysis enabled or run code analysis manually.

apps/sasquatch/src/main/res/values/env.xml
```diff
@@ -5,7 +5,8 @@
 
 <resources xmlns:tools="http://schemas.android.com/tools" tools:ignore="TypographyDashes">
     <string name="app_secret" tools:ignore="MissingTranslation">45d1d9f6-2492-4e68-bd44-7190351eb5f3</string>
-    <string name="log_url" tools:ignore="MissingTranslation" />
+    <string name="log_url" tools:ignore="MissingTranslation">https://in.appcenter.ms</string>
```
guperrot: We should not do that otherwise we never test default SDK value. We can read the default url from code, revert that.

PR #1158, opened 2019-05-15
Support setting latency of sending events

MatkovIvan:
* [x] Did you add unit tests?

## Description

Not all the events instrumented by developers in their application are the same. They differ in importance, event type, and priority. Some events are important for business needs and developers want the SDK to send those events as soon as possible as those events are used for metrics calculation and real-time aggregation). Some events are instrumented to understand the app usage and behavior, and the developers would like those events to be sent based on a different criterion that they want to control – example: send events only once or few times a day. Developers also want the ability to control the frequency and time interval for sending the data. SDK sending the data every 3 seconds is almost real-time and it affects bandwidth and battery usage for their app.

## Related PRs or issues

[AB#56382](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/56382)


sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -868,6 +915,30 @@ private void setDefaultTransmissionTarget(String transmissionTargetToken) {
...
+     * @param seconds the latency of sending events to Analytics.
+     * @return <code>true</code> if the interval is set, <code>false</code> otherwise.
+     */
+    private boolean setInstanceTransmissionInterval(int seconds) {
+        if (mChannel != null) {
+            AppCenterLog.error(LOG_TAG, "Transmission interval should be set before the service is started.");
+            return false;
+        }
+        if (seconds < MINIMUM_TRANSMISSION_INTERVAL_IN_SECONDS || seconds > MAXIMUM_TRANSMISSION_INTERVAL_IN_SECONDS) {
+            AppCenterLog.error(LOG_TAG, String.format(Locale.getDefault(),
```
jaelim-ms: This error message is overwritten during previous merge. There is @annakocheshkova's  comment [here](https://github.com/microsoft/appcenter-sdk-android/pull/1145/commits/bdd7d1ad682f04e16600c88705fbc139ad8276a1#r283722596).

We don't need to provide those in the error log.
Please revert this one.
```
AppCenterLog.error(LOG_TAG, "The transmission interval is not valid.");
```

CHANGELOG.md
```diff
@@ -2,15 +2,19 @@
...
+### AppCenterAnalytics
+
+* **[Feature]** Support setting latency of sending events via `Analytics.setTransmissionInterval`.
+
+### AppCenterAuth
 
 * **[Fix]** Fix missing proguard rules so that the app does not have to specify them.
 
-### App Center Data
+### AppCenterData
```
guperrot: The change for spaces was intentional and suggested by @achocron on the .net release for mbaas. Next time please do not revert that for next releases.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
guperrot: There was a confusion here, the original comment was about hardcoding constants in the message, the string.format solution is fine.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
```diff
@@ -124,6 +127,20 @@ static String getLogUrl(Context context, String startType) {
     }
 
     static void startAppCenter(Application application, String startTypeString) {
+        if (MainActivity.sSharedPreferences.contains(ANALYTICS_TRANSMISSION_INTERVAL_KEY)) {
```
annakocheshkova: We should add commented out code of how it should look after.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -769,6 +830,15 @@ private void setKeyValue(String key, String value) {
             editor.apply();
         }
 
+        private String getTransmissionInterval(int interval) {
+            String sign = interval >= 0 ? "" : "-";
```
annakocheshkova: Why do we need sign? Can it be negative?

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -769,6 +830,15 @@ private void setKeyValue(String key, String value) {
...
+            interval = Math.abs(interval);
+            int hours = interval / 60 / 60;
```
annakocheshkova: We have conversion classes for this in Java

PR #1161, opened 2019-05-17
Persist timer across process restart

AnastasiaKubova:

* [x] Did you add unit tests?

## Description

When the transmission interval is overridden, we need to remember (in storage) when we started the timer.
We don't remember the transmission interval itself, we remember the timestamp of when we started waiting (transmission interval might be different next time and we just use the value configured in current process).
When we restart the application, start a timer with the time difference that is needed to honor the interval (or send logs immediately if turns out to be 0 or negative, like restarting app after 2 days).

## Related PRs or issues

[AB#61876](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/61876)


sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -436,6 +436,39 @@ private int countLogs(String whereClause, String ...whereArgs) {
         return count;
     }
 
+    @Override
+    public long getOldLog(@NonNull String group) {
```
MatkovIvan:
```suggestion
    public long getOldetLogTime(@NonNull String group) {
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -436,6 +436,39 @@ private int countLogs(String whereClause, String ...whereArgs) {
...
+        AppCenterLog.debug(LOG_TAG, "Trying to get old log from the Persistence database for " + group);
+
+        /* Query database. */
+        SQLiteQueryBuilder builder = SQLiteUtils.newSQLiteQueryBuilder();
+        builder.appendWhere(COLUMN_GROUP + " = ?");
+        List<String> selectionArgs = new ArrayList<>();
+        selectionArgs.add(group);
+        Cursor cursor = null;
+        String[] selectionArgsArray = selectionArgs.toArray(new String[0]);
+        String[] columnArgsArray = {"min(timestamp)"};
```
MatkovIvan: Use const

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -436,6 +436,39 @@ private int countLogs(String whereClause, String ...whereArgs) {
...
+    public long getOldLog(@NonNull String group) {
+        Long lastTimestamp = 0L;
+
+        /* Log. */
+        AppCenterLog.debug(LOG_TAG, "Trying to get old log from the Persistence database for " + group);
+
+        /* Query database. */
+        SQLiteQueryBuilder builder = SQLiteUtils.newSQLiteQueryBuilder();
+        builder.appendWhere(COLUMN_GROUP + " = ?");
+        List<String> selectionArgs = new ArrayList<>();
```
MatkovIvan: Use just array

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -436,6 +436,39 @@ private int countLogs(String whereClause, String ...whereArgs) {
         return count;
     }
 
+    @Override
+    public long getOldLog(@NonNull String group) {
+        Long lastTimestamp = 0L;
+
+        /* Log. */
+        AppCenterLog.debug(LOG_TAG, "Trying to get old log from the Persistence database for " + group);
```
MatkovIvan:
```suggestion
        AppCenterLog.debug(LOG_TAG, "Trying to get the time of the oldest log from the Persistence database for " + group);
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/Persistence.java
```diff
@@ -71,6 +71,14 @@ public abstract long putLog(@NonNull Log log, @NonNull String group,
      */
     public abstract int countLogs(@NonNull Date timestamp);
 
+    /**
+     * Gets an old log by group.
+     *
+     * @param group @NonNull String group.
+     * @return A last log.
```
MatkovIvan: Update the comment. Not a log, but the time

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/Persistence.java
```diff
@@ -71,6 +71,14 @@ public abstract long putLog(@NonNull Log log, @NonNull String group,
      */
     public abstract int countLogs(@NonNull Date timestamp);
 
+    /**
+     * Gets an old log by group.
+     *
+     * @param group @NonNull String group.
```
MatkovIvan:
```suggestion
     * @param group The group of the storage for logs.
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/Persistence.java
```diff
@@ -71,6 +71,14 @@ public abstract long putLog(@NonNull Log log, @NonNull String group,
      */
     public abstract int countLogs(@NonNull Date timestamp);
 
+    /**
+     * Gets the time of the oldest.
+     *
+     * @param group The group of the storage for logs.
```
annakocheshkova:
```suggestion
     * @param group Storage group name to serach for the logs.
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/Persistence.java
```diff
@@ -71,6 +71,14 @@ public abstract long putLog(@NonNull Log log, @NonNull String group,
      */
     public abstract int countLogs(@NonNull Date timestamp);
 
+    /**
+     * Gets the time of the oldest.
```
annakocheshkova:
```suggestion
     * Gets the timestamp of the oldest log in the database by the given group.
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/Persistence.java
```diff
@@ -71,6 +71,14 @@ public abstract long putLog(@NonNull Log log, @NonNull String group,
...
+     *
+     * @param group The group of the storage for logs.
+     * @return The time of the oldest.
```
annakocheshkova:
```suggestion
     * @return The timestamp of the oldest found log in the databse.
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -436,6 +436,37 @@ private int countLogs(String whereClause, String ...whereArgs) {
...
+
+        /* Log. */
+        AppCenterLog.debug(LOG_TAG, "Trying to get the time of the oldest log from the Persistence database for " + group);
+
+        /* Query database. */
+        SQLiteQueryBuilder builder = SQLiteUtils.newSQLiteQueryBuilder();
+        builder.appendWhere(COLUMN_GROUP + " = ?");
+        Cursor cursor = null;
+        String[] selectionArgsArray = {group};
+        String[] columnArgsArray = {String.format("min(%s)", COLUMN_TIMESTAMP)};
```
annakocheshkova: min? At the first look, should be max...

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
```diff
@@ -436,6 +436,37 @@ private int countLogs(String whereClause, String ...whereArgs) {
...
+        }
+        ContentValues values = mDatabaseManager.nextValues(cursor);
+        if (values != null) {
+            lastTimestamp = values.getAsLong(String.format("min(%s)", COLUMN_TIMESTAMP));
+        }
+        if (cursor != null) {
+            try {
+                cursor.close();
+            } catch (RuntimeException ignore) {
+            }
```
annakocheshkova: We usually don't ignore. Please insert the comment like in other places

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -761,7 +767,16 @@ synchronized void checkPendingLogs(@NonNull GroupState groupState) {
             triggerIngestion(groupState);
         } else if (pendingLogCount > 0 && !groupState.mScheduled) {
             groupState.mScheduled = true;
-            mAppCenterHandler.postDelayed(groupState.mRunnable, groupState.mBatchTimeInterval);
+            Date now = new Date();
```
annakocheshkova: Please move the variable delaration closer to its usage.

annakocheshkova:
Oh my. 15 tests failed :(

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -56,6 +56,12 @@
     @VisibleForTesting
     static final int CLEAR_BATCH_SIZE = 100;
 
+    /**
+     * Transmission interval minimum value.
+     */
+    @VisibleForTesting
+    static final int MINIMUM_TRANSMISSION_INTERVAL_IN_SECONDS = 3;
```
Jamminroot: Wasn't this constant declared somewhere already? I think those can be merged and moved to fitting MS...Constants module in Core. 
Correct me if im wrong

sdk/appcenter/src/main/java/com/microsoft/appcenter/persistence/DatabasePersistence.java
AnastasiaKubova: yes, because we must get the oldest log but no last

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -761,7 +767,16 @@ synchronized void checkPendingLogs(@NonNull GroupState groupState) {
             triggerIngestion(groupState);
         } else if (pendingLogCount > 0 && !groupState.mScheduled) {
             groupState.mScheduled = true;
-            mAppCenterHandler.postDelayed(groupState.mRunnable, groupState.mBatchTimeInterval);
+            Date now = new Date();
+            long delay = MINIMUM_TRANSMISSION_INTERVAL_IN_SECONDS;
```
MatkovIvan:
```suggestion
            long delay = groupState.mBatchTimeInterval;
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -761,7 +767,16 @@ synchronized void checkPendingLogs(@NonNull GroupState groupState) {
...
+            if (groupState.mBatchTimeInterval > MINIMUM_TRANSMISSION_INTERVAL_IN_SECONDS) {
+                long oldestLogTime = mPersistence.getOldestLogTime(groupState.mName);
+                delay = groupState.mBatchTimeInterval - (now.getTime() - oldestLogTime);
```
MatkovIvan:
```suggestion
                delay -= now.getTime() - oldestLogTime;
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -761,7 +767,16 @@ synchronized void checkPendingLogs(@NonNull GroupState groupState) {
             triggerIngestion(groupState);
         } else if (pendingLogCount > 0 && !groupState.mScheduled) {
             groupState.mScheduled = true;
-            mAppCenterHandler.postDelayed(groupState.mRunnable, groupState.mBatchTimeInterval);
+            long delay = groupState.mBatchTimeInterval;
+            if (groupState.mBatchTimeInterval > MINIMUM_TRANSMISSION_INTERVAL_IN_SECONDS) {
+                long oldestLogTime = mPersistence.getOldestLogTime(groupState.mName);
```
guperrot: That's not what was specified, sorry if there was a confusion.
Oldest logs can be discarded so this is not reliable implementation.
You could change max storage size and send a lot of logs, then kill the app before 1 day timer is due, and we would never send any logs.

To avoid this corner case, the proposal in the story was to store the timestamp when timer started as a separate value (in shared preferences for example).

Also that will be much simpler code overall as no need to modify database code.

sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/DefaultChannelTest.java
```diff
@@ -1071,4 +1079,24 @@ public void scheduleLogsWithNewToken() {
         listenerArgumentCaptor.getValue().onNewAuthToken("test");
         verify(channel, times(2)).checkPendingLogs(any(DefaultChannel.GroupState.class));
     }
+
+    @Test
+    public void checkPendingLogsWithMultyInterval() throws Exception {
+        ArgumentCaptor<AuthTokenContext.Listener> listenerArgumentCaptor = ArgumentCaptor.forClass(AuthTokenContext.Listener.class);
```
Jamminroot: Consider adding
```obj-c
// If
// When
// Then
// Clear
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -856,6 +883,7 @@ public synchronized void shutdown() {
             public void run() {
                 mScheduled = false;
                 triggerIngestion(GroupState.this);
+                SharedPreferencesManager.remove(START_TIMER + mName);
```
russelarms: We don't know whether the call to the server side is successful or not. 
So in this implementation, if the call isn't successful and we haven't reached batchSizeLimit to the moment, the next retry will be delayed as everything is ok.
I suggest clearing timestamp after the http call succeeds.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -761,7 +775,20 @@ synchronized void checkPendingLogs(@NonNull GroupState groupState) {
...
             groupState.mScheduled = true;
-            mAppCenterHandler.postDelayed(groupState.mRunnable, groupState.mBatchTimeInterval);
+            long delay = groupState.mBatchTimeInterval;
+            if (groupState.mBatchTimeInterval > MINIMUM_TRANSMISSION_INTERVAL_IN_SECONDS) {
+                long now = System.currentTimeMillis();
+                long startTimer = SharedPreferencesManager.getLong(START_TIMER + groupState.mName, 0);
+                if(startTimer == 0) {
+                    SharedPreferencesManager.putLong(START_TIMER + groupState.mName, now);
+                } else {
+                    delay -= now - startTimer;
```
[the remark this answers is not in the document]
MatkovIvan: @russelarms I don't think that it's required.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
MatkovIvan: @russelarms It doesn't matter. The timer has been finished.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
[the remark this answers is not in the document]
MatkovIvan: @AnastasiaKubova Let's treat any values in the future as invalid and remove it from the storage

sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/DefaultChannelTest.java
annakocheshkova: There's no such convention on Android :)

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -761,7 +775,22 @@ synchronized void checkPendingLogs(@NonNull GroupState groupState) {
...
         } else if (pendingLogCount > 0 && !groupState.mScheduled) {
             groupState.mScheduled = true;
-            mAppCenterHandler.postDelayed(groupState.mRunnable, groupState.mBatchTimeInterval);
+            long delay = groupState.mBatchTimeInterval;
+            if (groupState.mBatchTimeInterval > MINIMUM_TRANSMISSION_INTERVAL) {
+                long now = System.currentTimeMillis();
+                long startTimer = SharedPreferencesManager.getLong(START_TIMER_PREFIX + groupState.mName);
+                if (startTimer == 0) {
+                    SharedPreferencesManager.putLong(START_TIMER_PREFIX + groupState.mName, now);
+                } else if (startTimer > now) {
```
guperrot: what is this case?

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -761,7 +775,22 @@ synchronized void checkPendingLogs(@NonNull GroupState groupState) {
             triggerIngestion(groupState);
         } else if (pendingLogCount > 0 && !groupState.mScheduled) {
             groupState.mScheduled = true;
-            mAppCenterHandler.postDelayed(groupState.mRunnable, groupState.mBatchTimeInterval);
+            long delay = groupState.mBatchTimeInterval;
+            if (groupState.mBatchTimeInterval > MINIMUM_TRANSMISSION_INTERVAL) {
+                long now = System.currentTimeMillis();
+                long startTimer = SharedPreferencesManager.getLong(START_TIMER_PREFIX + groupState.mName);
+                if (startTimer == 0) {
+                    SharedPreferencesManager.putLong(START_TIMER_PREFIX + groupState.mName, now);
```
guperrot: no we must always save the start time in case application is killed.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -761,7 +775,22 @@ synchronized void checkPendingLogs(@NonNull GroupState groupState) {
             triggerIngestion(groupState);
         } else if (pendingLogCount > 0 && !groupState.mScheduled) {
             groupState.mScheduled = true;
-            mAppCenterHandler.postDelayed(groupState.mRunnable, groupState.mBatchTimeInterval);
+            long delay = groupState.mBatchTimeInterval;
+            if (groupState.mBatchTimeInterval > MINIMUM_TRANSMISSION_INTERVAL) {
+                long now = System.currentTimeMillis();
+                long startTimer = SharedPreferencesManager.getLong(START_TIMER_PREFIX + groupState.mName);
```
guperrot: We must read it only once per process (after restart), the second time the interval is starting from current time, so not using the stored state.

guperrot: nvm, thats actually better to do it like that for large values in case SDK is idle.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
MatkovIvan: @guperrot It's "always". if there is no saved value -> save

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
MatkovIvan: https://github.com/microsoft/appcenter-sdk-android/pull/1165/files#diff-0ebb2cb4967c7aee91e0b189502daf32R786

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
guperrot: It is not, you are saving only if there is no save value and that's the only time it happens.
What I was saying is we have to save at every timer invocation.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
guperrot: The comment should be in this PR instead of the other one.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
guperrot: Ok it works because there is a call to remove later, we need more comments to explain how this works.

sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/DefaultChannelTest.java
```diff
@@ -1071,4 +1086,168 @@ public void scheduleLogsWithNewToken() {
...
+        long mockSystemTime = 1L;
+        when(System.currentTimeMillis()).thenReturn(mockSystemTime);
+
+        /* Create channel. Verify scheduling logs. */
+        Persistence mockPersistence = mock(Persistence.class);
+        when(mockPersistence.countLogs(TEST_GROUP)).thenReturn(5);
+
+        Channel.GroupListener mockListener = mock(Channel.GroupListener.class);
+        AppCenterIngestion mockIngestion = mock(AppCenterIngestion.class);
+        DefaultChannel channel = spy(new DefaultChannel(mock(Context.class), UUIDUtils.randomUUID().toString(), mockPersistence, mockIngestion, mAppCenterHandler));
```
MatkovIvan: `spy` isn't required here

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -756,12 +769,54 @@ synchronized void checkPendingLogs(@NonNull GroupState groupState) {
             return;
         }
         long pendingLogCount = groupState.mPendingLogCount;
-        AppCenterLog.debug(LOG_TAG, "checkPendingLogs(" + groupState.mName + ") pendingLogCount=" + pendingLogCount);
-        if (pendingLogCount >= groupState.mMaxLogsPerBatch) {
+        long batchTimeInterval = groupState.mBatchTimeInterval;
+        AppCenterLog.debug(LOG_TAG, String.format("checkPendingLogs(%s) pendingLogCount=%s batchTimeInterval=%s", groupState.mName, pendingLogCount, batchTimeInterval));
+
+        /* If the interval is custom. */
+        if (batchTimeInterval > MINIMUM_TRANSMISSION_INTERVAL) {
```
guperrot: There was a discussion on slack for the corner cases where transmission interval is custom and we have a lot of logs. We still need to send multiple batches when the timer is due.

When I read this code this will just sent 50 logs a day (if sending every day) and will never be  able to flush queue if we get a lot.

When timer is due we must send as many batches as necessary and this does not seem to be handled.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -756,12 +769,54 @@ synchronized void checkPendingLogs(@NonNull GroupState groupState) {
...
+                    AppCenterLog.debug(LOG_TAG, "The timer for " + groupState.mName + " channel finished.");
+                }
+                return;
+            }
+
+            /* We still have to wait for the rest of the interval. */
+            else {
+                batchTimeInterval -= now - startTimer;
+            }
+        } else if (pendingLogCount >= groupState.mMaxLogsPerBatch) {
```
guperrot: If we check pending logs after sending 1 of the 3 concurrent batches, and we still have more logs to sent after daily timer due. We still have to send the remaining logs immediately (if we have to send like 10 batches once a day). We might need a boolean for checkpendinglogs to allow this run or something like that.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
guperrot: To rephrase: send 200 logs and have transmission once day: we must still send 3 parallel batches, then the 4th immediately once one of them completes.

guperrot: Also must avoid infinite sending if you keep enqueueing while "flushing", just logs enqueued before starting first batch should be sent, so use the timestamp of when timer is due (not started) to query logs is one solution to avoid that. (All logs enqueued after we started flushing will be for next timer).

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -56,6 +57,17 @@
...
 
+    /**
+     * Start of schedule timestamp.
+     */
+    static final String START_TIMER_PREFIX = "START_TIMER_PREFIX_";
+
+    /**
+     * Transmission interval minimum value, in ms.
+     */
+    @VisibleForTesting
```
guperrot: Not used in tests, there is a warning in Android Studio here.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
MatkovIvan: @guperrot 
> send 200 logs and have transmission once day: we must still send 3 parallel batches, then the 4th immediately once one of them completes.

It should work now

MatkovIvan: https://github.com/microsoft/appcenter-sdk-android/pull/1161/files#diff-0ebb2cb4967c7aee91e0b189502daf32R800

sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/DefaultChannelTest.java
```diff
@@ -1071,4 +1086,164 @@ public void scheduleLogsWithNewToken() {
...
+        DefaultChannel channel = new DefaultChannel(mock(Context.class), UUIDUtils.randomUUID().toString(), mockPersistence, mockIngestion, mAppCenterHandler);
+        channel.addGroup(TEST_GROUP, 10, mockInterval, MAX_PARALLEL_BATCHES, mockIngestion, mockListener);
+
+        /* Check values. */
+        verify(mAppCenterHandler, never()).postDelayed(any(Runnable.class), anyLong());
+        verify(mockPersistence, times(2)).countLogs(eq(TEST_GROUP));
+    }
+
+    @Test
+    public void setAppSecretIngestionNoEquals() {
```
Jamminroot: I can't quite undestand what exactly this test is checking

sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/DefaultChannelTest.java
```diff
@@ -1071,4 +1086,164 @@ public void scheduleLogsWithNewToken() {
...
+        long mockInterval = 10000;
```
Jamminroot: Why do we need this?

sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/DefaultChannelTest.java
```diff
@@ -1071,4 +1086,164 @@ public void scheduleLogsWithNewToken() {
...
+        when(mockPersistence.countLogs(TEST_GROUP)).thenReturn(0);
+        Channel.GroupListener mockListener = mock(Channel.GroupListener.class);
+        AppCenterIngestion mockIngestion1 = mock(AppCenterIngestion.class);
+        AppCenterIngestion mockIngestion2 = mock(AppCenterIngestion.class);
+        DefaultChannel channel = new DefaultChannel(mock(Context.class), UUIDUtils.randomUUID().toString(), mockPersistence, mockIngestion1, mAppCenterHandler);
+        channel.addGroup(TEST_GROUP, 10, mockInterval, MAX_PARALLEL_BATCHES, mockIngestion2, mockListener);
+        channel.setAppSecret("app-secret");
+
+        /* Check values. */
+        verify(mockIngestion1, never()).sendAsync(anyString(), anyString(), any(UUID.class), any(LogContainer.class), any(ServiceCallback.class));
```
Jamminroot: We are making sure sendAsync() is never called - but are can it actually be called in this test? Correct me if im missing something.

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
guperrot: Yes the code is confusing but it works. Unit tests should reflect this.

sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/DefaultChannelTest.java
```diff
@@ -1071,4 +1086,164 @@ public void scheduleLogsWithNewToken() {
         listenerArgumentCaptor.getValue().onNewAuthToken("test");
         verify(channel, times(2)).checkPendingLogs(any(DefaultChannel.GroupState.class));
     }
+
+    @Test
+    public void checkPendingLogsCheckDelay() throws InterruptedException {
```
guperrot: Those tests are never checking what batches we are actually sending. We need a unit test that  does steps similar to `analyticsSuccess` but with non default interval and check that we don't send batches too early when we have a lot of logs. (we need the 200 logs test basically with custom interval).

sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/DefaultChannelTest.java
AnastasiaKubova: This test coverage the point, when `groupState.Ingestion != Ingestion` in setAppSecret method

sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/DefaultChannelTest.java
AnastasiaKubova: Please, clarify what you mean?

sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/DefaultChannelTest.java
```diff
@@ -67,6 +66,8 @@
 @PrepareForTest(System.class)
 public class DefaultChannelTest extends AbstractDefaultChannelTest {
 
+    private final String START_TIMER_PREFIX = "START_TIMER_PREFIX_";
```
guperrot: It was fine to use that one. The warning was on `MINIMUM_TRANSMISSION_INTERVAL` that isnt used in tests.
START_TIMER_PREFIX could be VisibleForTest here.

guperrot:
Question:

Change transmission interval to 20s and restart.
Wait start session to be flushed (can also send an event before, does not matter).

```
2019-05-21 18:23:35.134 6359-6382/com.microsoft.appcenter.sasquatch.project D/AppCenter: checkPendingLogs(group_analytics) pendingLogCount=0 batchTimeInterval=20000
2019-05-21 18:23:35.135 6359-6382/com.microsoft.appcenter.sasquatch.project D/AppCenter: The timer value for group_analytics has been saved.
```

There is no timer started though but we saved it: is that on purpose to avoid waiting one day from next log after being idle or is that a side effect of the implementation?

Also its weird because we delete the pref once timer is done so it looks something not wanted but now I'm not sure.

PR #1171, opened 2019-05-24
Support instrument registry in newer SDK, add 

owenniblock:

* [x] Did you add unit tests?

## Description

This PR adds 2 things:

- Currently the logic for checking Environment variables will fail on newer Android SDK versions (the location of `InstrumentationRegistry` has changed. This work contains a fix for this so that several locations are checked before surfacing the failure.
- Adds `AppCenter.isRunningInAppCenterTestCloud` to the API so users can easily check whether the app is running in App Center Test Cloud. We've had several requests for this functionality from users in Test and this seemed like the simplest way to surface that value in the Android Environment.

## Related PRs or issues

N/A


sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -292,6 +299,30 @@ public static boolean isConfigured() {
...
+     */
+    public static boolean isRunningInAppCenterTestCloud() {
+        try
+        {
+            Bundle arguments = InstrumentationRegistryHelper.getArguments();
+            String runningValue = arguments.getString(RUNNING_IN_APP_CENTER);
+
+            return runningValue != null && runningValue == "1";
+        }
+        catch (LinkageError e)
```
guperrot: Use multi-catch

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -292,6 +299,30 @@ public static boolean isConfigured() {
...
+     * @return true if running in App Center Test, false otherwise
+     * (and where no test dependencies in release.
+     */
+    public static boolean isRunningInAppCenterTestCloud() {
+        try
+        {
+            Bundle arguments = InstrumentationRegistryHelper.getArguments();
+            String runningValue = arguments.getString(RUNNING_IN_APP_CENTER);
+
+            return runningValue != null && runningValue == "1";
```
guperrot: Cannot use equals for string, it would work only based on luck with intern strings.
Yoda condition can be used to avoid null checking.

```suggestion
            return "1".equals(runningValue);
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,59 @@
...
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
 
 /**
  * Wraps InstrumentationRegistry class to enable mocking in unit tests.
  */
 public class InstrumentationRegistryHelper {
+    private static List<String> locations = Arrays.asList("androidx.test.platform.app.InstrumentationRegistry",
```
guperrot: Conventions:

* Add a blank line before
* It can be a constant: add `final` and rename to `LOCATIONS`.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,59 @@
...
-        return InstrumentationRegistry.getArguments();
+        Iterator<String> iterator = locations.iterator();
+        while (iterator.hasNext()) {
+            String location = iterator.next();
+
+            try {
+                Class<?> aClass = Class.forName(location);
+                Method getArguments = aClass.getMethod("getArguments", (Class[]) null);
+                return (Bundle) getArguments.invoke(null, (Object[]) null);
+                // We need to support api level 8 and up, so we cannot combine catches into one
```
guperrot: We always put a blank line before a comment and we use `/*`.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,59 @@
...
+                // Ignore unless last item
+                if (!iterator.hasNext()) {
+                    throw e;
+                }
+            } catch (LinkageError e) {
+                // Ignore unless last item
+                if (!iterator.hasNext()) {
+                    throw e;
+                }
+            } catch (ClassNotFoundException e) {
```
guperrot: For ignoring we name it `ignore` or `ignored` that will disable the code analysis warning. Also just use Exception here for reflection we do that.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,59 @@
...
+        Iterator<String> iterator = locations.iterator();
+        while (iterator.hasNext()) {
+            String location = iterator.next();
+
+            try {
+                Class<?> aClass = Class.forName(location);
+                Method getArguments = aClass.getMethod("getArguments", (Class[]) null);
+                return (Bundle) getArguments.invoke(null, (Object[]) null);
+                // We need to support api level 8 and up, so we cannot combine catches into one
+            } catch (IllegalStateException e) {
```
guperrot: Use multi-catch for IllegalStateException | LinkageError

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### App Center
 
 * **[Fix]** Handle incorrect usage of `AppCenter.setLogUrl` API to provide readable error message.
+* **[Fix]** Check for running in App Center Test will now work on newer Android SDKs.
```
guperrot: Conflict

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
guperrot: Also this comment is mysterious.  API level minimum is 16 and I don't understand it.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,59 @@
...
+                // Ignore
+            } catch (NoSuchMethodException e) {
+                // Ignore
+            } catch (IllegalAccessException e) {
+                // Ignore
+            } catch (InvocationTargetException e) {
+                // Ignore
+            }
+        }
+
```
guperrot: remove blank line here.

guperrot:
```suggestion
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
owenniblock: Removed these comments - they came from the original source and I don't think they're really necessary

owenniblock:
Thanks for the thorough PR review - just some Unit Tests to add and then this will be ready for a re-review.

owenniblock:
@guperrot - unit tests added for the API method `isRunningInAppCenterTestCloud` - please review and let me know if there's anything else. Might be a week and a bit before I can make further changes.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,45 @@
...
-        return InstrumentationRegistry.getArguments();
+        Iterator<String> iterator = LOCATIONS.iterator();
+        while (iterator.hasNext()) {
+            String location = iterator.next();
+
+            try {
+                Class<?> aClass = Class.forName(location);
+                Method getArguments = aClass.getMethod("getArguments", (Class[]) null);
+                return (Bundle) getArguments.invoke(null, (Object[]) null);
+            } catch (IllegalStateException | LinkageError e) {
```
guperrot: LinkageError can no longer happen since reflection is now used instead of direct calls. Please remove all references to the exception.

sdk/appcenter/src/test/java/com/microsoft/appcenter/AbstractAppCenterTest.java
```diff
@@ -105,6 +105,12 @@ static void addArgumentToRegistry(String value) {
         when(InstrumentationRegistryHelper.getArguments()).thenReturn(mockBundle);
     }
 
+    static void addRunningInAppCenterToRegistry(String value) {
```
guperrot: Could move this method and all the new tests to a new dedicated sub-class.
Also there are missing tests that make code coverage check fail.

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -292,6 +300,26 @@ public static boolean isConfigured() {
...
+     */
+    public static boolean isRunningInAppCenterTestCloud() {
+        try
+        {
+            Bundle arguments = InstrumentationRegistryHelper.getArguments();
+            String runningValue = arguments.getString(RUNNING_IN_APP_CENTER);
+
+            return "1".equals(runningValue);
+        }
+        catch (LinkageError | IllegalStateException e)
```
guperrot: Remove LinkageError

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,45 @@
...
+                if (!iterator.hasNext()) {
+                    throw e;
+                }
+            } catch (Exception ignored) {
```
guperrot: We might want to `throw new IllegalStateException(e)` and not ignore. That way it will preserve the https://github.com/microsoft/appcenter-sdk-android/blob/develop/sdk/appcenter/src/main/java/com/microsoft/appcenter/ServiceInstrumentationUtils.java#L54 debug message.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,45 @@
...
+            }
+        }
+        return new Bundle();
```
guperrot: Should actually not return an empty bundle if we could not query any test facility it must print this message: https://github.com/microsoft/appcenter-sdk-android/blob/develop/sdk/appcenter/src/main/java/com/microsoft/appcenter/ServiceInstrumentationUtils.java#L54

Suggestion: simplify code by using Java "foreach", assign exception to a local variable. If we exit for loop, throw that exception instead of returning empty bundle (wrapping as IllegalStateException is fine).

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -292,6 +300,26 @@ public static boolean isConfigured() {
...
 
+    /**
+     * Check whether app is running in App Center Test.
+     *
+     * @return true if running in App Center Test, false otherwise
+     * (and where no test dependencies in release.
+     */
+    public static boolean isRunningInAppCenterTestCloud() {
+        try
+        {
```
Jamminroot: Please follow formatting to style

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -292,6 +300,26 @@ public static boolean isConfigured() {
         return getInstance().isInstanceConfigured();
     }
 
+    /**
+     * Check whether app is running in App Center Test.
+     *
+     * @return true if running in App Center Test, false otherwise
+     * (and where no test dependencies in release.
```
dhei: Typo? I don't see a close parentheses `)`.

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
dhei: @owenniblock, please follow SDK coding/formatting guideline at https://msmobilecenter.visualstudio.com/Mobile-Center/_wiki/wikis/Mobile-Center.wiki?wikiVersion=GBwikiMaster&pagePath=%2FHome%2FTeams%2FSDKs%2FGet%20started%20with%20SDK%20development%2FAndroid%20SDK%20extra%20topics&pageId=2135&anchor=before-pushing-your-changes when using Android Studio. Thank you.

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -104,6 +106,12 @@
     @VisibleForTesting
     static final String TRANSMISSION_TARGET_TOKEN_KEY = "target";
 
+    /**
+     * Environment variable name for test to see if we're running in App Center Test
+     */
+    @VisibleForTesting
+    static String RUNNING_IN_APP_CENTER = "RUNNING_IN_APP_CENTER";
```
dhei:
```suggestion
    static final String RUNNING_IN_APP_CENTER = "RUNNING_IN_APP_CENTER";
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -292,6 +300,26 @@ public static boolean isConfigured() {
...
+     * @return true if running in App Center Test, false otherwise
+     * (and where no test dependencies in release.
+     */
+    public static boolean isRunningInAppCenterTestCloud() {
+        try
+        {
+            Bundle arguments = InstrumentationRegistryHelper.getArguments();
+            String runningValue = arguments.getString(RUNNING_IN_APP_CENTER);
+
+            return "1".equals(runningValue);
```
dhei: Please make "1" a `static final` variable.

owenniblock:
I believe I've addressed everything but please let me know if I've missed anything :)

I guess we should think about adding the same method to the iOS SDK - I'll discuss that here and try and get something on our board.

CHANGELOG.md
guperrot: We already prepared (and thus froze code) for release 2.1.0, so this will have to be moved to 2.2.0.

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -292,6 +306,23 @@ public static boolean isConfigured() {
...
+     * Check whether app is running in App Center Test.
+     *
+     * @return true if running in App Center Test, false otherwise
+     * (and where no test dependencies in release).
+     */
+    public static boolean isRunningInAppCenterTestCloud() {
+        try {
+            Bundle arguments = InstrumentationRegistryHelper.getArguments();
+            String runningValue = arguments.getString(RUNNING_IN_APP_CENTER);
+
```
guperrot:
```suggestion
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,46 @@
...
      * @return the instrumentation arguments.
-     * @throws LinkageError          if the class, method is not found or does not match, typically no test dependencies in release.
-     * @throws IllegalStateException if no argument Bundle has been registered.
+     * @throws IllegalStateException if it cannot call getArguments using any of the provided classes.
      */
-    public static Bundle getArguments() throws LinkageError, IllegalStateException {
-        return InstrumentationRegistry.getArguments();
+    public static Bundle getArguments() throws IllegalStateException {
+        Exception exception = null;
+
```
guperrot:
```suggestion
```

CHANGELOG.md
guperrot:
```suggestion
* **[Fix]** Check for running in App Center Test will now work when using AndroidX instead of the support library.
```
This is not directly related to Android version but rather which test framework being used.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,46 @@
...
+        for (String location : LOCATIONS) {
+            try {
+                Class<?> aClass = getClass(location);
+                Method getArguments = aClass.getMethod("getArguments", (Class[]) null);
+                return (Bundle) getArguments.invoke(null, (Object[]) null);
+            } catch (Exception e) {
+                exception = e;
+            }
+        }
+
```
guperrot:
```suggestion
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,46 @@
 package com.microsoft.appcenter.utils;
 
 import android.os.Bundle;
-import android.support.test.InstrumentationRegistry;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
```
guperrot: Unused imports

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,46 @@
...
      */
-    public static Bundle getArguments() throws LinkageError, IllegalStateException {
-        return InstrumentationRegistry.getArguments();
+    public static Bundle getArguments() throws IllegalStateException {
+        Exception exception = null;
+
+        for (String location : LOCATIONS) {
+            try {
+                Class<?> aClass = getClass(location);
+                Method getArguments = aClass.getMethod("getArguments", (Class[]) null);
```
guperrot: Code analysis warning here which is actually true, this method does not have parameters.
Thus the code has to be:
```suggestion
                Method getArguments = aClass.getMethod("getArguments");
```

Was this code actually tested in a instrumentation environment with either AndroidX or support?

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,46 @@
...
+                return (Bundle) getArguments.invoke(null, (Object[]) null);
+            } catch (Exception e) {
+                exception = e;
+            }
+        }
+
+        throw new IllegalStateException(exception);
+    }
+
+    public static Class<?> getClass(String className) throws ClassNotFoundException {
```
guperrot: Please make this method package private (no keyword) and move the test class to same package.

sdk/appcenter/src/test/java/com/microsoft/appcenter/IsRunningInAppCenterTest.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+package com.microsoft.appcenter;
```
guperrot: Needs to moved to `com.microsoft.appcenter.utils`

sdk/appcenter/src/test/java/com/microsoft/appcenter/IsRunningInAppCenterTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import android.content.Context;
+
+import com.microsoft.appcenter.utils.InstrumentationRegistryHelper;
```
guperrot: Please clean up / optimize imports with Android Studio.

guperrot:
Please also remove the `depedencies {}` block in `sdk/appcenter.gradle` as the compileOnly was to avoid reflection and use LinkageError instead, now that code has been replaced by reflection so the dependencies are not used anymore.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
owenniblock: The original code came from [our Espresso Extension](https://github.com/microsoft/AppCenter-Test-Espresso-Extensions/blob/7610ea64ffd4709beeebe038fed4a2bdb84057ab/src/main/java/com/microsoft/appcenter/espresso/Factory.java#L40) - I was confident it would work as I'm lead to believe that is working correctly. However the code has deviated quite a bit from my original push so I'll run some tests to make sure it works in practice.

guperrot: The signature is definitely incorrect for reflection, getArguments does not have a class array parameter.
You need to call getArguments without parameter.

sdk/appcenter/src/test/java/com/microsoft/appcenter/IsRunningInAppCenterTest.java
owenniblock: That's more fiddly that I expected it to be - suggest we keep the first 3 Tests which specifically call `AppCenter` API in this package and move the other 2 Tests (which are really concerning InstrumentationRegistryHelper directly in to the `InstrumentationRegistryHelperTest.java` file so they more naturally sit within that package.

guperrot: sounds good

owenniblock:
I've done some testing in the local app to make sure it works. I added the following tests:

```
    @Test
    public void notRunningInAppCenter() {
        assertFalse(AppCenter.isRunningInAppCenterTestCloud());
    }

    @Test
    public void isRunningInAppCenter() {
        assertTrue(AppCenter.isRunningInAppCenterTestCloud());
    }
```

The first 1 succeeds when running normally with:
> ./gradlew --parallel coverageReport

The 2nd one succeeds when running with:
> ./gradlew --parallel coverageReport  -Pandroid.testInstrumentationRunnerArguments.RUNNING_IN_APP_CENTER=1

I won't push up this change because either one will fail when running in App Center Test but it validates the approach.

I'll push up the latest version now. We discussed adding `isRunningInAppCenter` to the iOS SDK and because it's trivial in iOS we decided that it would be of little benefit.

owenniblock:
Sorry for the `force-push` seemed the best way to sort out the history / CHANGELOG.md file.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # App Center SDK for Android Change Log
 
+## Version 2.2.0 (Under development)
+
+### App Center
+
+* **[Fix]** Check for running in App Center Test will now work when using AndroidX instead of the support library.
+* **[Feature]** Add `AppCenter.isRunningInAppCenterTestCloud` to provide method to check if app is running in Test Cloud.
```
guperrot:
```suggestion
* **[Feature]** Add `AppCenter.isRunningInAppCenterTestCloud` to provide method to check if the application is running in Test Cloud.
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
guperrot: The `Class[]` has to be removed to resolve the warnings, you can skip varargs when calling a method without argument by reflection.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,40 @@
...
      */
-    public static Bundle getArguments() throws LinkageError, IllegalStateException {
-        return InstrumentationRegistry.getArguments();
+    public static Bundle getArguments() throws IllegalStateException {
+        Exception exception = null;
+        for (String location : LOCATIONS) {
+            try {
+                Class<?> aClass = getClass(location);
+                Method getArguments = aClass.getMethod("getArguments", (Class[]) null);
+                return (Bundle) getArguments.invoke(null, (Object[]) null);
```
guperrot: Here too, remove the second parameter.

owenniblock: Done - sorry, I was sure I'd done these!

guperrot:
>     @Test
>     public void notRunningInAppCenter() {
>         assertFalse(AppCenter.isRunningInAppCenterTestCloud());
>     }
> 
>     @Test
>     public void isRunningInAppCenter() {
>         assertTrue(AppCenter.isRunningInAppCenterTestCloud());
>     }

Thanks for the tip how to test.
I will try it also.

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -104,6 +106,18 @@
...
+    /**
+     * Environment variable name for test to see if we're running in App Center Test.
+     */
+    @VisibleForTesting
+    static final String RUNNING_IN_APP_CENTER = "RUNNING_IN_APP_CENTER";
+
+    /**
+     * A string value for environment variables denoting `true`.
+     */
+    @VisibleForTesting
```
guperrot:
```suggestion
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -104,6 +106,18 @@
...
+    static final String TRUE_ENVIRONMENT_STRING = "1";
```
guperrot:
```suggestion
    private static final String TRUE_ENVIRONMENT_STRING = "1";
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
```diff
@@ -6,22 +6,40 @@
...
+                Method getArguments = aClass.getMethod("getArguments");
+                return (Bundle) getArguments.invoke(null);
+            } catch (Exception e) {
+                exception = e;
+            }
+        }
+        throw new IllegalStateException(exception);
+    }
+
+    static Class<?> getClass(String className) throws ClassNotFoundException {
```
guperrot: Now you don't use it in test anymore, has to be inlined to make the warning go away.

sdk/appcenter/src/test/java/com/microsoft/appcenter/IsRunningInAppCenterTest.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+package com.microsoft.appcenter;
+
+import android.os.Bundle;
```
guperrot: Unused imports again. Please use Android Studio to commit and check all options except TODO, that will avoid these comments.

sdk/appcenter/src/test/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelperTest.java
```diff
@@ -5,12 +5,48 @@
 
 package com.microsoft.appcenter.utils;
 
+import com.microsoft.appcenter.AbstractAppCenterTest;
+import com.microsoft.appcenter.AppCenter;
+import java.lang.reflect.Method;
+import org.junit.Rule;
```
guperrot: Imports to cleanup again.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/InstrumentationRegistryHelper.java
owenniblock: We do use this is in the test don't we?

```
        Method getClass = method(InstrumentationRegistryHelper.class, "getClass", String.class);
        doThrow(new ClassNotFoundException()).when(InstrumentationRegistryHelper.class, getClass)
```

guperrot: Why calling via reflection? Please increase visibility and add `@VisibleForTesting`.

owenniblock: I struggled to get a mockito `mock` or `spy` on a static class which stubs `getClass` but not `getArguments` - I had another go at this but couldn't find a better way to do it using the API directly.

guperrot: I guess it's fine as is. There are other ways to do that but it's not worth it (isolate a utility and mockStatic, or mock the LOCATIONS with different values).

guperrot:
```suggestion
    /**
     * Get class by name. Separate method so that can be overridden by mocks in unit tests. Do not inline.
     *
     * @param className class name to resolve.
     * @return the Class object from name.
     * @throws ClassNotFoundException if the class could no be found.
     */
    private static Class<?> getClass(String className) throws ClassNotFoundException {
```

guperrot:
(Restarting build due to emulator starting issue.)

PR #1172, opened 2019-05-24
Fix store interval when channel paused

MatkovIvan:
* [x] Did you add unit tests?

## Description

Tracking an event with 1min interval / pause / resume after 30s / event is sent after 30s. Pausing doesn’t reset the timer. It’s probably fine since it's the same behavior if paused for any reasons like going to background on iOS or network down.
The issue is the behavior is different if the `trackEvent` happens while pausing; in this case the timer will wait for the entire latency. For instance: Pause / Tracking an event with 1min interval / resume after 30s / event is sent after 1min 30s.

## Related PRs or issues

[AB#62192](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/62192)

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
@@ -34,7 +34,6 @@ import com.microsoft.appcenter.utils.context.AuthTokenInfo;
 import com.microsoft.appcenter.utils.storage.SharedPreferencesManager;
 
 import java.io.IOException;
-import java.text.DateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -767,7 +766,7 @@ public class DefaultChannel implements Channel {
     @VisibleForTesting
     synchronized void checkPendingLogs(@NonNull GroupState groupState) {
         AppCenterLog.debug(LOG_TAG, String.format("checkPendingLogs(%s) pendingLogCount=%s batchTimeInterval=%s",
-                groupState.mName, groupState.mPendingLogCount, DateFormat.getTimeInstance().format(new Date(groupState.mBatchTimeInterval))));
+                groupState.mName, groupState.mPendingLogCount, groupState.mBatchTimeInterval));
         Long batchTimeInterval = resolveTriggerInterval(groupState);
 
         /* Check if there is no need to trigger ingestion. */
@@ -817,7 +816,7 @@ public class DefaultChannel implements Channel {
                 return groupState.mBatchTimeInterval;
             }
 
-            /* We still have to wait for the rest of the interval. */
+            /* Wait for the rest of the interval. */
             return Math.max(groupState.mBatchTimeInterval - (now - startTimer), 0);
         } else {
 
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
@@ -33,6 +33,7 @@ import com.microsoft.appcenter.utils.context.AuthTokenInfo;
 import com.microsoft.appcenter.utils.storage.SharedPreferencesManager;
 
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


MatkovIvan:
@guperrot that's from task description:
> Fix: store the oldest log timestamp even when the channel is paused.

Also, I made some refactoring to improve the readability of this piece of code and align it to iOS implementation (move resolving interval to the separate function), so GH shows a bit more lines in the diff now

sdk/appcenter/src/main/java/com/microsoft/appcenter/channel/DefaultChannel.java
```diff
@@ -764,54 +765,75 @@ public synchronized void enqueue(@NonNull Log log, @NonNull final String groupNa
...
         }
+    }
 
-        /* Postpone triggering ingestion. */
-        if (pendingLogCount > 0 && !groupState.mScheduled) {
-            groupState.mScheduled = true;
-            mAppCenterHandler.postDelayed(groupState.mRunnable, batchTimeInterval);
+    private Long resolveDefaultTriggerInterval(@NonNull GroupState groupState) {
+        if (groupState.mPendingLogCount >= groupState.mMaxLogsPerBatch) {
+            return 0L;
```
annakocheshkova: Can you elaborate?

MatkovIvan: @annakocheshkova This piece isn't changed. It's like: if there are enough logs for "full" batch then send it immediately.

sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/DefaultChannelTest.java
```diff
@@ -1104,6 +1080,34 @@ public void checkPendingLogsStoresStartTime() {
...
+        verify(mAppCenterHandler, never()).postDelayed(any(Runnable.class), anyLong());
+
+        channel.pauseGroup(TEST_GROUP, null);
+        channel.enqueue(mock(Log.class), TEST_GROUP, Flags.DEFAULTS);
+
+        /* Verify that timer starts and current time is saved into preferences. */
+        verifyStatic();
+        SharedPreferencesManager.putLong(eq(START_TIMER_PREFIX + TEST_GROUP), eq(now));
+        verify(mockIngestion, never()).sendAsync(anyString(), anyString(), any(UUID.class), any(LogContainer.class), any(ServiceCallback.class));
+        verify(mAppCenterHandler, never()).postDelayed(any(Runnable.class), anyLong());
```
guperrot: This test is incomplete, we must simulate some time has passed then resume and check sent immediately.

guperrot: Also please move this new test to `DefaultChannelPauseAndResumeTest`, we already split the test some time back as the main class is cluttered.

Let's try to keep them split by themes.

sdk/appcenter/src/test/java/com/microsoft/appcenter/channel/DefaultChannelTest.java
```diff
@@ -1104,6 +1080,34 @@ public void checkPendingLogsStoresStartTime() {
...
+        channel.addGroup(TEST_GROUP, 10, CUSTOM_INTERVAL, MAX_PARALLEL_BATCHES, mockIngestion, mock(Channel.GroupListener.class));
+        verifyStatic(never());
+        SharedPreferencesManager.putLong(eq(START_TIMER_PREFIX + TEST_GROUP), anyLong());
+        verify(mockIngestion, never()).sendAsync(anyString(), anyString(), any(UUID.class), any(LogContainer.class), any(ServiceCallback.class));
+        verify(mAppCenterHandler, never()).postDelayed(any(Runnable.class), anyLong());
+
+        channel.pauseGroup(TEST_GROUP, null);
+        channel.enqueue(mock(Log.class), TEST_GROUP, Flags.DEFAULTS);
+
+        /* Verify that timer starts and current time is saved into preferences. */
```
guperrot: The comment is wrong here, the timer does not start.

PR #1174, opened 2019-06-01
Fix etag being overwritten in cache

anjbur:
* [x] Did you add unit tests?

## Description

Bugfix so that on offline update and delete, any previous cached etag value will not be overwritten.

## Related PRs or issues

[AB#61109](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/61109)

## Misc

### DIFF
--- a/sdk/appcenter-data/src/androidTest/java/com/microsoft/appcenter/data/LocalDocumentStorageAndroidTest.java
+++ b/sdk/appcenter-data/src/androidTest/java/com/microsoft/appcenter/data/LocalDocumentStorageAndroidTest.java
@@ -149,6 +149,24 @@ public class LocalDocumentStorageAndroidTest {
     }
 
     @Test
+    public void offlineDeletePreservesEtagNullDocumentValue() {
+
+        /* Write document to the cache with an etag set. */
+        DocumentWrapper<Void> document = new DocumentWrapper<>(null, DefaultPartitions.USER_DOCUMENTS, ID, "etag", 0);
+        mLocalDocumentStorage.writeOnline(USER_TABLE_NAME, document, new WriteOptions());
+        DocumentWrapper<String> createdDocument = mLocalDocumentStorage.read(USER_TABLE_NAME, DefaultPartitions.USER_DOCUMENTS, ID, String.class, new ReadOptions());
+        assertNotNull(createdDocument.getETag());
+        assertEquals("etag", createdDocument.getETag());
+
+        /* Delete document offline, etag should be replaced. */
+        boolean success = mLocalDocumentStorage.deleteOffline(USER_TABLE_NAME, DefaultPartitions.USER_DOCUMENTS, ID, new WriteOptions());
+        assertTrue(success);
+        DocumentWrapper<Void> deletedDoc = mLocalDocumentStorage.read(USER_TABLE_NAME, DefaultPartitions.USER_DOCUMENTS, ID, Void.class, null);
+        assertNotNull(deletedDoc.getETag());
+        assertEquals("etag", deletedDoc.getETag());
+    }
+
+    @Test
     public void readExpiredDocument() {
 
         /* Write a document and mock device ttl to be already expired a few seconds ago. */
--- a/sdk/appcenter-data/src/androidTest/java/com/microsoft/appcenter/data/LocalDocumentStorageAndroidTest.java
+++ b/sdk/appcenter-data/src/androidTest/java/com/microsoft/appcenter/data/LocalDocumentStorageAndroidTest.java
@@ -110,6 +110,45 @@ public class LocalDocumentStorageAndroidTest {
     }
 
     @Test
+    public void offlineUpdatePreservesEtag() {
+
+        /* Write document to the cache with an etag set. */
+        DocumentWrapper<String> document = new DocumentWrapper<>(TEST_VALUE, DefaultPartitions.USER_DOCUMENTS, ID, "etag", 0);
+        mLocalDocumentStorage.writeOnline(USER_TABLE_NAME, document, new WriteOptions());
+        DocumentWrapper<String> createdDocument = mLocalDocumentStorage.read(USER_TABLE_NAME, DefaultPartitions.USER_DOCUMENTS, ID, String.class, new ReadOptions());
+        assertNotNull(createdDocument.getETag());
+        assertEquals("etag", createdDocument.getETag());
+
+        /* Replace document offline, etag should be replaced. */
+        mLocalDocumentStorage.createOrUpdateOffline(USER_TABLE_NAME, DefaultPartitions.APP_DOCUMENTS, ID, "value2", String.class, new WriteOptions());
+        DocumentWrapper<String> updatedDoc = mLocalDocumentStorage.read(USER_TABLE_NAME, DefaultPartitions.USER_DOCUMENTS, ID, String.class, new ReadOptions());
+        assertNotNull(updatedDoc.getETag());
+        assertEquals("etag", updatedDoc.getETag());
+    }
+
+    @Test
+    public void offlineDeletePreservesEtag() {
+
+        /* Delete a non-existant document offline, etag should be null. */
+        mLocalDocumentStorage.deleteOffline(USER_TABLE_NAME, DefaultPartitions.USER_DOCUMENTS, ID, new WriteOptions());
+        DocumentWrapper<String> deletedDoc = mLocalDocumentStorage.read(USER_TABLE_NAME, DefaultPartitions.USER_DOCUMENTS, ID, String.class, null);
+        assertNull(deletedDoc.getETag());
+
+        /* Write document to the cache with an etag set. */
+        DocumentWrapper<String> document = new DocumentWrapper<>(TEST_VALUE, DefaultPartitions.USER_DOCUMENTS, ID, "etag", 0);
+        mLocalDocumentStorage.writeOnline(USER_TABLE_NAME, document, new WriteOptions());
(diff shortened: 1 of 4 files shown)


sdk/appcenter-data/src/androidTest/java/com/microsoft/appcenter/data/LocalDocumentStorageAndroidTest.java
```diff
@@ -109,6 +109,63 @@ public void writeReadDelete() {
...
+    public void offlineUpdatePreservesEtag() {
+
+        /* Write document to the cache with an etag set. */
+        DocumentWrapper<String> document = new DocumentWrapper<>(TEST_VALUE, DefaultPartitions.USER_DOCUMENTS, ID, "etag", 0);
+        mLocalDocumentStorage.writeOnline(USER_TABLE_NAME, document, new WriteOptions());
+        DocumentWrapper<String> createdDocument = mLocalDocumentStorage.read(USER_TABLE_NAME, DefaultPartitions.USER_DOCUMENTS, ID, String.class, new ReadOptions());
+        assertNotNull(createdDocument.getETag());
+        assertEquals("etag", createdDocument.getETag());
+
+        /* Replace document offline, etag should be replaced. */
```
guperrot:
```suggestion
        /* Replace document offline, etag should not be replaced. */
```

sdk/appcenter-data/src/androidTest/java/com/microsoft/appcenter/data/LocalDocumentStorageAndroidTest.java
```diff
@@ -109,6 +109,63 @@ public void writeReadDelete() {
...
+        assertNull(deletedDoc.getETag());
+
+        /* Write document to the cache with an etag set. */
+        DocumentWrapper<String> document = new DocumentWrapper<>(TEST_VALUE, DefaultPartitions.USER_DOCUMENTS, ID, "etag", 0);
+        mLocalDocumentStorage.writeOnline(USER_TABLE_NAME, document, new WriteOptions());
+        DocumentWrapper<String> createdDocument = mLocalDocumentStorage.read(USER_TABLE_NAME, DefaultPartitions.USER_DOCUMENTS, ID, String.class, new ReadOptions());
+        assertNotNull(createdDocument.getETag());
+        assertEquals("etag", createdDocument.getETag());
+
+        /* Delete document offline, etag should be replaced. */
```
guperrot:
```suggestion
        /* Delete document offline, etag should not be replaced. */
```

sdk/appcenter-data/src/androidTest/java/com/microsoft/appcenter/data/LocalDocumentStorageAndroidTest.java
```diff
@@ -109,6 +109,63 @@ public void writeReadDelete() {
...
+    public void offlineDeletePreservesEtagNullDocumentValue() {
+
+        /* Write document to the cache with an etag set. */
+        DocumentWrapper<Void> document = new DocumentWrapper<>(null, DefaultPartitions.USER_DOCUMENTS, ID, "etag", 0);
+        mLocalDocumentStorage.writeOnline(USER_TABLE_NAME, document, new WriteOptions());
+        DocumentWrapper<String> createdDocument = mLocalDocumentStorage.read(USER_TABLE_NAME, DefaultPartitions.USER_DOCUMENTS, ID, String.class, new ReadOptions());
+        assertNotNull(createdDocument.getETag());
+        assertEquals("etag", createdDocument.getETag());
+
+        /* Delete document offline, etag should be replaced. */
```
guperrot:
```suggestion
        /* Delete document offline, etag should not be replaced. */
```

guperrot:
```
> Task :sdk:appcenter-data:compileDebugJavaWithJavac FAILED
/Users/vsts/agent/2.150.3/work/1/s/sdk/appcenter-data/src/main/java/com/microsoft/appcenter/data/LocalDocumentStorage.java:35: error: cannot find symbol
import static com.microsoft.appcenter.data.Utils.FAILED_TO_DESERIALIZE_DOCUMENT;
^
  symbol:   static FAILED_TO_DESERIALIZE_DOCUMENT
  location: class
/Users/vsts/agent/2.150.3/work/1/s/sdk/appcenter-data/src/main/java/com/microsoft/appcenter/data/LocalDocumentStorage.java:216: error: cannot find symbol
        if (cachedDocument.getError() != null && !cachedDocument.getError().getMessage().equals(FAILED_TO_DESERIALIZE_DOCUMENT)) {
                                                                                                ^
  symbol:   variable FAILED_TO_DESERIALIZE_DOCUMENT
  location: class LocalDocumentStorage
2 errors
```

sdk/appcenter-data/src/main/java/com/microsoft/appcenter/data/LocalDocumentStorage.java
```diff
@@ -206,7 +211,13 @@ boolean deleteOffline(String table, DocumentWrapper<Void> documentWrapper, Write
      * @return true if storage update succeeded, false otherwise.
      */
     boolean deleteOffline(String table, String partition, String documentId, WriteOptions writeOptions) {
-        DocumentWrapper<Void> writeDocument = new DocumentWrapper<>(null, partition, documentId);
+        DocumentWrapper<Void> cachedDocument = read(table, partition, documentId, Void.class, null);
+        DocumentWrapper<Void> writeDocument;
+        if (cachedDocument.getError() != null && !cachedDocument.getError().getMessage().equals(FAILED_TO_DESERIALIZE_DOCUMENT)) {
```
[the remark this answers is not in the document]
anjbur: We have to read from the cache to get the current etag so that it can be preserved when we update the pending operation in the cache. We don't get a document class type on delete though, which is required for read. Passing `Void.class` on read works, but the document fails to deserialize (which doesn't matter because we're only interested in the etag in the metadata, which isn't impacted by the deserialization of the document content).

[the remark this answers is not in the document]
guperrot: That looks like an anti-pattern. Why do we try to deserialize something that we'll know will fail and then check for this error while all of this could be avoided altogether?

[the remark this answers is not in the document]
anjbur: I was basing it off the pattern in `createOrUpdateOffline` where there are different behaviors depending on if the read failed with a specific error and then whether it failed/succeeded generally. I can refactor to add a method that only reads the metadata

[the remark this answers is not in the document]
guperrot: Why not simply always call line 219? cachedDocument.getETag() will return null anyway on "error but not quite" so the flow would be the same, we probably don't need a if here, am I missing something?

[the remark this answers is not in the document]
anjbur: No you're right that makes sense, fixed it to remove the if.

sdk/appcenter-data/src/androidTest/java/com/microsoft/appcenter/data/LocalDocumentStorageAndroidTest.java
```diff
@@ -109,6 +109,45 @@ public void writeReadDelete() {
...
+    public void offlineUpdatePreservesEtag() {
+
+        /* Write document to the cache with an etag set. */
+        DocumentWrapper<String> document = new DocumentWrapper<>(TEST_VALUE, DefaultPartitions.USER_DOCUMENTS, ID, "etag", 0);
+        mLocalDocumentStorage.writeOnline(USER_TABLE_NAME, document, new WriteOptions());
+        DocumentWrapper<String> createdDocument = mLocalDocumentStorage.read(USER_TABLE_NAME, DefaultPartitions.USER_DOCUMENTS, ID, String.class, new ReadOptions());
+        assertNotNull(createdDocument.getETag());
+        assertEquals("etag", createdDocument.getETag());
+
+        /* Replace document offline, etag should be not replaced. */
```
guperrot:
```suggestion
        /* Replace document offline, etag should not be replaced. */
```

That was the original suggestion ;) Can be applied from files tab directly on github as batch.

sdk/appcenter-data/src/androidTest/java/com/microsoft/appcenter/data/LocalDocumentStorageAndroidTest.java
```diff
@@ -109,6 +109,45 @@ public void writeReadDelete() {
...
+        assertNull(deletedDoc.getETag());
+
+        /* Write document to the cache with an etag set. */
+        DocumentWrapper<String> document = new DocumentWrapper<>(TEST_VALUE, DefaultPartitions.USER_DOCUMENTS, ID, "etag", 0);
+        mLocalDocumentStorage.writeOnline(USER_TABLE_NAME, document, new WriteOptions());
+        DocumentWrapper<String> createdDocument = mLocalDocumentStorage.read(USER_TABLE_NAME, DefaultPartitions.USER_DOCUMENTS, ID, String.class, new ReadOptions());
+        assertNotNull(createdDocument.getETag());
+        assertEquals("etag", createdDocument.getETag());
+
+        /* Delete document offline, etag should be not replaced. */
```
guperrot:
```suggestion
        /* Delete document offline, etag should not be replaced. */
```

sdk/appcenter-data/src/androidTest/java/com/microsoft/appcenter/data/LocalDocumentStorageAndroidTest.java
anjbur: Ah good to know. Should help prevent having a typo in typo fixes 😅

PR #1175, opened 2019-06-03
Expand UserInformation to include access token and id token.

dhei:

* [x] Did you add unit tests?

## Description

Expand UserInformation to include access token and id token:

```java
public class UserInformation {
    public String getAccountId();
    public String getAccessToken(); 
    public String getIdToken();
}
```

## Related PRs or issues

[AB#62364](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/62364)


sdk/appcenter-auth/src/main/java/com/microsoft/appcenter/auth/Auth.java
```diff
@@ -306,7 +306,7 @@ private synchronized void setInstanceConfigUrl(String configUrl) {
     private synchronized void removeTokenAndAccount() {
         AuthTokenContext authTokenContext = AuthTokenContext.getInstance();
         removeAccount(authTokenContext.getHomeAccountId());
-        authTokenContext.setAuthToken(null, null, null);
```
guperrot: We don't need to modify the auth context since there is no getUser.
The only time we need access token is when the public API is called.
Many changes have thus to be reverted here.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 
 ### App Center Auth
 
+* **[Feature]** Expose the ID Token and Access Token JWTs in the `UserInformation` object passed to the sign in callback.
```
guperrot: "Token JWT" is redundant as the T stands for Token.
Also Android is not using a completion handler but returning a future, so it's not a callback per say.
Thus I suggest:

```suggestion
* **[Feature]** Expose the ID Token and Access Token (as raw JWT format) in the `UserInformation` object returned from the sign-in method.
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/UserInformation.java
```diff
@@ -15,13 +15,37 @@
...
+    /**
+     * Access token for user.
+     */
+    private final String mAccessToken;
+
+    /**
+     * Id token for user.
+     */
+    private final String mIdToken;
+
```
guperrot:
```suggestion
```

sdk/appcenter-auth/src/test/java/com/microsoft/appcenter/auth/AuthTest.java
```diff
@@ -96,6 +96,57 @@
     @Captor
     private ArgumentCaptor<Map<String, String>> mHeadersCaptor;
```
guperrot: Looks like whitespaces are changed on every single line, can this be fixed?

jaelim-ms: If you are referring below methods, it is moved to the top due to private static methods order.

guperrot: Looks like last commits fixed it somehow.

PR #1192, opened 2019-06-18
Fix the issue: cannot clear userid

jinquan2019:

* [ ] Did you add unit tests?

## Description
Set the value of userid is null when userid is empty string

### DIFF
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
@@ -878,8 +878,7 @@ public class SettingsActivity extends AppCompatActivity {
             String summary = MainActivity.sSharedPreferences.getString(preferencesKey, defaultValue);
             if (summary == null) {
                 return getString(R.string.unset_summary);
-            }
-            else if (summary.isEmpty()) {
+            } else if (summary.isEmpty()) {
                 return getString(R.string.empty_summary);
             }
             return summary;
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
@@ -566,20 +566,12 @@ public class SettingsActivity extends AppCompatActivity {
                     Toast.makeText(getActivity(), String.format(getActivity().getString(R.string.target_id_changed_format), defaultTargetId), Toast.LENGTH_SHORT).show();
                 }
             });
-            initEditText(R.string.user_id_key, R.string.user_id_title, USER_ID_KEY, getString(R.string.user_id_unset), new EditTextListener() {
+            initEditText(R.string.user_id_key, R.string.user_id_title, USER_ID_KEY, null, new EditTextListener() {
 
                 @Override
                 public void onSave(String value) {
-                    if (value != null && value.isEmpty()) {
-                        setKeyValue(USER_ID_KEY, getString(R.string.user_id_empty));
-                        MainActivity.setUserId(value);
-                    }
-                    else {
-                        if (!value.equals(getString(R.string.user_id_unset)) && !value.equals(getString(R.string.user_id_empty))) {
-                            setKeyValue(USER_ID_KEY, value);
-                            MainActivity.setUserId(value);
-                        }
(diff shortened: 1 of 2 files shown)


jwallra:
We don't have any documentation that the `userId` can't be empty so we'll have to have a docs change for this

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -570,6 +570,9 @@ public void onReset() {
 
                 @Override
                 public void onSave(String value) {
+                    if (value != null && value.isEmpty()){
```
guperrot: It is by design to test empty string passed to SDK and see that AppCenter accepts it and sends it and that One Collector rejects it with `E/AppCenter: userId must not be empty.`.

You have to use the Reset button to erase.

@jaelim-ms we should cancel this bug.

jaelim-ms: This is for child targets which we don't have a dialog and reset button. We can implement this one for it but it could be tested by the one in the root level.

guperrot: Ok this is the wrong fix then and needs to be started over.
After talking to @jaelim-ms what we need to modify is this UI:

Screen Shot 2019-06-18 at 1 56 42 PM

As you can see UserId is blank there, we need to say something like "unset" for null and "empty string" for empty string for the displayed value in the settings list.

The editor (onSave) must not be patched.

guperrot: > This is for child targets which we don't have a dialog and reset button. We can implement this one for it but it could be tested by the one in the root level.
The bug was reported at root level, not for children.

guperrot:
> We don't have any documentation that the `userId` can't be empty so we'll have to have a docs change for this

Because we don't forbid empty string for AppCenter, it's only for OneCollector. This PR has the wrong fix. See https://github.com/microsoft/appcenter-sdk-android/pull/1192#discussion_r295025087

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
jinquan2019: I update the fix to address the comment. But I still patch the onSave method.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -566,12 +566,20 @@ public void onReset() {
...
                 }
             });
-            initEditText(R.string.user_id_key, R.string.user_id_title, USER_ID_KEY, null, new EditTextListener() {
+            initEditText(R.string.user_id_key, R.string.user_id_title, USER_ID_KEY, getString(R.string.user_id_unset), new EditTextListener() {
 
                 @Override
                 public void onSave(String value) {
-                    setKeyValue(USER_ID_KEY, value);
-                    MainActivity.setUserId(value);
+                    if (value != null && value.isEmpty()) {
```
guperrot: Please do not change onSave and what is stored, please rollback this (it was already asked before).
It's only when we display the value that we will use the new texts.

jinquan2019: Update the fix via add a method to get summary when the userid is null or empty

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -874,6 +874,17 @@ private String getCrashesFileAttachmentSummary() {
             return getString(R.string.appcenter_crashes_file_attachment_summary_empty);
         }
 
+        private String getSummary(final String preferencesKey, final String defaultValue) {
+            String summary = MainActivity.sSharedPreferences.getString(preferencesKey, defaultValue);
+            if (summary == null) {
+                return getString(R.string.unset_summary);
+            }
+            else if (summary.isEmpty()) {
```
guperrot: reformat file

PR #1195, opened 2019-06-24
Break data tests into separate files

nikitag-ms:
* [ ] Did you add unit tests?

## Description

Break data tests into separate files. No new tests have been added, no existing ones deleted. Only some code moved around.


sdk/appcenter-data/src/test/java/com/microsoft/appcenter/data/DataCreateUpdateDeleteTest.java
```diff
@@ -0,0 +1,495 @@
...
+import static org.mockito.Matchers.refEq;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.powermock.api.mockito.PowerMockito.mock;
+import static org.powermock.api.mockito.PowerMockito.mockStatic;
+import static org.powermock.api.mockito.PowerMockito.verifyNoMoreInteractions;
+import static org.powermock.api.mockito.PowerMockito.when;
+
+@PrepareForTest({
```
guperrot: Can this be moved to Abstract class without side effect? Seems like duplicated in all children.

sdk/appcenter-data/src/test/java/com/microsoft/appcenter/data/DataListTest.java
```diff
@@ -0,0 +1,655 @@
...
+
+    @Captor
+    private ArgumentCaptor<DocumentWrapper<TestDocument>> mTestDocumentWrapperCaptor;
+
+    private void mockSignOut() {
+        Mockito.when(mAuthTokenContext.getAccountId()).thenReturn(null);
+    }
+
+    @Before
+    public void setUpAuth() {
```
guperrot: Isn't that duplicated also?

nikitag-ms: Yes, it is somewhat duplicating across files. But because there are so many subclasses off `AbstractDataTest`, there are a few other tests in some other files that assume that auth context is not set up. So if I moved this one under a `@Before` in the parent class, other tests would need to be adjusted.

PR #1199, opened 2019-06-26
Upgrade android Data DB schema version

nikitag-ms:

* [x] Did you add unit tests?

## Description

We need to drop both user and readonly tables because we discovered we had been doing inserts instead of upserts into the cache in the original SDK release. Dropping the table will allow to clean up what essentially are duplicate rows (the same partition and document ID, but different `oid`).

### DIFF
--- a/sdk/appcenter-data/src/main/java/com/microsoft/appcenter/data/LocalDocumentStorage.java
+++ b/sdk/appcenter-data/src/main/java/com/microsoft/appcenter/data/LocalDocumentStorage.java
@@ -8,6 +8,7 @@ package com.microsoft.appcenter.data;
 import android.content.ContentValues;
 import android.content.Context;
 import android.database.Cursor;
+import android.database.sqlite.SQLiteDatabase;
 import android.database.sqlite.SQLiteQueryBuilder;
 import android.support.annotation.NonNull;
 import android.support.annotation.VisibleForTesting;
@@ -87,7 +88,7 @@ class LocalDocumentStorage {
     /**
      * Current version of the schema.
      */
-    private static final int VERSION = 1;
+    private static final int VERSION = 2;
 
     /**
      * `Where` clause to select by partition and document ID.
@@ -103,8 +104,14 @@ class LocalDocumentStorage {
 
     private final DatabaseManager mDatabaseManager;
 
-    LocalDocumentStorage(Context context, String userTable) {
-        mDatabaseManager = new DatabaseManager(context, DATABASE, READONLY_TABLE, VERSION, SCHEMA, new DatabaseManager.DefaultListener());
+    LocalDocumentStorage(Context context, final String userTable) {
+        mDatabaseManager = new DatabaseManager(
+                context,
+                DATABASE,
+                READONLY_TABLE,
+                VERSION,
+                SCHEMA,
+                new LocalDocumentStorageDatabaseListener(userTable));
         if (userTable != null) {
             createTableIfDoesNotExist(userTable);
         }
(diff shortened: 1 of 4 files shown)


sdk/appcenter-data/src/main/java/com/microsoft/appcenter/data/LocalDocumentStorage.java
```diff
@@ -418,3 +425,34 @@ void updatePendingOperation(LocalDocument operation) {
         return new DocumentWrapper<>(new DataException("Document was not found in the cache."));
     }
 }
+
+class LocalDocumentStorageDatabaseListener implements DatabaseManager.Listener {
+
+    private String userTable;
```
guperrot: mUserTable

sdk/appcenter-data/src/main/java/com/microsoft/appcenter/data/LocalDocumentStorage.java
```diff
@@ -418,3 +425,34 @@ void updatePendingOperation(LocalDocument operation) {
...
+
+    LocalDocumentStorageDatabaseListener(String userTable) {
+        this.userTable = userTable;
```
guperrot: mUserTable = userTable, no this

sdk/appcenter-data/src/main/java/com/microsoft/appcenter/data/LocalDocumentStorage.java
```diff
@@ -418,3 +425,34 @@ void updatePendingOperation(LocalDocument operation) {
...
+    }
+
+    @Override
+    public void onCreate(SQLiteDatabase db) {
+    }
+
+    @Override
+    public boolean onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
+        /*
```
guperrot: add new line and use the /* convention that we have where every line has the *

sdk/appcenter-data/src/main/java/com/microsoft/appcenter/data/LocalDocumentStorage.java
```diff
@@ -418,3 +425,34 @@ void updatePendingOperation(LocalDocument operation) {
...
+            Between versions 1 and 2 we need to drop both user and readonly tables
+            because we discovered we had been doing inserts instead of upserts into the cache in the original SDK release.
+            Dropping the table will allow to clean up what essentially are duplicate rows
+            (the same partition and document ID, but different `oid`).
+         */
+        if (oldVersion == 1 && newVersion == 2 && userTable != null) {
```
guperrot: why testing newVersion == 2? If later you upgrade from 1 to 3 directly you also need to execute that. We just need to check oldVersion is 1

guperrot: Also why the fix does not apply to app documents?

sdk/appcenter-data/src/test/java/com/microsoft/appcenter/data/LocalDocumentStorageTest.java
```diff
@@ -295,4 +297,36 @@ public void getPendingOperationsOnNonUserTable() {
...
+        localDocumentStorageDatabaseListenerTest(2, 2, null, false, true);
+        localDocumentStorageDatabaseListenerTest(2, 2, userTableName, false, true);
+    }
+
+    private void localDocumentStorageDatabaseListenerTest(int oldVersion, int newVersion, String userTableName, boolean expectTableDrop, boolean expectedOnUpgradeResult) {
+        SQLiteDatabase db = mock(SQLiteDatabase.class);
+        LocalDocumentStorageDatabaseListener listener =
+                new LocalDocumentStorageDatabaseListener(userTableName);
+
+        /* Verify no interactions in `onCreate` */
```
guperrot: Dots at end of comments.

sdk/appcenter-data/src/main/java/com/microsoft/appcenter/data/LocalDocumentStorage.java
nikitag-ms: App documents will get deleted automatically by the `DatabaseManager` (line 123) as the comment points out.

PR #1206, opened 2019-07-03
Fix gradle dependencies for Data module

guperrot:
When used without Auth and if the app does not have gson, it would not
pull gson in the app binary before the change.

* [ ] Did you add unit tests?

## Description

Fix gradle dependencies for Data module


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 * **[Fix]** Fix `isExpired` method in `LocalDocument` for incorrect handling of the `TimeToLive.INFINITE` value. 
 * **[Feature]** Add `ReadOptions` parameter to the `list` API.
 * **[Feature]** Serialize `null` document values.
+* **[Fix]** Fix declaring `gson` as a build time dependency instead of runtime.
```
jaelim-ms: Not blocking and not your change. We need to group it together by Fix and Feature. L15.

guperrot:
emulator was hanging at start

PR #1216, opened 2019-07-29
Handle the memory warning event and save the info

AnastasiaKubova:
Please have a look at our [guidelines for contributions](https://github.com/microsoft/appcenter-sdk-android/blob/develop/CONTRIBUTING.md) and consider the following before you submit the PR:

* [ ] ~Has `CHANGELOG.md` been updated?~
* [x] Are tests passing locally?
* [x] Are the files formatted correctly?
* [x] Did you add unit tests?
* [ ] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?
* [ ] Did you check UI tests on the sample app? They are not executed on CI.

## Description

Add the handle the memory warning event and save the info.

## Related PRs or issues

[AB#65505](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/65505)

PR #1217, opened 2019-07-29
Added UI for OOM

annakocheshkova:
* [ ] Did you add unit tests?

## Description

1. UI button to generate OOM warning
2. UI tile to display OOM warning

[AB#65510](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/65510)

### DIFF
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
@@ -166,7 +166,7 @@ public class CrashActivity extends AppCompatActivity {
                     nativeAbortCall();
                 }
             }),
-            new Crash(R.string.title_low_memory_warning, R.string.description_oom, new Runnable() {
+            new Crash(R.string.title_low_memory_warning, R.string.description_low_memory_warning, new Runnable() {
 
                 @Override
                 public void run() {
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
@@ -22,6 +22,7 @@ import com.microsoft.appcenter.crashes.Crashes;
 import com.microsoft.appcenter.crashes.model.TestCrashException;
 import com.microsoft.appcenter.sasquatch.R;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Random;
@@ -157,6 +158,18 @@ public class CrashActivity extends AppCompatActivity {
                 @Override
                 public void run() {
                     nativeAbortCall();
+                }
+            }),
+            new Crash(R.string.title_oom, R.string.description_oom, new Runnable() {
+
+                @SuppressWarnings({"MismatchedQueryAndUpdateOfCollection", "InfiniteLoopStatement"})
+                @Override
+                public void run() {
+                    int value = 5;
+                    List<Integer> list = new ArrayList<>();
+                    while (true) {
+                        list.add(value);
+                    }
                 }
             })
     );
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
@@ -6,11 +6,15 @@
 package com.microsoft.appcenter.sasquatch.activities;
 
 import android.content.Intent;
+import android.graphics.Bitmap;
 import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.annotation.VisibleForTesting;
 import android.support.v7.app.AppCompatActivity;
+import android.util.Log;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.AdapterView;
@@ -26,17 +30,16 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static com.microsoft.appcenter.sasquatch.activities.CrashSubActivity.INTENT_EXTRA_CRASH_TYPE;
+import static com.microsoft.appcenter.sasquatch.activities.MainActivity.LOG_TAG;
 
 public class CrashActivity extends AppCompatActivity {
 
     private boolean mCrashSuperPauseNotCalled;
 
     private boolean mCrashSuperDestroyNotCalled;
-
-    @SuppressWarnings("FieldCanBeLocal")
-    private static List<Integer[]> sList;
 
     private final List<Crash> sCrashes = Arrays.asList(
             new Crash(R.string.title_test_crash, R.string.description_test_crash, new Runnable() {
@@ -168,13 +171,21 @@ public class CrashActivity extends AppCompatActivity {
                 @SuppressWarnings({"MismatchedQueryAndUpdateOfCollection"})
                 @Override
                 public void run() {
-                    sList = new ArrayList<>();
-                    for (int i = 0; i < 10000000; i++) {
-                        sList.add(new Integer[10]);
-                    }
+                    final AtomicInteger i = new AtomicInteger(0);
+                    final Handler handler = new Handler(Looper.getMainLooper());
+                    handler.post(new Runnable() {
+                        @Override
+                        public void run() {
+                            nativeAllocateLargeBuffer();
+                            Log.d(LOG_TAG, "Memory allocated: " + i.addAndGet(128) + "MB");
+                            handler.post(this);
+                        }
+                    });
                 }
             })
     );
+
+    private native void nativeAllocateLargeBuffer();
 
     private native void nativeDereferenceNullPointer();
 
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
@@ -160,7 +160,7 @@ public class CrashActivity extends AppCompatActivity {
                     nativeAbortCall();
                 }
             }),
-            new Crash(R.string.title_oom, R.string.description_oom, new Runnable() {
+            new Crash(R.string.title_low_memory_warning, R.string.description_oom, new Runnable() {
 
                 @SuppressWarnings({"MismatchedQueryAndUpdateOfCollection", "InfiniteLoopStatement"})
                 @Override
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
@@ -35,6 +35,9 @@ public class CrashActivity extends AppCompatActivity {
 
     private boolean mCrashSuperDestroyNotCalled;
 
+    @SuppressWarnings("FieldCanBeLocal")
+    private static List<Integer[]> sList;
+
     private final List<Crash> sCrashes = Arrays.asList(
             new Crash(R.string.title_test_crash, R.string.description_test_crash, new Runnable() {
 
@@ -162,13 +165,12 @@ public class CrashActivity extends AppCompatActivity {
             }),
             new Crash(R.string.title_low_memory_warning, R.string.description_oom, new Runnable() {
 
-                @SuppressWarnings({"MismatchedQueryAndUpdateOfCollection", "InfiniteLoopStatement"})
+                @SuppressWarnings({"MismatchedQueryAndUpdateOfCollection"})
                 @Override
                 public void run() {
-                    int value = 5;
-                    List<Integer> list = new ArrayList<>();
-                    while (true) {
-                        list.add(value);
+                    sList = new ArrayList<>();
+                    for (int i = 0; i < 10000000; i++) {
+                        sList.add(new Integer[10]);
                     }
                 }
             })
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
@@ -168,7 +168,6 @@ public class CrashActivity extends AppCompatActivity {
             }),
             new Crash(R.string.title_low_memory_warning, R.string.description_oom, new Runnable() {
 
-                @SuppressWarnings({"MismatchedQueryAndUpdateOfCollection"})
                 @Override
                 public void run() {
                     final AtomicInteger i = new AtomicInteger(0);
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
@@ -56,10 +56,10 @@ public final class TestFeatures {
 
     @SuppressWarnings("unchecked")
     private static boolean hadMemoryWarning() {
-        // TODO: Replace with return Crashes.hadMemoryWarningInLastSession().get();
+
+        /* TODO: Replace with return Crashes.hadMemoryWarningInLastSession().get(); when updating the demo during release process. */
         try {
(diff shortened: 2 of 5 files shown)


apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
```diff
@@ -42,13 +45,31 @@ public static void initialize(Activity parentActivity) {
         sTestFeatureModels.add(new TestFeatureTitle(R.string.crashes_title));
         sTestFeatureModels.add(new TestFeature(R.string.title_crashes, R.string.description_crashes, CrashActivity.class));
         sTestFeatureModels.add(new TestFeature(R.string.title_error, R.string.description_error, ManagedErrorActivity.class));
+        sTestFeatureModels.add(new TestFeature(R.string.title_had_memory_warning, hadMemoryWarning() ? R.string.description_had_memory_warning : R.string.description_didnt_have_memory_warning, new View.OnClickListener() {
```
MatkovIvan: Can we pass null here?

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
```diff
@@ -42,13 +45,31 @@ public static void initialize(Activity parentActivity) {
...
 
+    @SuppressWarnings("unchecked")
+    private static boolean hadMemoryWarning() {
+        // TODO: Replace with return Crashes.hadMemoryWarningInLastSession();
+        try {
+            @SuppressWarnings("unchecked") final Class<? extends AppCenterService> crashes = (Class<? extends AppCenterService>) Class.forName("com.microsoft.appcenter.crashes.Crashes");
+            final Method hadMemoryWarning = crashes.getMethod("hadMemoryWarningInLastSession");
+            return ((AppCenterFuture<Boolean>) hadMemoryWarning.invoke(null)).get();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
MatkovIvan: just ignore it

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
```diff
@@ -42,13 +45,31 @@ public static void initialize(Activity parentActivity) {
...
         sTestFeatureModels.add(new TestFeatureTitle(R.string.miscellaneous_title));
         sTestFeatureModels.add(new TestFeature(R.string.title_custom_properties, R.string.description_custom_properties, CustomPropertiesActivity.class));
         sTestFeatureModels.add(new TestFeature(R.string.title_device_info, R.string.description_device_info, DeviceInfoActivity.class));
         sTestFeatureModels.add(new TestFeatureTitle(R.string.title_data));
         sTestFeatureModels.add(new TestFeature(R.string.title_data, R.string.description_data, DataActivity.class));
     }
 
+    @SuppressWarnings("unchecked")
+    private static boolean hadMemoryWarning() {
+        // TODO: Replace with return Crashes.hadMemoryWarningInLastSession();
```
MatkovIvan: `boolean` != `AppCenterFuture<Boolean>`

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
```diff
@@ -42,13 +45,31 @@ public static void initialize(Activity parentActivity) {
...
+        try {
+            @SuppressWarnings("unchecked") final Class<? extends AppCenterService> crashes = (Class<? extends AppCenterService>) Class.forName("com.microsoft.appcenter.crashes.Crashes");
+            final Method hadMemoryWarning = crashes.getMethod("hadMemoryWarningInLastSession");
+            return ((AppCenterFuture<Boolean>) hadMemoryWarning.invoke(null)).get();
```
MatkovIvan: Can we avoid block UI here?
BTW Why do we need async getter and `AppCenterFuture`?

apps/sasquatch/src/main/res/values/strings.xml
```diff
@@ -24,6 +24,9 @@
     <string name="error_device_info" tools:ignore="MissingTranslation">Could not retrieve device information</string>
     <string name="title_error" tools:ignore="MissingTranslation">Errors</string>
     <string name="description_error" tools:ignore="MissingTranslation">Send an exception</string>
+    <string name="title_had_memory_warning" tools:ignore="MissingTranslation">OOM warning</string>
```
MatkovIvan: Why OOM? it's just a warning

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
annakocheshkova: No. The listener is final. I wanted to change that but didn't

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
annakocheshkova: There's `.get()`

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
annakocheshkova: https://github.com/microsoft/appcenter-sdk-android/pull/1216/files#diff-f948cee626f7c221be3d4eee50bedcb8R340

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
annakocheshkova: Ah you mean the original api

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
guperrot: final variables can be null, not sure what is the problem here with null

guperrot: Just need to check for null when using it

annakocheshkova: It either complains it's not initialized or can't find the matching constructor

annakocheshkova: Had to add a cast which I don't like

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
```diff
@@ -42,13 +46,26 @@ public static void initialize(Activity parentActivity) {
...
     }
 
+    @SuppressWarnings("unchecked")
+    private static boolean hadMemoryWarning() {
+        // TODO: Replace with return Crashes.hadMemoryWarningInLastSession().get();
+        try {
+            @SuppressWarnings("unchecked") final Class<? extends AppCenterService> crashes = (Class<? extends AppCenterService>) Class.forName("com.microsoft.appcenter.crashes.Crashes");
+            final Method hadMemoryWarning = crashes.getMethod("hadMemoryWarningInLastSession");
+            return ((AppCenterFuture<Boolean>) hadMemoryWarning.invoke(null)).get();
+        } catch (Exception e) {
```
MatkovIvan: I guess lint will complain
```suggestion
        } catch (Exception ignored) {
```

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
```diff
@@ -42,13 +46,26 @@ public static void initialize(Activity parentActivity) {
...
         sTestFeatureModels.add(new TestFeatureTitle(R.string.miscellaneous_title));
         sTestFeatureModels.add(new TestFeature(R.string.title_custom_properties, R.string.description_custom_properties, CustomPropertiesActivity.class));
         sTestFeatureModels.add(new TestFeature(R.string.title_device_info, R.string.description_device_info, DeviceInfoActivity.class));
         sTestFeatureModels.add(new TestFeatureTitle(R.string.title_data));
         sTestFeatureModels.add(new TestFeature(R.string.title_data, R.string.description_data, DataActivity.class));
     }
 
+    @SuppressWarnings("unchecked")
+    private static boolean hadMemoryWarning() {
+        // TODO: Replace with return Crashes.hadMemoryWarningInLastSession().get();
```
guperrot:
```suggestion

        /* TODO: Replace with return Crashes.hadMemoryWarningInLastSession().get(); when updating the demo during release process. */
```

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
```diff
@@ -42,13 +46,26 @@ public static void initialize(Activity parentActivity) {
...
+        try {
+            @SuppressWarnings("unchecked") final Class<? extends AppCenterService> crashes = (Class<? extends AppCenterService>) Class.forName("com.microsoft.appcenter.crashes.Crashes");
```
guperrot: The Crashes class exists in both variants, delete this.
```suggestion
```

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
```diff
@@ -42,13 +46,26 @@ public static void initialize(Activity parentActivity) {
...
+            final Method hadMemoryWarning = crashes.getMethod("hadMemoryWarningInLastSession");
```
guperrot:
```suggestion
            Method hadMemoryWarning = Crashes.class.getMethod("hadMemoryWarningInLastSession");
```

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
```diff
@@ -42,13 +46,26 @@ public static void initialize(Activity parentActivity) {
...
+            return ((AppCenterFuture<Boolean>) hadMemoryWarning.invoke(null)).get();
+        } catch (Exception ignored) {
```
guperrot: It is not ignored as it's handled with a return
```suggestion
        } catch (Exception e) {
```

jaelim-ms: I think its meaning is that the exception instance is ignored, not the exception itself. 🤔

apps/sasquatch/src/main/res/values/crashes.xml
```diff
@@ -37,6 +37,8 @@
     <string name="description_native_stack_overflow_crash" tools:ignore="MissingTranslation">Caused by infinite recursion</string>
     <string name="title_native_abort_crash" tools:ignore="MissingTranslation">NDK call abort()</string>
     <string name="description_native_abort_crash" tools:ignore="MissingTranslation">Simply call abort() to terminate</string>
+    <string name="title_oom" tools:ignore="MissingTranslation">Low memory warning</string>
```
jaelim-ms: nit: `title_low_memory_warning` seems better.

apps/sasquatch/src/main/res/values/strings.xml
```diff
@@ -24,6 +24,9 @@
     <string name="error_device_info" tools:ignore="MissingTranslation">Could not retrieve device information</string>
     <string name="title_error" tools:ignore="MissingTranslation">Errors</string>
     <string name="description_error" tools:ignore="MissingTranslation">Send an exception</string>
+    <string name="title_had_memory_warning" tools:ignore="MissingTranslation">Low memory warning</string>
+    <string name="description_had_memory_warning" tools:ignore="MissingTranslation">The application had a memory warning</string>
+    <string name="description_didnt_have_memory_warning" tools:ignore="MissingTranslation">The application didn\'t have a memory warning</string>
```
jaelim-ms:
```suggestion
    <string name="description_didnot_have_memory_warning" tools:ignore="MissingTranslation">The application didn\'t have a memory warning</string>
```

achocron: Actually @jaelim-ms why not
```suggestion
    <string name="description_did_not_have_memory_warning" tools:ignore="MissingTranslation">The application didn\'t have a memory warning</string>
```

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/features/TestFeatures.java
guperrot: No it's a convention in Android Studio.
If you have an exception called ignored, it's used to suppress a warning when your catch block is empty.
If your catch block is not empty you should not use that variable name.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
```diff
@@ -158,6 +159,18 @@ public void run() {
                 public void run() {
                     nativeAbortCall();
                 }
+            }),
+            new Crash(R.string.title_low_memory_warning, R.string.description_oom, new Runnable() {
+
+                @SuppressWarnings({"MismatchedQueryAndUpdateOfCollection", "InfiniteLoopStatement"})
+                @Override
+                public void run() {
+                    int value = 5;
```
guperrot: This is not really different from the existing oom crash buttons.
Basically I tried on emulator and cannot make this work, I never get a warning.
@jaelim-ms how did you test?

jaelim-ms: What I did was similar but my list contained an array of int.
```
public void run() {
  sList = new ArrayList<>();
  for (int i = 0; i < 10000000; i++) {
    sList.add(new Integer[10]);
  }
}
```
Most important thing is that the list should NOT be a local variable since it could be GCed to reserve additional space.

guperrot: Let's update the test and make sure it's tested in the app successfully then.

annakocheshkova: I had a warning when I tested, although I tested just on app, not the SDK API. For some reason the warning handler does get called only if it is overriden in the Application. I will do some more testing now

annakocheshkova: Umfff... So finally with the help of NDK works.

apps/sasquatch/src/main/res/values/crashes.xml
```diff
@@ -37,6 +37,8 @@
     <string name="description_native_stack_overflow_crash" tools:ignore="MissingTranslation">Caused by infinite recursion</string>
     <string name="title_native_abort_crash" tools:ignore="MissingTranslation">NDK call abort()</string>
     <string name="description_native_abort_crash" tools:ignore="MissingTranslation">Simply call abort() to terminate</string>
+    <string name="title_low_memory_warning" tools:ignore="MissingTranslation">Low memory warning</string>
+    <string name="description_oom" tools:ignore="MissingTranslation">Generate low memory warning/crash</string>
```
jaelim-ms: Why don't you use same text here?
```suggestion
    <string name="description_low_memory_warning" tools:ignore="MissingTranslation">Generate low memory warning/crash</string>
```

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/CrashActivity.java
```diff
@@ -158,9 +165,27 @@ public void run() {
                 public void run() {
                     nativeAbortCall();
                 }
+            }),
+            new Crash(R.string.title_low_memory_warning, R.string.description_oom, new Runnable() {
```
jaelim-ms:
```suggestion
            new Crash(R.string.title_low_memory_warning, R.string.description_low_memory_warning, new Runnable() {
```

apps/sasquatch/src/main/res/values/crashes.xml
annakocheshkova: Oops, overlooked

guperrot:
The build passes on develop but is broken on the base memory branch apparently.

There is one Crash unit test that fails, please make sure to fix the base branch tests.

PR #1222, opened 2019-08-08
Fix delayed start push

guperrot:

* [x] Did you add unit tests?

## Description

Noticed this while testing something on React Native. Firebase now initialize on its own and can trigger the token update before AppCenter is initialized (which is especially visible with delayed push scenario). And we have an optimization not to send the same token twice but it was used before checking for started/enabled...

### DIFF
--- a/sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
+++ b/sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
@@ -271,16 +271,27 @@ public class Push extends AbstractAppCenterService {
      */
     @SuppressWarnings("WeakerAccess") /* protected so that Xamarin can use it. */
     protected synchronized void onTokenRefresh(final String pushToken) {
-        if (pushToken != null && !pushToken.equals(mLatestPushToken)) {
-            AppCenterLog.debug(LOG_TAG, "Push token refreshed: " + pushToken);
-            mLatestPushToken = pushToken;
+        Runnable disabledRunnable = new Runnable() {
+
+            @Override
+            public void run() {
+                AppCenterLog.debug(LOG_TAG, "Push token refreshed before start/enabled, ignoring.");
+            }
+        };
+        if (mContext == null) {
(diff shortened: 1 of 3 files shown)


sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -271,16 +271,27 @@ private void enqueuePushInstallationLog(@NonNull String pushToken) {
      */
     @SuppressWarnings("WeakerAccess") /* protected so that Xamarin can use it. */
     protected synchronized void onTokenRefresh(final String pushToken) {
```
MatkovIvan: @guperrot Should it still be `synchronized`?

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
```diff
@@ -271,16 +271,27 @@ private void enqueuePushInstallationLog(@NonNull String pushToken) {
...
+                }
+            };
             post(new Runnable() {
 
                 @Override
                 public void run() {
-                    enqueuePushInstallationLog(pushToken);
+                    if (pushToken != null && !pushToken.equals(mLatestPushToken)) {
+                        AppCenterLog.debug(LOG_TAG, "Push token refreshed: " + pushToken);
+                        mLatestPushToken = pushToken;
```
MatkovIvan: It isn't in `synchronized` anymore, is it OK?

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
guperrot: yes for mContext

sdk/appcenter-push/src/main/java/com/microsoft/appcenter/push/Push.java
guperrot: we access this variable always from the worker thread

PR #1244, opened 2019-09-06
Changes to support multiple selectable app secrets in Sasquatch

micarls:
## Description

In the process of adding AAD sign-in support to the SDK, we want the Sasquatch test app to be able to choose between AAD vs B2C app secret GUIDs and still support entering a custom GUID. This is intended to support that.

CHANGELOG.md is not modified because these changes do not affect the SDK itself, only the test app. It will be updated when AAD proper support is added to the SDK.


apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -494,23 +494,46 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
                     return true;
                 }
             });
-            initEditText(R.string.app_secret_key, R.string.app_secret_title, APP_SECRET_KEY, getString(R.string.app_secret), new EditTextListener() {
+            final String appSecret = MainActivity.sSharedPreferences.getString(APP_SECRET_KEY, "9e0d97c1-7838-46d0-9dab-1a0ef66aec6e");
```
guperrot: Do not hardcode it: getString(R.string.app_secret)

guperrot: And better, use the first value from the new array resource and remove the old single dimension resource.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -494,23 +494,46 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
...
+                    if (newValue == null) {
+                        return true;
+                    }
+                    String value = newValue.toString();
+
                     if (!TextUtils.isEmpty(value)) {
-                        setKeyValue(APP_SECRET_KEY, value);
-                        Toast.makeText(getActivity(), String.format(getActivity().getString(R.string.app_secret_changed_format), value), Toast.LENGTH_SHORT).show();
+                        if (value.equals("Custom")) {
+                            /* If user selected custom, popup a text editor so they can enter whatever they want */
```
guperrot:
```suggestion

                            /* If user selected custom, popup a text editor so they can enter whatever they want. */
```

guperrot: (newline, dot)

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -494,23 +494,46 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
...
+                            ShowPreferenceTextEditor(getPreferenceManager().findPreference(getString(R.string.app_secret_key)), R.string.app_secret_title, APP_SECRET_KEY, getString(R.string.app_secret), new EditTextListener() {
+                                @Override
```
guperrot:
```suggestion

                                @Override
```

Always a new line before annotations.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -740,32 +763,38 @@ private void initEditText(int key, final int title, final String preferencesKey,
...
             });
         }
 
+        private boolean ShowPreferenceTextEditor(final Preference preference, final int title, final String preferencesKey, final String defaultValue, final EditTextListener listener)
+        {
+            final EditText input = new EditText(getActivity());
+            input.setInputType(InputType.TYPE_CLASS_TEXT);
+            input.setText(MainActivity.sSharedPreferences.getString(preferencesKey, defaultValue));
+
+            /* Actually show the text entry dialog */
```
guperrot:
```suggestion
            /* Actually show the text entry dialog. */
```

apps/sasquatch/src/main/res/values/array.xml
```diff
@@ -12,4 +12,14 @@
         <item>NO_SECRET</item>
         <item>SKIP_START</item>
     </string-array>
+    <string-array name="appcenter_secret_values" tools:ignore="MissingTranslation">
+        <item>B2C</item>
+        <item>AAD</item>
+        <item>Custom</item>
+    </string-array>
+    <string-array name="appcenter_secret_guids" tools:ignore="MissingTranslation">
+        <item>9e0d97c1-7838-46d0-9dab-1a0ef66aec6e</item>
```
guperrot: This is the prod file, so prod values should be in this one.
There is another resource file for INT in projectDependency folder that needs different values for INT.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
```diff
@@ -384,6 +385,12 @@ public void accept(ErrorReport data) {
         listView.setOnItemClickListener(TestFeatures.getOnItemClickListener());
     }
 
+    /* Get the default app secret from the app secret array. */
+    protected static String GetDefaultAppSecret(Resources resources) {
```
guperrot: package-private is better (no keyword), protected is needed when subclassing.

Also this is a C# method name, while this is Java, it needs to be camelCase.

apps/sasquatch/src/main/res/values/array.xml
```diff
@@ -18,8 +18,8 @@
         <item>Custom</item>
     </string-array>
     <string-array name="appcenter_secret_guids" tools:ignore="MissingTranslation">
```
guperrot: I'd rather just call them appcenter_secrets, type or format does not have to be in the name of the key.

apps/sasquatch/src/main/res/values/array.xml
micarls: Thanks, fixed.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
micarls: Thanks, I've fixed this and also removed the now-duplicate "app_secret" resource

apps/sasquatch/src/main/res/values/array.xml
micarls: Thanks, fixed

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -494,23 +494,48 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
...
+                                    }
+                                }
+
+                                @Override
+                                public void onReset() {
+                                    setKeyValue(APP_SECRET_KEY, defaultAppSecret);
+                                    Toast.makeText(getActivity(), String.format(getActivity().getString(R.string.app_secret_changed_format), defaultAppSecret), Toast.LENGTH_SHORT).show();
+                                }
+                            });
+                        }
```
guperrot: Please reformat with Android Studio

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -740,32 +765,38 @@ private void initEditText(int key, final int title, final String preferencesKey,
...
-                            })
-                            .setNegativeButton(R.string.cancel, null)
-                            .create().show();
-                    return true;
+                    return ShowPreferenceTextEditor(preference, title, preferencesKey, defaultValue, listener);
                 }
             });
         }
 
+        private boolean ShowPreferenceTextEditor(final Preference preference, final int title, final String preferencesKey, final String defaultValue, final EditTextListener listener)
```
guperrot: This is a C# method name, but here it's a Java file. In Java method names are camelCase.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
```diff
@@ -494,23 +494,48 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
...
+                public boolean onPreferenceChange(Preference preference, Object newValue) {
+                    if (newValue == null) {
+                        return true;
+                    }
+                    String value = newValue.toString();
+
                     if (!TextUtils.isEmpty(value)) {
-                        setKeyValue(APP_SECRET_KEY, value);
-                        Toast.makeText(getActivity(), String.format(getActivity().getString(R.string.app_secret_changed_format), value), Toast.LENGTH_SHORT).show();
+                        if (value.equals("Custom")) {
```
guperrot: Might was well read the value from resource.

apps/sasquatch/src/projectDependency/res/values/array.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  ~ Copyright (c) Microsoft Corporation. All rights reserved.
+  ~ Licensed under the MIT License.
+  -->
+
+<resources xmlns:tools="http://schemas.android.com/tools">
+    <string-array name="appcenter_secrets" tools:ignore="MissingTranslation">
+        <item>9e0d97c1-7838-46d0-9dab-1a0ef66aec6e</item>
```
guperrot: Screen Shot 2019-09-09 at 2 12 04 PM

Same for the other variant.

Also please move this key to `env.xml` even if that one is an array, the files are actually not typed. `env.xml` is more convenient to have all the settings in 1 place.

micarls: Thanks. I've moved to env.xml, and it looks like now that it's in env.xml, TypographyDashes is already in the ignore list due to resources element having it.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
micarls: Good idea, now assuming it is last item in array

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
micarls: Fixed

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java
micarls: Thanks, fixed

apps/sasquatch/src/projectDependency/res/values/env.xml
```diff
@@ -5,13 +5,19 @@
   -->
 
 <resources xmlns:tools="http://schemas.android.com/tools" tools:ignore="TypographyDashes">
-    <string name="app_secret" tools:ignore="MissingTranslation">9e0d97c1-7838-46d0-9dab-1a0ef66aec6e</string>
     <string name="log_url" tools:ignore="MissingTranslation">https://in-integration.dev.avalanch.es</string>
     <string name="log_url_one_collector" tools:ignore="MissingTranslation">https://mobile.events.data.microsoft.com/OneCollector/1.0</string>
     <string name="install_url" tools:ignore="MissingTranslation">https://install.portal-server-core-integration.dev.avalanch.es</string>
     <string name="api_url" tools:ignore="MissingTranslation">https://api-gateway-core-integration.dev.avalanch.es/v0.1</string>
-    <string name="auth_config_url" tools:ignore="MissingTranslation">https://config-integration.dev.avalanch.es</string>
+    <string name="auth_config_url" tools:ignore="MissingTranslation">https://mbaasapi0fde96d9b.blob.core.windows.net</string>
```
micarls: Oops, fixing this now.

guperrot:
This bug is still happening: https://github.com/microsoft/appcenter-sdk-android/pull/1244#pullrequestreview-285738638

guperrot:
We also need to change the label for the signIn button to remove "B2C" and provide instead a text that will work for both aad and b2c.

micarls:
Thanks Guillaume, I did miss those comments in my changes yesterday. I believe the PR addresses all existing feedback now.

PR #1247, opened 2019-09-09
BYOI setAuthToken

guperrot:

* [x] Did you add unit tests?

## Description

Add AppCenter.setAuthToken

## Related PRs or issues

[AB#67952](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/67952)

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
@@ -1192,7 +1192,7 @@ public class AppCenter {
             claims = JwtClaims.parse(authToken);
         }
         if (claims != null) {
-            AppCenterLog.debug(LOG_TAG, "Token has been refreshed.");
+            AppCenterLog.debug(LOG_TAG, "Authentication token has been refreshed.");
             authTokenContext.setAuthToken(authToken, claims.getSubject(), claims.getExpirationDate());
         } else {
             AppCenterLog.debug(LOG_TAG, "Removing authentication token (sign out).");
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
@@ -454,6 +454,18 @@ public class AppCenter {
     }
 
     /**
+     * Set up App Center to identify the current user based on a authentication token.
+     * The authentication token is used in all logs sent by App Center and can be used in specific features by other modules.
+     * When the authentication token is about to expire or is expired, App Center will attempt to
+     * refresh it using the {@link AuthTokenListener} passed in {@link #setAuthTokenListener(AuthTokenListener)}.
+     *
+     * @param authToken JWT identifying the user or null to remove authentication.
+     */
+    public static void setAuthToken(String authToken) {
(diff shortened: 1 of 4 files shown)


sdk/appcenter/src/test/java/com/microsoft/appcenter/AppCenterAuthTest.java
```diff
@@ -42,6 +42,44 @@ public void setUp() throws Exception {
...
+    }
+
+    @Test
+    public void unsetAuthToken() {
+
+        /* When we unset auth token. */
+        AppCenter.setAuthToken(null);
+
+        /* Then it's removed. */
+        verify(mAuthTokenContext).setAuthToken(null, null, null);
```
MatkovIvan: And `JwtClaims.parse` isn't called.

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -1170,6 +1182,24 @@ public void run() {
...
+    /**
+     * Implements {@link #setAuthToken} at instance level.
+     */
+    private void setInstanceAuthToken(String authToken) {
+        AuthTokenContext authTokenContext = AuthTokenContext.getInstance();
+        JwtClaims claims = null;
+        if (authToken != null) {
+            claims = JwtClaims.parse(authToken);
+        }
+        if (claims != null) {
```
jaelim-ms: You could invert the condition.

sdk/appcenter/src/test/java/com/microsoft/appcenter/AppCenterAuthTest.java
jaelim-ms: I think it is fine but we can also verify that isn't called.

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
jwallra: I prefer this organization, but I suppose it's a stylistic choice

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -1170,6 +1182,24 @@ public void run() {
...
+            AppCenterLog.debug(LOG_TAG, "Token has been refreshed.");
```
jwallra: To match the other log in the method

```suggestion
            AppCenterLog.debug(LOG_TAG, "Authentication token has been refreshed.");
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
jaelim-ms: We can get rid of unnecessary one bit operation there. 😄

guperrot: Not really the logging matters here.
If you pass null we just log that you remove the token, it's not an error.
If you pass an invalid JWT, we log an error and utils AND we remove the token.

guperrot: It's more common to have the success block before the failure one in a if/else. Both work.

PR #1255, opened 2019-09-16
[Android Q] Distribution code deprecation warning

russelarms:

* [ ] Did you add unit tests?

## Description

As part of being up-to-date, we should use the latest APIs for distribute before they are removed in future versions of Android (maybe Android 11 will remove the old one?).

https://developer.android.com/reference/android/content/Intent.html#ACTION_INSTALL_PACKAGE is deprecated.

Starting API level 21 we should have started using https://developer.android.com/reference/android/content/pm/PackageInstaller.html.

Since we are compatible with API level min 16 we need to conditionally use PackageInstaller on API level 21+ (or at least 29+ if issues arise between 21-28 range) and fallback to the old code for older platforms.

## Related PRs or issues

[AB#67823](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/67823)

### DIFF
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
@@ -7,14 +7,17 @@ package com.microsoft.appcenter.sasquatch.activities;
 
 import android.annotation.SuppressLint;
 import android.app.Application;
-import android.content.res.Resources;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.content.pm.PackageInstaller;
+import android.content.res.Resources;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.StrictMode;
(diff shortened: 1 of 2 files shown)


apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
```diff
@@ -225,10 +228,48 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            handlePackageUpdate(intent);
+        }
         Log.d(LOG_TAG, "onNewIntent triggered");
         Push.checkLaunchedFromNotification(this, intent);
     }
 
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    protected void handlePackageUpdate(Intent intent) {
+        Bundle extras = intent.getExtras();
```
guperrot: That should be a method that our SDK provides.

guperrot:
Let's close it for now.

PR #1256, opened 2019-09-16
[Android] Data module hangs when no actual internet connection

nikitag-ms:
* [ ] Did you add unit tests?

## Description

[Android] Data module hangs when no actual internet connection. The idea (at least at first) is not to solve the problem completely, but at least to eliminate hanging (essentially, make it on par with what Auth does now). I've spent time trying to find a less risky way to do an actual solution (for example, to "go offline" when no internet), but couldn't find a good way without having to rewrite a good portion of the Data module. At least starting from something like this would help while we are working on a more sophisticated approach.

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientRetryer.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientRetryer.java
@@ -30,20 +30,10 @@ public class HttpClientRetryer extends HttpClientDecorator {
      * Retry intervals to use, array index is to use the value for each retry. When we used all the array values, we give up and forward the last error.
      */
     @VisibleForTesting
-    static final long[] GENERAL_RETRY_INTERVALS = new long[]{
+    static final long[] RETRY_INTERVALS = new long[]{
             TimeUnit.SECONDS.toMillis(10),
             TimeUnit.MINUTES.toMillis(5),
             TimeUnit.MINUTES.toMillis(20)
-    };
-
-    /**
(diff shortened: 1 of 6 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/http/HttpClientRetryer.java
```diff
@@ -44,13 +46,15 @@
...
+    private final boolean mRetryOnNetworkingFailures;
+
     /**
      * Init with default retry policy.
      *
      * @param decoratedApi API to decorate.
      */
-    HttpClientRetryer(HttpClient decoratedApi) {
-        this(decoratedApi, new Handler(Looper.getMainLooper()));
+    HttpClientRetryer(boolean retryOnNetworkingFailures, HttpClient decoratedApi) {
```
guperrot: But now if our own endpoint is down, this will still hang for a long time.
Do we really need to retry anything in Data if we want to fail fast and go offline mode?

sdk/appcenter/src/main/java/com/microsoft/appcenter/http/TLS1_2SocketFactory.java
```diff
@@ -32,6 +33,11 @@
      */
     private static final String[] ENABLED_PROTOCOLS = { TLS1_2_PROTOCOL };
 
+    /**
+     * Socket timeout
+     */
+    private static final int SO_TIMEOUT = 5000;
```
guperrot: this is not the right place, this code is executed only on specific versions of Android.
Please look where we actually configure the url connection, we already have constants for connect and read timeouts.

PR #1258, opened 2019-09-19
Use separate HTTP clients for online and offline operations

nikitag-ms:
* [x] Did you add unit tests?

## Description

Use separate HTTP clients for online and offline operations. This is to fail fast on online operations and still keep retrying on offline ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # App Center SDK for Android Change Log
 
-## Version 2.3.1 (Under development)
+## Version 2.3.1
+
+### App Center Data
+
+* **[Fix]** Reduced retries on Data-related operations to fail fast and avoid the perception of calls "hanging".
 
 ___
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # App Center SDK for Android Change Log
 
-## Version 2.3.1
+## Version 2.3.1 (Under development)
 
 ### App Center Data
 
--- a/sdk/appcenter-data/src/test/java/com/microsoft/appcenter/data/AbstractDataTest.java
+++ b/sdk/appcenter-data/src/test/java/com/microsoft/appcenter/data/AbstractDataTest.java
@@ -18,6 +18,7 @@ import com.microsoft.appcenter.data.client.TokenExchange;
 import com.microsoft.appcenter.data.models.RemoteOperationListener;
 import com.microsoft.appcenter.http.AbstractAppCallTemplate;
 import com.microsoft.appcenter.http.HttpClient;
+import com.microsoft.appcenter.http.HttpClientNetworkStateHandler;
 import com.microsoft.appcenter.http.HttpClientRetryer;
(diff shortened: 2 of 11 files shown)


sdk/appcenter-data/src/test/java/com/microsoft/appcenter/data/AbstractDataTest.java
```diff
@@ -317,13 +326,16 @@ String verifyCosmosDbFlow(String documentId, String cosmosCallApiMethod, String
...
+            boolean retryHttpCalls, String tokenExchangeSuccessResponsePayload,
             Exception tokenExchangeFailureResponse) throws JSONException {
         ArgumentCaptor<AbstractAppCallTemplate> tokenExchangeTemplateCallbackArgumentCaptor =
                 ArgumentCaptor.forClass(AbstractAppCallTemplate.class);
         ArgumentCaptor<TokenExchange.TokenExchangeServiceCallback> tokenExchangeServiceCallbackArgumentCaptor =
                 ArgumentCaptor.forClass(TokenExchange.TokenExchangeServiceCallback.class);
-        verify(mHttpClient).callAsync(
+        HttpClient callHttpClient = retryHttpCalls ? mHttpClientWithRetryer : mHttpClientNoRetryer;
+        HttpClient unusedHttpClient = !retryHttpCalls ? mHttpClientWithRetryer : mHttpClientNoRetryer;
+        verifyZeroInteractions(unusedHttpClient);
```
guperrot: In the future prefer verify never on a specific call.

sdk/appcenter-data/src/test/java/com/microsoft/appcenter/data/DataTest.java
```diff
@@ -276,23 +278,23 @@ public void shouldNotOperateOnLocalStorageWhenPartitionNotExist() {
         /* Local storage create document should complete with not find partition error. */
         assertNotNull(doc);
         assertNotNull(doc.getError());
-        assertTrue(doc.getError().getMessage().contains(failedMessage));
+        assertTrue(Objects.requireNonNull(doc.getError().getMessage()).contains(failedMessage));
```
guperrot:
```
    public static <T> T requireNonNull(T obj) {
        if (obj == null)
            throw new NullPointerException();
        return obj;
    }
```

That's what `Objects.requireNonNull` does so it does not make the test easy to read on failure for null (failed with exception instead of assert).

Prefer assertNotNull on previous line.

guperrot:
Besides the 2 comments above, looks good.

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # App Center SDK for Android Change Log
 
-## Version 2.3.1 (Under development)
+## Version 2.3.1
```
guperrot: Do not remove (Under development), release team will do.
Since the changelog is visible on develop and that is the default branch it could give inaccurate info that the version is available.

guperrot:
```suggestion
## Version 2.3.1 (Under development)
```

PR #1262, opened 2019-09-26
Fix issue where auth tokens were refreshed multiple times

achocron:

* [x] Did you add unit tests?

Fixes an issue where a single token can be refreshed multiple times.

Sorry about the messy history- accidentally pushed to feature branch at first, then reverted, and this pr reverts that revert.

[AB#69442](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/69442)

### DIFF
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
@@ -123,10 +123,8 @@ public class AuthTokenContext {
     /**
      * Set or unset the refresh token update listener. There can be only 1 active.
      */
-    public synchronized void setRefreshListener(@NonNull RefreshListener refreshListener) {
-        if (mRefreshListener.compareAndSet(null, refreshListener)) {
-            mLastTokenRefreshed = null;
-        } else {
+    public void setRefreshListener(@NonNull RefreshListener refreshListener) {
+        if (!mRefreshListener.compareAndSet(null, refreshListener)) {
             AppCenterLog.error(LOG_TAG, "Cannot use 2 authentication modules at the same time.");
         }
     }
@@ -359,21 +357,25 @@ public class AuthTokenContext {
      * @param authTokenInfo auth token to check for expiration.
      */
     public void checkIfTokenNeedsToBeRefreshed(@NonNull AuthTokenInfo authTokenInfo) {
-        RefreshListener refreshListener;
-        AuthTokenHistoryEntry lastEntry;
-        synchronized(this) {
-            lastEntry = getLastHistoryEntry();
-            if (lastEntry == null || authTokenInfo.getAuthToken() == null ||
-                    !authTokenInfo.getAuthToken().equals(lastEntry.getAuthToken()) ||
-                    !authTokenInfo.isAboutToExpire() || authTokenInfo.getAuthToken().equals(mLastTokenRefreshed)) {
-                return;
+        String accountId = getAccountIdToRefreshAndUpdateCache(authTokenInfo);
+        if (accountId != null) {
+            RefreshListener refreshListener = mRefreshListener.get();
+            if (refreshListener != null) {
+                refreshListener.onTokenRequiresRefresh(accountId);
             }
-            mLastTokenRefreshed = authTokenInfo.getAuthToken();
         }
-        refreshListener = mRefreshListener.get();
-        if (refreshListener != null) {
-            refreshListener.onTokenRequiresRefresh(lastEntry.getHomeAccountId());
+    }
+
+    private synchronized String getAccountIdToRefreshAndUpdateCache(@NonNull AuthTokenInfo authTokenInfo) {
+        AuthTokenHistoryEntry lastEntry = getLastHistoryEntry();
+        if (lastEntry == null || authTokenInfo.getAuthToken() == null ||
+                !authTokenInfo.getAuthToken().equals(lastEntry.getAuthToken()) ||
+                !authTokenInfo.isAboutToExpire() ||
+                authTokenInfo.getAuthToken().equals(mLastTokenRefreshed)) {
+            return null;
         }
+        mLastTokenRefreshed = authTokenInfo.getAuthToken();
+        return lastEntry.getHomeAccountId();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -118,8 +123,10 @@ public static synchronized void unsetInstance() {
     /**
      * Set or unset the refresh token update listener. There can be only 1 active.
      */
-    public void setRefreshListener(@NonNull RefreshListener refreshListener) {
-        if (!mRefreshListener.compareAndSet(null, refreshListener)) {
+    public synchronized void setRefreshListener(@NonNull RefreshListener refreshListener) {
+        if (mRefreshListener.compareAndSet(null, refreshListener)) {
+            mLastTokenRefreshed = null;
```
guperrot: I don't understand why we need to reset that when setting up the listener the first time. Plus it introduces synchronization on a path where use another concurrency mechanism instead.

achocron: I'm less sure of it now than I was before. But the thought is to handle the scenario in `sameAuthTokenRefreshedAgainIfNoListenerOnFirstTry`

achocron:
Don't merge, realized there is a bug.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
guperrot: Synced offline, this is not correct the changes in this function.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -352,16 +357,25 @@ public synchronized void removeOldestTokenIfMatching(String token) {
      * @param authTokenInfo auth token to check for expiration.
      */
     public void checkIfTokenNeedsToBeRefreshed(@NonNull AuthTokenInfo authTokenInfo) {
+        String accountId = getAccountIdToRefreshAndUpdateCache(authTokenInfo);
+        if (accountId != null) {
+            RefreshListener refreshListener = mRefreshListener.get();
+            if (refreshListener != null) {
```
MatkovIvan: Please check this at the first place to make sure that any listener will handle already saved `mLastTokenRefreshed` value

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -352,16 +357,25 @@ public synchronized void removeOldestTokenIfMatching(String token) {
...
-            return;
-        }
-        RefreshListener refreshListener = mRefreshListener.get();
-        if (refreshListener != null) {
-            refreshListener.onTokenRequiresRefresh(lastEntry.getHomeAccountId());
+                !authTokenInfo.isAboutToExpire() ||
+                authTokenInfo.getAuthToken().equals(mLastTokenRefreshed)) {
+            return null;
         }
+        mLastTokenRefreshed = authTokenInfo.getAuthToken();
```
guperrot: We need to reset this when removing listener.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -352,16 +358,25 @@ public synchronized void removeOldestTokenIfMatching(String token) {
...
+        RefreshListener refreshListener = mRefreshListener.get();
+        if (refreshListener != null) {
+            String accountId = getAccountIdToRefreshAndUpdateCache(authTokenInfo);
+            if (accountId != null) {
+                refreshListener.onTokenRequiresRefresh(accountId);
+            }
+        }
+    }
+
+    private synchronized String getAccountIdToRefreshAndUpdateCache(@NonNull AuthTokenInfo authTokenInfo) {
```
jaelim-ms: Any reason of extracting this as a method? I would rather combine two methods to improve readability.
`getAccountIdToRefreshAndUpdateCache` method doesn't need to return a value (makes more sense to return boolean instead of String), the method is more likely detecting a token that is required to be update so returning null for accountId would mislead the behavior of the method.

guperrot: It's for synchronized, need to be at method level for code coverage.

jaelim-ms:
```suggestion
    /**
     * Check if the last/current token needs a refresh because expiring soon.
     *
     * @param authTokenInfo Auth token to check.
     * @return Home accountId with last token about to expire or already expired.
     */
    private synchronized String getHomeAccountIdToRefreshToken(@NonNull AuthTokenInfo authTokenInfo) {
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
```diff
@@ -352,16 +358,25 @@ public synchronized void removeOldestTokenIfMatching(String token) {
      * @param authTokenInfo auth token to check for expiration.
      */
     public void checkIfTokenNeedsToBeRefreshed(@NonNull AuthTokenInfo authTokenInfo) {
+        RefreshListener refreshListener = mRefreshListener.get();
+        if (refreshListener != null) {
+            String accountId = getAccountIdToRefreshAndUpdateCache(authTokenInfo);
```
jaelim-ms:
```suggestion
            String accountId = getHomeAccountIdToRefreshToken(authTokenInfo);
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
jaelim-ms: @guperrot's suggestion. Took it.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/context/AuthTokenContext.java
jaelim-ms: @guperrot's suggestion. Took it.

PR #1295, opened 2019-10-24
Handled ErrorReport for wrapper SDKs

guperrot:

* [x] Did you add unit tests?

## Description

Expose utilities for Unity SDK.

## Related PRs or issues

[AB#71515](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/71515)

### DIFF
--- a/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
+++ b/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
@@ -575,7 +575,6 @@ public class Crashes extends AbstractAppCenterService {
         };
     }
 
-
     synchronized Device getDeviceInfo(Context context) throws DeviceInfoHelper.DeviceInfoException {
         if (mDevice == null) {
             mDevice = DeviceInfoHelper.getDeviceInfo(context);
--- a/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
+++ b/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
@@ -29,6 +29,7 @@ import com.microsoft.appcenter.crashes.model.ErrorReport;
 import com.microsoft.appcenter.crashes.model.NativeException;
 import com.microsoft.appcenter.crashes.model.TestCrashException;
 import com.microsoft.appcenter.crashes.utils.ErrorLogHelper;
+import com.microsoft.appcenter.ingestion.models.Device;
 import com.microsoft.appcenter.ingestion.models.Log;
 import com.microsoft.appcenter.ingestion.models.json.DefaultLogSerializer;
 import com.microsoft.appcenter.ingestion.models.json.LogFactory;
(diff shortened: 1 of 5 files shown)


sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -572,12 +575,19 @@ public void onCallBack(ErrorReport report) {
         };
     }
 
+
```
guperrot:
```suggestion
```

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/WrapperSdkExceptionManager.java
```diff
@@ -5,22 +5,27 @@
...
 
 import java.io.File;
 import java.util.Collection;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
 
+import static com.microsoft.appcenter.crashes.Crashes.LOG_TAG;
+
```
guperrot:
```suggestion
```

guperrot:
About test app: I tested breakpad still worked (code was impacted with the getDeviceInfo).

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -616,7 +626,11 @@ public Exception buildExceptionModel() {
         }, properties, attachments);
     }
 
-    private synchronized void queueException(@NonNull final ExceptionModelBuilder exceptionModelBuilder, Map<String, String> properties, final Iterable<ErrorAttachmentLog> attachments) {
+    private synchronized UUID queueException(@NonNull final ExceptionModelBuilder exceptionModelBuilder, Map<String, String> properties, final Iterable<ErrorAttachmentLog> attachments) {
+
+        /* Snapshot userId as early as possible. */
```
dhei: Can you explain why do you want to do it as early as possible?

guperrot: Because post happens on background thread and userId could change.
We fixed something similar in 1DS long time ago.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -250,7 +256,6 @@ static synchronized void unsetInstance() {
      * @param throwable An exception.
      */
     @SuppressWarnings({"SameParameterValue", "WeakerAccess"})
-    // TODO remove the suppress when releasing to jCenter
```
michaelsmitha: I see you remove the suppress on the `trackException` below but leave it here. why?

guperrot: Because we don't use this signature in test app so it will always warn.
The other one, Android Studio is smart enough to detect we use it via reflection so it was warning about redundant suppression.

PR #1296, opened 2019-10-25
We never released trackException in Java, it's trackError in .NET

guperrot:

* [ ] Did you add unit tests?

## Description

When we did trackException work for Xamarin, PMs wanted to name it trackError.
To this day this is the name of the concept: "Error" in the Portal.
Java work has always been hidden before.

### DIFF
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/ManagedErrorActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/ManagedErrorActivity.java
@@ -71,17 +71,14 @@ public class ManagedErrorActivity extends PropertyActivity {
         } catch (Throwable t) {
             try {
                 Map<String, String> properties = readStringProperties();
-
-                @SuppressWarnings("JavaReflectionMemberAccess")
-                Method method = Crashes.class.getDeclaredMethod("trackException", Throwable.class, Map.class, Iterable.class);
-                method.setAccessible(true);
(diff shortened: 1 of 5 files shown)


apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/ManagedErrorActivity.java
```diff
@@ -71,17 +71,14 @@ protected void send(View view) {
         } catch (Throwable t) {
             try {
                 Map<String, String> properties = readStringProperties();
-
-                @SuppressWarnings("JavaReflectionMemberAccess")
-                Method method = Crashes.class.getDeclaredMethod("trackException", Throwable.class, Map.class, Iterable.class);
-                method.setAccessible(true);
+                Method method = Crashes.class.getMethod("trackError", Throwable.class, Map.class, Iterable.class);
```
guperrot: Before trackException would work in jCenter with name and properties via reflecting on a private API.

So the accessible trick was already broken after adding attachments (signature changed), now it's even worse by renaming.

Just remove that and we will need jCenter update.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -251,27 +251,28 @@ static synchronized void unsetInstance() {
     }
 
     /**
-     * Track an exception.
+     * Track a handled error.
      *
-     * @param throwable An exception.
+     * @param throwable The throwable describing the handled error.
```
guperrot: Jadadoc update based on Xamarin one.

PR #1300, opened 2019-10-30
Update issue templates

guperrot:
[AB#71624](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/71624)

Since we apply bug label only after we confirm, I didn't name that bug report but problem report with support label by default.


guperrot:
Tested by changing default branch to this one.

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
```
dhei: Maybe more specific?

```suggestion
about: Suggest an idea for App Center Android SDK
```

Issue #1301, opened 2019-10-31
In-app update only works once, after that the same version is installed again and again


thhook:
### **Description**

The in-app update only works once. After that it keeps installing the same version over and over again without downloading further releases.

### **Repro Steps**

1. Download and install an app with enabled AppCenter distribution
2. Add a newer release of the app in AppCenter
3. Open the app on the device: an update dialog is shown
4. Download the new release and install it when finished
5. Add another newer release in AppCenter
6. Open the app again: an update dialog with the correct new release version is shown. Yet when clicking on "download" it always reuses the already installed current version, which leads to a kind of "endless loop"

### **Details**

1. Which SDK version are you using?
    - 2.4.1
2. Which OS version did you experience the issue on?
    - Android 9
3. What device version did you see this error on?  Were you using an emulator or a physical device?
    - Samsung Galaxy S8+ physical device
4. What third party libraries are you using?
 
5. Please enable verbose logging for your app using `AppCenter.setLogLevel(Log.VERBOSE)` before your call to `AppCenter.start(...)` and include the logs here:

    - First app session, no updates available
```
2019-10-31 15:02:24.099 5283-5283/com.bmw.CNRDAppCenterTest D/AppCenterDistribute: InstallerPackageName=com.google.android.packageinstaller
2019-10-31 15:02:24.145 5283-5283/com.bmw.CNRDAppCenterTest D/AppCenterDistribute: No token, need to open browser to url=https://install.appcenter.ms/apps/<omitted>/update-setup/?release_hash=c943efaf261c54447fd5281e53707efb1b58bed14e88141ec06606ed3c13e462&redirect_id=com.bmw.CNRDAppCenterTest&redirect_scheme=appcenter&request_id=a85c0e43-cd64-4b17-bd00-eff0dbb89836&platform=Android&enable_failure_redirect=true&install_id=605cf24e-9445-489c-8dbc-b378d5e67068
2019-10-31 15:02:24.157 5283-5283/com.bmw.CNRDAppCenterTest D/AppCenterDistribute: Default browser seems to be com.android.chrome/com.google.android.apps.chrome.IntentDispatcher
2019-10-31 15:02:24.157 5283-5283/com.bmw.CNRDAppCenterTest D/AppCenterDistribute: And its not the picker.
[...]

PR #1303, opened 2019-11-01
Fix in-app update cache issue

guperrot:

* [x] Did you add unit tests?

## Description

The way we cancel previous downloads since version 2.4.0 is by using the release cache.
If the download was completed successfully, we were not loading the release cache after installing the update (from version A to B), that step was skipped because before version 2.4.0 there was no reason to load release details in that case and the code was not updated.
Thus when receiving the update from B to C, release cache was not pruned because not loaded so the comparison/cancel was skipped in `updateReleaseDetails()`.

## Related PRs or issues

#1301 
[AB#71993](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/71993)

### DIFF
--- a/sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeTest.java
+++ b/sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeTest.java
@@ -579,6 +579,27 @@ public class DistributeTest extends AbstractDistributeTest {
     }
 
     @Test
+    public void loadCacheReleaseDetailsIfProcessRestarted() {
+
+        /* Mock mReleaseDownloader is downloading. */
+        when(mReleaseDownloader.isDownloading()).thenReturn(true);
+        mockStatic(DistributeUtils.class);
+        when(DistributeUtils.getStoredDownloadState()).thenReturn(DOWNLOAD_STATE_COMPLETED);
+
+        /* Mock that download time is bigger than packageInfo.lastUpdateTime. */
+        when(SharedPreferencesManager.getLong(eq(PREFERENCE_KEY_DOWNLOAD_TIME))).thenReturn(3L);
+
+        /* mReleaseDetails is not null and it's a mandatory update. */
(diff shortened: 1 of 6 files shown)


sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeTest.java
```diff
@@ -578,6 +578,27 @@ private void resumeWorkflow(Activity activity) {
...
+    @Test
+    public void loadCacheReleaseDetailsIfProcessRestarted() {
+
+        /* Mock mReleaseDownloader is downloading. */
+        when(mReleaseDownloader.isDownloading()).thenReturn(true);
+        mockStatic(DistributeUtils.class);
+        when(DistributeUtils.getStoredDownloadState()).thenReturn(DOWNLOAD_STATE_COMPLETED);
+
+        /* Mock that download time is bigger than packageInfo.lastUpdateTime. */
+        when(SharedPreferencesManager.getLong(eq(PREFERENCE_KEY_DOWNLOAD_TIME))).thenReturn(3L);
```
guperrot: We don't hit that code when update is not mandatory we need another test too.

guperrot:
Found a problem.

guperrot:
There is a side effect with this fix where would no longer check for releases if the last downloaded one is optional and not yet installed.

guperrot:
Still does not work e2e for some reason.

guperrot:
Optional update scenario verified e2e with Android 10 with 04af510.
Need to verify Android 4.x and mandatory updates and disable/reenable e2e to be sure.

guperrot:
Cache is lost by disabling SDK which can cause the same issue.

guperrot:
Disabling and mandatory verified e2e on cf2602b.
Now need to retest everything on Android 4.x to be sure.

guperrot:
Android 4.x tested successfully.

PR #1345, opened 2020-01-08
Fix possible deadlock when activity resumed during SDK background init

MatkovIvan:
* [x] Has `CHANGELOG.md` been updated?
* [x] Are tests passing locally?
* [x] Are the files formatted correctly?
* [ ] Did you add unit tests?
* [ ] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?
* [x] Did you check UI tests on the sample app? They are not executed on CI.

## Description

Android SDK can deadlock with distribute while opening browser if the SDK background thread is initializing. The cause is code using `AppCenter.getInstallId.get()` inside `DistributeUtils` to open browser instead of using asynchronous version (`thenAccept`).

## Related PRs or issues

[AB#74885](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/74885)

PR #1349, opened 2020-01-10
Fix possible deadlock when activity resumed during SDK background init

MatkovIvan:
* [x] Did you add unit tests?

## Description

Android SDK can deadlock with distribute while opening browser if the SDK background thread is initializing. The cause is code using `AppCenter.getInstallId.get()` inside `DistributeUtils` to open browser instead of using asynchronous version (`thenAccept`).

## Related PRs or issues

#1345
[AB#74885](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/74885)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * **[Fix]** Avoid opening browser to check for sign-in information after receiving an SSL error while checking for app updates (which often happens when using a public WIFI).
 * **[Fix]** When in-app update permissions become invalid and need to open browser again, updates are no longer postponed after sign-in (if user previously selected the action to postpone for a day).
+* **[Fix]** Fix a possible deadlock when activity resumes during background operation for `Distribute.isEnabled()` call.
 
 ___
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 
 * **[Fix]** Avoid opening browser to check for sign-in information after receiving an SSL error while checking for app updates (which often happens when using a public WIFI).
 * **[Fix]** When in-app update permissions become invalid and need to open browser again, updates are no longer postponed after sign-in (if user previously selected the action to postpone for a day).
-* **[Fix]** Fix a possible deadlock when activity resumes during background operation for `Distribute.isEnabled()` call.
+* **[Fix]** Fix a possible deadlock when activity resumes during background operation for some `Distribute` public APIs like `Distribute.isEnabled()`.
 
 ___
 
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -507,7 +507,7 @@ public class Distribute extends AbstractAppCenterService {
                     }
                 });
             } else {
-                AppCenterLog.verbose(LOG_TAG, "Distribute workflow will be resumed on activity resume event.");
+                AppCenterLog.debug(LOG_TAG, "Distribute workflow will be resumed on activity resume event.");
             }
         } else {
 
@@ -631,7 +631,7 @@ public class Distribute extends AbstractAppCenterService {
      */
     @UiThread
     private synchronized void resumeDistributeWorkflow() {
-        AppCenterLog.verbose(LOG_TAG, "Resume distribute workflow...");
+        AppCenterLog.debug(LOG_TAG, "Resume distribute workflow...");
         if (mPackageInfo != null && mForegroundActivity != null && !mWorkflowCompleted && isInstanceEnabled()) {
 
             /* Don't go any further it this is a debug app. */
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -34,7 +34,6 @@ import android.text.TextUtils;
 import android.widget.Toast;
 
 import com.microsoft.appcenter.AbstractAppCenterService;
-import com.microsoft.appcenter.AppCenter;
 import com.microsoft.appcenter.Flags;
 import com.microsoft.appcenter.channel.Channel;
 import com.microsoft.appcenter.distribute.channel.DistributeInfoTracker;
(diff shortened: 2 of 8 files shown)


sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/DistributeUtils.java
```diff
@@ -138,7 +138,7 @@ static void updateSetupUsingBrowser(Activity activity, String installUrl, String
         url += "&" + PARAMETER_REQUEST_ID + "=" + requestId;
         url += "&" + PARAMETER_PLATFORM + "=" + PARAMETER_PLATFORM_VALUE;
         url += "&" + PARAMETER_ENABLE_UPDATE_SETUP_FAILURE_REDIRECT_KEY + "=" + "true";
-        url += "&" + PARAMETER_INSTALL_ID + "=" + AppCenter.getInstallId().get().toString();
+        url += "&" + PARAMETER_INSTALL_ID + "=" + IdHelper.getInstallId().toString();
```
guperrot: This will crash in the work around where resume method is called after start if the SDK background thread that initialize storage is not fast enough.

That's also a reason why we should use thenAccept.

MatkovIvan: No, it cannot be called before storage initialization.
See https://github.com/microsoft/appcenter-sdk-android/blob/9f7aff78841cf769ccef330926a48d5af5961540/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java#L475

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/IdHelper.java
```diff
@@ -22,7 +22,7 @@
      * @return the installID
      */
     @NonNull
-    public static UUID getInstallId() {
+    public static synchronized UUID getInstallId() {
```
guperrot: No need for synchronized it as initialization is already guaranteed by the time distribute calls this. InstallId is called in finishConfiguration by core, which is before Distribute can get a reference to the channel. If no reference to channel then resume does nothing and is replayed by the applyEnabledState later.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -925,6 +932,8 @@ synchronized void storeUpdateSetupFailedParameter(@NonNull String requestId, @No
         } else if (requestId.equals(SharedPreferencesManager.getString(PREFERENCE_KEY_REQUEST_ID))) {
             AppCenterLog.debug(LOG_TAG, "Stored update setup failed parameter.");
             SharedPreferencesManager.putString(PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY, updateSetupFailed);
+        } else {
+            AppCenterLog.warn(LOG_TAG, "Ignoring redirection parameters as requestId is invalid.");
```
guperrot: Unrelated to the bug and it should be debug.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -941,6 +950,8 @@ synchronized void storeTesterAppUpdateSetupFailedParameter(@NonNull String reque
         } else if (requestId.equals(SharedPreferencesManager.getString(PREFERENCE_KEY_REQUEST_ID))) {
             AppCenterLog.debug(LOG_TAG, "Stored tester app update setup failed parameter.");
             SharedPreferencesManager.putString(PREFERENCE_KEY_TESTER_APP_UPDATE_SETUP_FAILED_MESSAGE_KEY, testerAppUpdateSetupFailed);
+        } else {
+            AppCenterLog.warn(LOG_TAG, "Ignoring redirection parameters as requestId is invalid.");
```
guperrot: Unrelated to the bug and it should be debug.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -625,6 +631,7 @@ private synchronized void cancelPreviousTasks() {
      */
     @UiThread
     private synchronized void resumeDistributeWorkflow() {
+        AppCenterLog.verbose(LOG_TAG, "Resume distribute workflow...");
```
guperrot: debug, verbose is only for protocol logging like HTTP.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -496,13 +496,19 @@ protected synchronized void applyEnabledState(boolean enabled) {
...
             mChannel.addListener(mDistributeInfoTracker);
-            HandlerUtils.runOnUiThread(new Runnable() {
 
-                @Override
-                public void run() {
-                    resumeDistributeWorkflow();
-                }
-            });
+            /* Resume distribute workflow only if there is foreground activity. */
+            if (mForegroundActivity != null) {
```
guperrot: We don't need that if, I supposed that's an optimization? Unrelated to this bug though.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
MatkovIvan: Related to deduplication part. No, it's warn in another place, it should be consistent

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
MatkovIvan: Related to deduplication part. No, it's warn in another place, it should be consistent

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 * **[Fix]** Avoid opening browser to check for sign-in information after receiving an SSL error while checking for app updates (which often happens when using a public WIFI).
 * **[Fix]** When in-app update permissions become invalid and need to open browser again, updates are no longer postponed after sign-in (if user previously selected the action to postpone for a day).
+* **[Fix]** Fix a possible deadlock when activity resumes during background operation for `Distribute.isEnabled()` call.
```
guperrot: Isn't setEnabled or notifyUserConfirmation also impacted? Should be more generic in that case.

MatkovIvan: `notifyUserConfirmation` calls `isInstanceEnabledAsync` internally, so it's the same.

MatkovIvan:
```suggestion
* **[Fix]** Fix a possible deadlock when activity resumes during background operation for some `Distribute` public APIs like `Distribute.isEnabled()`.
```

PR #1353, opened 2020-01-17
Set/get in update track

guperrot:

* [x] Did you add unit tests?

## Description

See [AB#75285](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/75285)

### DIFF
--- a/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventActivity.java
+++ b/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventActivity.java
@@ -188,7 +188,7 @@ public class EventActivity extends PropertyActivity {
     }
 
     @Override
-    protected void send(View view) {
+    public void send(View view) {
         AnalyticsTransmissionTarget target = getSelectedTarget();
         PersistenceFlag persistenceFlag = PersistenceFlag.values()[mPersistenceFlagSpinner.getSelectedItemPosition()];
         int flags = getFlags(persistenceFlag);
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -474,10 +474,17 @@ public class Distribute extends AbstractAppCenterService {
         /* Clear workflow finished state if launch recreated, to achieve check on "startup". */
         if (activity.getClass().getName().equals(mLauncherActivityClassName)) {
             AppCenterLog.info(LOG_TAG, "Launcher activity restarted.");
-            if (mChannel != null && getStoredDownloadState() == DOWNLOAD_STATE_COMPLETED) {
-                mWorkflowCompleted = false;
-                mBrowserOpenedOrAborted = false;
-            }
+            resetWorkflow();
+        }
+    }
+
+    /**
+     * Reset current state to allow a new update check.
+     */
+    private void resetWorkflow() {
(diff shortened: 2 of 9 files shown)


sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/UpdateTrack.java
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+package com.microsoft.appcenter.distribute;
+
+/**
+ * Track to use for in-app updates.
+ */
+public enum UpdateTrack {
```
MatkovIvan: Let's use just int with annotations (like in other places) verification to easier wrap it into another languages

guperrot: True, though annotations were not supposed to be public as we depend on a package for that. We can use annotations only for internal code, not for the public signature.

guperrot: We discussed some spec changes anyway and will include the change to int to the PR.

guperrot: Actually we return that value with a getter so it's not a good thing for the customer. It's ok if it makes bindings slightly more difficult, it's more important that the API is easy to use in Java.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -452,10 +487,17 @@ public synchronized void onActivityCreated(Activity activity, Bundle savedInstan
...
-            }
+            resetWorkflow();
+        }
+    }
+
+    /**
+     * Reset current state to allow a new update check.
+     */
+    private void resetWorkflow() {
+        if (mChannel != null && getStoredDownloadState() == DOWNLOAD_STATE_COMPLETED) {
```
guperrot: This kind of check is what prevents the UI from switching to the new value @clpolet. Resume is checking for that too. I tested in app with a timer and got:

> 2020-01-21 16:40:35.053 9757-9757/com.microsoft.appcenter.sasquatch.project D/AppCenterDistribute: Previous dialog is still being shown in the same activity.
> 2020-01-21 16:40:35.053 9757-9757/com.microsoft.appcenter.sasquatch.project V/AppCenterDistribute: Already checking or checked latest release.

Looks like we are good.

guperrot:
Ran into problems with the old agent pool that was used for CI (also affecting develop). Migrated to the new azure pipelines setting to fix.

guperrot:
Verified lint locally with the modification and it passes, strange that it appears only now.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventActivity.java
```diff
@@ -188,7 +188,7 @@ private boolean onlyStringProperties() {
     }
 
     @Override
-    protected void send(View view) {
+    public void send(View view) {
```
guperrot: old VM could not run tests anymore for some reason (even on develop: https://msmobilecenter.visualstudio.com/SDK/_build/results?buildId=833380).
Switch to another pool and got the new lint error, which means lint behavior depends on what is installed on the machine for some reason.

PR #1360, opened 2020-01-28
Use new endpoints for private setup, public in-app updates

owlstronaut:

* [ ] Did you add unit tests?

## Description

Android uses the new endpoints for public/private

## Related PRs or issues

[AB#75579](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/75579)

### DIFF
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -1068,9 +1068,9 @@ public class Distribute extends AbstractAppCenterService {
         String releaseHash = computeReleaseHash(mPackageInfo);
         String url = mApiUrl;
         if (updateToken == null) {
-            url += String.format(GET_LATEST_PUBLIC_RELEASE_PATH_FORMAT, mAppSecret, releaseHash, getReportingParametersForUpdatedRelease(true));
+            url += String.format(GET_LATEST_PUBLIC_RELEASE_PATH_FORMAT, mAppSecret, releaseHash, getReportingParametersForUpdatedRelease(true, distributionGroupId));
         } else {
-            url += String.format(GET_LATEST_PRIVATE_RELEASE_PATH_FORMAT, mAppSecret, releaseHash, getReportingParametersForUpdatedRelease(false));
+            url += String.format(GET_LATEST_PRIVATE_RELEASE_PATH_FORMAT, mAppSecret, releaseHash, getReportingParametersForUpdatedRelease(false, distributionGroupId));
         }
         Map<String, String> headers = new HashMap<>();
         if (updateToken != null) {
@@ -1287,11 +1287,10 @@ public class Distribute extends AbstractAppCenterService {
      *                 For private group we report distribution_group_id and release_id.
      */
     @NonNull
-    private String getReportingParametersForUpdatedRelease(boolean isPublic) {
+    private String getReportingParametersForUpdatedRelease(boolean isPublic, String distributionGroupId) {
         String reportingParameters = "";
         AppCenterLog.debug(LOG_TAG, "Check if we need to report release installation..");
         String lastDownloadedReleaseHash = SharedPreferencesManager.getString(PREFERENCE_KEY_DOWNLOADED_RELEASE_HASH);
-        String distributionGroupId = SharedPreferencesManager.getString(PREFERENCE_KEY_DISTRIBUTION_GROUP_ID);
         if (!TextUtils.isEmpty(lastDownloadedReleaseHash)) {
             if (isCurrentReleaseWasUpdated(lastDownloadedReleaseHash)) {
                 AppCenterLog.debug(LOG_TAG, "Current release was updated but not reported yet, reporting..");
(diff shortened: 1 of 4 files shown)


sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1067,11 +1067,8 @@ synchronized void getLatestReleaseDetails(final String distributionGroupId, Stri
         }
         String releaseHash = computeReleaseHash(mPackageInfo);
         String url = mApiUrl;
-        if (updateToken == null) {
-
-            /* TODO Remove this group and use new endpoint. The hardcoded is temporary to keep testing on sasquatch int. */
-            String publicDistributionGroupId = distributionGroupId == null ? "3d054d79-8b26-426c-9a49-fed752c777d2" : distributionGroupId;
-            url += String.format(GET_LATEST_PUBLIC_RELEASE_PATH_FORMAT, mAppSecret, publicDistributionGroupId, releaseHash, getReportingParametersForUpdatedRelease(true, ""));
+        if (getUpdateTrack() == UpdateTrack.PUBLIC) {
```
guperrot: Please keep the in memory check with token (even if we have a token we don't pass it to this function if public mode, this is done in another function), otherwise this will be subject to value changes during current workflow and various other race conditions.

guperrot:
```suggestion
        if (updateToken == null) {
```

owlstronaut:
Waiting on 2 things
- [ ] sending the distributionGroupId as a query param in the public api so new installs get counted
- [ ] back-end for private to be fixed so we can test
- [ ] test public counting/updating works

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1202,10 +1200,7 @@ private synchronized void handleApiCallSuccess(Object releaseCallId, String rawR
             /* Reset state. */
             mCheckReleaseApiCall = null;
 
-            /* If we didn't know what distribution group we were originally tied to (public track). */
```
guperrot: Rollback this change, this is handled in another place for updates.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1284,25 +1281,23 @@ private synchronized void updateReleaseDetails(ReleaseDetails releaseDetails) {
...
      *                            For private group we report distribution_group_id and release_id.
-     * @param distributionGroupId distribution group id.
      */
     @NonNull
-    private String getReportingParametersForUpdatedRelease(boolean isPublic, String distributionGroupId) {
+    private String getReportingParametersForUpdatedRelease(boolean isPublic) {
         String reportingParameters = "";
         AppCenterLog.debug(LOG_TAG, "Check if we need to report release installation..");
         String lastDownloadedReleaseHash = SharedPreferencesManager.getString(PREFERENCE_KEY_DOWNLOADED_RELEASE_HASH);
+        String lastDistributionGroupId = SharedPreferencesManager.getString(PREFERENCE_KEY_DOWNLOADED_DISTRIBUTION_GROUP_ID);
```
guperrot: This is actually being cleared and the main key is updated at start time. So we need to use the main distribution groupId key like before this PR.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1284,25 +1281,23 @@ private synchronized void updateReleaseDetails(ReleaseDetails releaseDetails) {
 
     /**
      * Get reporting parameters for updated release.
-     *
-     * @param isPublic            are the parameters for public group or not.
-     *                            For public group we report install_id and release_id.
+     *  @param isPublic            are the parameters for public group or not.
```
dhei: Run formatter?
```suggestion
     *  @param isPublic           are the parameters for public group or not.
```

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1285,24 +1282,23 @@ private synchronized void updateReleaseDetails(ReleaseDetails releaseDetails) {
...
+     *                 For public group we report install_id, distribution_group_id and release_id.
+     *                 For private group we report distribution_group_id and release_id.
      */
     @NonNull
-    private String getReportingParametersForUpdatedRelease(boolean isPublic, String distributionGroupId) {
+    private String getReportingParametersForUpdatedRelease(boolean isPublic) {
         String reportingParameters = "";
         AppCenterLog.debug(LOG_TAG, "Check if we need to report release installation..");
         String lastDownloadedReleaseHash = SharedPreferencesManager.getString(PREFERENCE_KEY_DOWNLOADED_RELEASE_HASH);
+        String distributionGroupId = SharedPreferencesManager.getString(PREFERENCE_KEY_DISTRIBUTION_GROUP_ID);
```
guperrot: read it from the parameter like before, that will avoid an additional read from storage

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1285,10 +1282,9 @@ private synchronized void updateReleaseDetails(ReleaseDetails releaseDetails) {
     /**
      * Get reporting parameters for updated release.
      *
-     * @param isPublic            are the parameters for public group or not.
-     *                            For public group we report install_id and release_id.
-     *                            For private group we report distribution_group_id and release_id.
-     * @param distributionGroupId distribution group id.
```
dhei: Add this param back to javadoc.

owlstronaut: ah right, thanks

PR #1361, opened 2020-02-04
Feature/in app update track before start only

guperrot:

* [x] Did you add unit tests?

## Description

Implement the spec change where setInAppUpdateTrack works only before start.

## Related PRs or issues

[AB#75928](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/75928)


sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -192,12 +191,7 @@
     /**
      * Update track as set and returned by the API. The change with that value might not have been processed yet.
```
```suggestion
     * Update track as set and returned by the API. Can only be set before start.
```

PR #1368, opened 2020-02-14
Use new SDK 3.0.0 in demo app。

zhangcc01:

## Description

Use new SDK 3.0.0 in demo app。


apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
```diff
@@ -113,13 +113,7 @@ static void startAppCenter(Application application, String startTypeString) {
         int savedTrack = sSharedPreferences.getInt(application.getString(R.string.appcenter_distribute_track_state_key), 0);
         if (savedTrack != 0) {
             try {
```
annakocheshkova: Please remove try {... } catch(..) around the call

zhangcc01: @annakocheshkova Yes, I have removed in new commit.

apps/sasquatch/src/main/AndroidManifest.xml
```diff
@@ -65,9 +65,6 @@
         <activity
             android:name=".activities.AuthenticationProviderActivity"
             android:label="@string/title_auth" />
-        <activity
```
guperrot: That's a merge of the 2 activity Java files between build variants but the file still exists. You cannot remove the one reference to the manifest.

zhangcc01: @guperrot Updated in new commit.

PR #1371, opened 2020-02-19
Implement Distribute.disableAutomaticCheckForUpdate()

pranotipr:

* [x] Did you add unit tests?

## Description

Implemented method to disable the automatic check for updates on the Distribute start.

## Misc

[AB#75290](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/75290)


sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
```diff
@@ -0,0 +1,84 @@
...
+        when(SharedPreferencesManager.getString(PREFERENCE_KEY_REQUEST_ID)).thenReturn("r");
+
+        /* Enable in-app updates. */
+        Distribute.disableAutomaticCheckForUpdate();
+        start();
+        Distribute.setEnabled(true).get();
+        Distribute.getInstance().storeRedirectionParameters("r", "g", null);
+
+        /* No HTTP call done. */
+        verify(mHttpClient, never()).callAsync(anyString(), anyString(), eq(Collections.<String, String>emptyMap()), any(HttpClient.CallTemplate.class), any(ServiceCallback.class));
```
thyeggman: I'm not so sure this is the right verification to be doing.  If line 63 is deleted, the behavior doesn't change (as expected), but it's needed to cover the correct part of the code.  So the "functional" line doesn't make a difference.  If anyone has any suggestions to improve this test it would be welcome.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -683,6 +712,7 @@ public void run() {
 
     @WorkerThread
     private synchronized void handleCheckForUpdate() {
+        mManualCheckForUpdateRequested = true;
```
jaeklim: Is it okay to assign `true` even though `tryResetWorkflow()` returns `false`?

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
```diff
@@ -0,0 +1,115 @@
...
+import static com.microsoft.appcenter.distribute.DistributeConstants.PREFERENCE_KEY_REQUEST_ID;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+
```
jaeklim:
```suggestion
```

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
```diff
@@ -0,0 +1,115 @@
...
+public class DistributeDisableAutomaticCheckForUpdateTest extends AbstractDistributeTest {
+
+    @Test
+    public void disableAutomaticCheckForUpdateBeforeDistributeStartDoesNotCheckForUpdate() {
+
+        /* Disable automatic check for update then start. */
+        Distribute.disableAutomaticCheckForUpdate();
+        start();
+        Distribute.setEnabled(true).get();
```
jaeklim: This is unnecessary unless you want to enable it when Distribute is disabled.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
```diff
@@ -0,0 +1,115 @@
...
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+
+public class DistributeDisableAutomaticCheckForUpdateTest extends AbstractDistributeTest {
+
+    @Test
+    public void disableAutomaticCheckForUpdateBeforeDistributeStartDoesNotCheckForUpdate() {
```
jaeklim: This is a test for Public track. We should also make sure it doesn't open a browser on private track.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
thyeggman: I guess I don't understand why it wouldn't be...

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
thyeggman: This test has nothing to do with if a browser open or not

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1039,6 +1075,12 @@ synchronized void storeRedirectionParameters(@NonNull String requestId, @NonNull
             processDistributionGroupId(distributionGroupId);
             AppCenterLog.debug(LOG_TAG, "Stored redirection parameters.");
             cancelPreviousTasks();
+
+            /* Do not proceed if automatic check for update is disabled and manual check for update has not been called. */
+            if (mAutomaticCheckForUpdateDisabled && !mManualCheckForUpdateRequested) {
```
thyeggman: Might not be needed because this would only matter if we wanted to allow the flag state to be recognized in the middle of the authenticate/http call flow

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -884,6 +914,12 @@ else if (mReleaseDownloader == null || !mReleaseDownloader.isDownloading()) {
                 return;
             }
 
+            /* Do not proceed if automatic check for update is disabled and manual check for update has not been called. */
+            if (mAutomaticCheckForUpdateDisabled && !mManualCheckForUpdateRequested) {
+                AppCenterLog.info(LOG_TAG, "Automatic check for update is disabled. Will not check for update.");
```
MatkovIvan: We probably can improve the wording here (most likely folks have more edits here 😅 )
```suggestion
                AppCenterLog.info(LOG_TAG, "Automatic check for update is disabled. The SDK will not check for update now.");
```

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Test
+    public void manualCheckWhileAutomaticCheckDisabledChecksForUpdate() {
+
+        /* Start then call disable automatic check for update after Distribute has started. */
+        Distribute.disableAutomaticCheckForUpdate();
+        start();
+        Distribute.checkForUpdate();
+        Distribute.getInstance().onActivityResumed(mActivity);
```
MatkovIvan: Put verify right after `checkForUpdate` or comment why `onActivityResumed` is needed here. Also, please verify that there was no HTTP calls right before `checkForUpdate` call

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
```diff
@@ -0,0 +1,110 @@
...
+        verify(mHttpClient).callAsync(anyString(), anyString(), eq(Collections.<String, String>emptyMap()), any(HttpClient.CallTemplate.class), any(ServiceCallback.class));
+    }
+
+    @Test
+    public void manualCheckWhileAutomaticCheckDisabledChecksForUpdate() {
+
+        /* Start then call disable automatic check for update after Distribute has started. */
+        Distribute.disableAutomaticCheckForUpdate();
+        start();
+        verify(mHttpClient, times(0)).callAsync(anyString(), anyString(), eq(Collections.<String, String>emptyMap()), any(HttpClient.CallTemplate.class), any(ServiceCallback.class));
```
MatkovIvan: `times(0)` -> `never()`

thyeggman:
```suggestion
        verify(mHttpClient, never()).callAsync(anyString(), anyString(), eq(Collections.<String, String>emptyMap()), any(HttpClient.CallTemplate.class), any(ServiceCallback.class));
```

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.Test;
+
+import java.util.Collections;
+
+import static com.microsoft.appcenter.distribute.DistributeConstants.PREFERENCE_KEY_REQUEST_ID;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
```
MatkovIvan: Looks unused now, please do pre commit checks locally

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -285,6 +285,16 @@
      */
     private boolean mEnabledForDebuggableBuild;
 
+    /**
+     * Flags to check if automatic check for update is disabled.
```
```suggestion
     * Flag to check if automatic check for update is disabled.
```

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -406,6 +416,24 @@ public static void checkForUpdate() {
         getInstance().instanceCheckForUpdate();
     }
 
+    /**
+     * Disable automatic check for update before the service starts.
+     */
+    public static void disableAutomaticCheckForUpdate() {
```
guperrot: move it next to other static methods

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -542,6 +570,7 @@ protected synchronized void applyEnabledState(boolean enabled) {
             mTesterAppOpenedOrAborted = false;
             mBrowserOpenedOrAborted = false;
             mWorkflowCompleted = false;
+            mManualCheckForUpdateRequested = false;
```
guperrot: this should probably go into `cancelPreviousTasks`

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
guperrot: You need to reset it to false when the workflowCompleted is updated to true and a need unit test around that: restarting launcher activity after having done a successful manual check for update must do another check until called again explicitly.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Test
+    public void manualCheckWhileAutomaticCheckDisabledChecksForUpdate() {
+
+        /* Start then call disable automatic check for update after Distribute has started. */
+        Distribute.disableAutomaticCheckForUpdate();
+        start();
+        verify(mHttpClient, never()).callAsync(anyString(), anyString(), eq(Collections.<String, String>emptyMap()), any(HttpClient.CallTemplate.class), any(ServiceCallback.class));
+        Distribute.checkForUpdate();
```
guperrot: Would be nice to have a variant of this test that reverse order of foreground / checkForUpdate calls (it should also work).

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
thyeggman: It's already next to the other static methods, though?

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/EventActivity.java
```diff
@@ -67,6 +68,7 @@
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        Distribute.checkForUpdate();
```
guperrot: Revert changes to this file.

apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/MainActivity.java
```diff
@@ -109,6 +109,8 @@ static String getLogUrl(Context context, String startType) {
 
     static void startAppCenter(Application application, String startTypeString) {
 
+        Distribute.disableAutomaticCheckForUpdate();
```
guperrot: Revert changes to this file.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
```diff
@@ -61,4 +62,65 @@ public void manualCheckWhileAutomaticCheckDisabledChecksForUpdate() {
...
     }
+
+    @Test
+    public void manualCheckAfterActivityResumedWhileAutomaticCheckDisabledChecksForUpdate() {
+
+        /* Start then call disable automatic check for update after Distribute has started. */
+        Distribute.disableAutomaticCheckForUpdate();
+        start();
+        verify(mHttpClient, never()).callAsync(anyString(), anyString(), eq(Collections.<String, String>emptyMap()), any(HttpClient.CallTemplate.class), any(ServiceCallback.class));
+        Distribute.getInstance().onActivityResumed(mActivity);
```
guperrot: Swap lines 72 and 73, it would never start a HTTP call in background even when automatic updates are enabled.

versions.gradle
```diff
@@ -6,7 +6,7 @@
 // Version constants
 
 ext {
-    versionCode = 53
+    versionCode = 523
```
guperrot: Revert changes to this file.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
```diff
@@ -61,4 +62,65 @@ public void manualCheckWhileAutomaticCheckDisabledChecksForUpdate() {
...
+    }
+
+    @Test
+    public void checkForUpdateAfterManualCheckCompletesDoesNotCheckForUpdatesAgain() {
+
+        /* Start in public mode with automatic check for updates disabled. */
+        Distribute.disableAutomaticCheckForUpdate();
+        start();
+        Distribute.getInstance().onActivityResumed(mActivity);
+        Distribute.checkForUpdate();
```
guperrot: We also need a variant of this test where we swap lines 86 and 87 and we verify the first call is also made (no need to repeat the second part of the test).

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
```diff
@@ -61,4 +62,65 @@ public void manualCheckWhileAutomaticCheckDisabledChecksForUpdate() {
...
+
+        /* Check http call done. */
+        ArgumentCaptor<ServiceCallback> httpCallback = ArgumentCaptor.forClass(ServiceCallback.class);
+        verify(mHttpClient).callAsync(anyString(), anyString(), eq(Collections.<String, String>emptyMap()), any(HttpClient.CallTemplate.class), httpCallback.capture());
+
+        /* Complete call with no new release (this will return the default mock mReleaseDetails with version 0). */
+        httpCallback.getValue().onCallSucceeded(mock(HttpResponse.class));
+
+        /* Resume the activity again and manually check for updates. */
+        Distribute.getInstance().onActivityResumed(mActivity);
```
guperrot: The activity is already in foreground in this test (line 109), this line is unnecessary.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
```diff
@@ -61,4 +62,65 @@ public void manualCheckWhileAutomaticCheckDisabledChecksForUpdate() {
...
+        Distribute.checkForUpdate();
+
+        /* Check http call done. */
+        ArgumentCaptor<ServiceCallback> httpCallback = ArgumentCaptor.forClass(ServiceCallback.class);
+        verify(mHttpClient).callAsync(anyString(), anyString(), eq(Collections.<String, String>emptyMap()), any(HttpClient.CallTemplate.class), httpCallback.capture());
+
+        /* Complete call with no new release (this will return the default mock mReleaseDetails with version 0). */
+        httpCallback.getValue().onCallSucceeded(mock(HttpResponse.class));
+
+        /* Restart and check for updates again. */
```
guperrot:
```suggestion
        /* Restart. */
```

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
pranotipr: Added another test to check that.

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
pranotipr: Added the code to reset.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
pranotipr: Good catch, it was a mistake.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
thyeggman: Isn't that just the same as the test on line 53?  Is there a difference between that test and the one you're describing?

guperrot: oh I didn't realize that.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeDisableAutomaticCheckForUpdateTest.java
```diff
@@ -0,0 +1,125 @@
...
+        verify(mHttpClient).callAsync(anyString(), anyString(), eq(Collections.<String, String>emptyMap()), any(HttpClient.CallTemplate.class), httpCallback.capture());
+
+        /* Complete call with no new release (this will return the default mock mReleaseDetails with version 0). */
+        httpCallback.getValue().onCallSucceeded(mock(HttpResponse.class));
+
+        /* Restart. */
+        restartResumeLauncher(mActivity);
+
+        /* Do not call again. */
+        verify(mHttpClient).callAsync(anyString(), anyString(), eq(Collections.<String, String>emptyMap()), any(HttpClient.CallTemplate.class), notNull(ServiceCallback.class));
```
guperrot: Can we continue the test and call checkForUpdate manually one last time (and check second call) after the restart / check not called.

PR #1385, opened 2020-03-26
Add ApplicationLifecycleListener

russelarms:

* [ ] ~~Has `CHANGELOG.md` been updated?~~
* [x] Did you add unit tests?

## Description

Implement Application Lifecycle listener inside Core.

## Related PRs or issues

[AB#78295](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/78295)


sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
```diff
@@ -674,6 +680,8 @@ public void post(@NonNull Runnable runnable, Runnable disabledRunnable) {
                 handlerAppCenterOperation(runnable, disabledRunnable);
             }
         };
+        mApplicationLifecycleListener = new ApplicationLifecycleListener(mHandler);
+        mApplication.registerActivityLifecycleCallbacks(mApplicationLifecycleListener);
```
guperrot: shouldn't that register be at same place as the existing register?

guperrot: I see there is one already at the right place, thus do we really need this one?

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/ApplicationLifecycleListener.java
```diff
@@ -0,0 +1,148 @@
...
+    private Runnable mDelayedPauseRunnable = new Runnable() {
+        @Override
+        public void run() {
+            dispatchPauseIfNeeded();
+            dispatchStopIfNeeded();
+        }
+    };
+
+    public ApplicationLifecycleListener(Handler handler) {
+        this.mHandler = handler;
```
guperrot:
```suggestion
        mHandler = handler;
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/ApplicationLifecycleListener.java
```diff
@@ -0,0 +1,148 @@
...
+     */
+    private Handler mHandler;
+
+    /**
+     * A collection of ApplicationLifecycleCallbacks.
+     */
+    private Collection<ApplicationLifecycleCallbacks> mLifecycleCallbacks = new ArrayList<>();
+
+    private Runnable mDelayedPauseRunnable = new Runnable() {
+        @Override
```
guperrot:
```suggestion

        @Override
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/ApplicationLifecycleListener.java
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * Background thread handler.
+     */
+    private Handler mHandler;
+
+    /**
+     * A collection of ApplicationLifecycleCallbacks.
+     */
+    private Collection<ApplicationLifecycleCallbacks> mLifecycleCallbacks = new ArrayList<>();
```
guperrot: Use a CopyOnWriteArraySet for listeners. It's registered in background but the event is foreached in UI thread. Also we always use the Set interface for listeners as it's unique registration (avoid duplicates).

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/ApplicationLifecycleListener.java
```diff
@@ -0,0 +1,148 @@
...
+    @Override
+    public void onActivitySaveInstanceState(@NonNull Activity activity, @NonNull Bundle outState) {
+    }
+
+    @Override
+    public void onActivityDestroyed(@NonNull Activity activity) {
+    }
+
+    public interface ApplicationLifecycleCallbacks {
+        void onApplicationStarted();
```
guperrot: How about we rename to something closer to iOS using the foreground / background terminology.

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
russelarms: We put `mApplication.registerActivityLifecycleCallbacks(mApplicationLifecycleListener);`
 in the `startService` method at first. But it caused events duplication:
`startService` is called multiple times and we registered to activity lifecycle callbacks multiple times. 

Of course, it could be fixed with a flag, but this approach seems simpler.

sdk/appcenter/src/test/java/com/microsoft/appcenter/utils/ApplicationLifecycleListenerTest.java
```diff
@@ -0,0 +1,225 @@
...
+        /* Call onActivityDestroyed. */
+        mApplicationLifecycleListener.onActivityDestroyed(mActivityMock);
+
+        /* Call onActivityStarted. */
+        mApplicationLifecycleListener.onActivityStarted(mActivityMock);
+        verify(mockAppCenterService1, times(2)).onApplicationEnterForeground();
+        verify(mockAppCenterService2, times(2)).onApplicationEnterForeground();
+    }
+
+    private class MockCallbacks extends AbstractAppCenterService implements ApplicationLifecycleListener.ApplicationLifecycleCallbacks {
```
MatkovIvan:
```suggestion
    private class MockService extends AbstractAppCenterService {
```

I guess we already have some mock service for other tests.. Can we reuse it?

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/ApplicationLifecycleListener.java
```diff
@@ -0,0 +1,157 @@
...
+
+    @Override
+    public void onActivitySaveInstanceState(@NonNull Activity activity, @NonNull Bundle outState) {
+    }
+
+    @Override
+    public void onActivityDestroyed(@NonNull Activity activity) {
+    }
+
+    public interface ApplicationLifecycleCallbacks {
```
MatkovIvan: Comments

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Version 3.1.1 (Under development)
 
+### App Center Distribute
+
+* **[Feature]** Updates are checked automatically when an application enters a foreground state.
```
MatkovIvan: This PR doesn't do it

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/ApplicationLifecycleListener.java
```diff
@@ -0,0 +1,164 @@
...
+            dispatchPauseIfNeeded();
+            dispatchStopIfNeeded();
+        }
+    };
+
+    public ApplicationLifecycleListener(Handler handler) {
+        mHandler = handler;
+    }
+
+    public void registerApplicationLifecycleCallbacks(ApplicationLifecycleCallbacks lifecycleCallback) {
```
MatkovIvan: Add javadoc to this method

sdk/appcenter/src/test/java/com/microsoft/appcenter/utils/ApplicationLifecycleListenerTest.java
```diff
@@ -0,0 +1,279 @@
...
+
+        /* Call onActivityStopped. */
+        mApplicationLifecycleListener.onActivityStopped(mActivityMock);
+
+        /* Verify the callback was not called if we transition to another activity inside the app. */
+        verify(mockAppCenterService1, times(1)).onApplicationEnterForeground();
+        verify(mockAppCenterService2, times(1)).onApplicationEnterForeground();
+    }
+
+    private class MockService extends AbstractAppCenterService {
```
MatkovIvan: Can we reuse `DummyService` from other tests?

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
guperrot: Why do we also need it at https://github.com/microsoft/appcenter-sdk-android/pull/1385/files#diff-918508e799dc541fd29d897160e9b0a5R935 then?

sdk/appcenter/src/test/java/com/microsoft/appcenter/utils/ApplicationLifecycleListenerTest.java
```diff
@@ -0,0 +1,263 @@
...
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+@RunWith(PowerMockRunner.class)
+public class ApplicationLifecycleListenerTest {
+
+    @Mock
+    Handler mHandlerMock;
```
guperrot: If it can be private, it should.

sdk/appcenter/src/test/java/com/microsoft/appcenter/utils/ApplicationLifecycleListenerTest.java
```diff
@@ -0,0 +1,263 @@
...
+
+    @Mock
+    Activity mActivityMock;
```
guperrot: If it can be private, it should.

sdk/appcenter/src/test/java/com/microsoft/appcenter/utils/ApplicationLifecycleListenerTest.java
```diff
@@ -0,0 +1,263 @@
...
+            public Void answer(InvocationOnMock invocation) {
+                ((Runnable) invocation.getArguments()[0]).run();
+                return null;
+            }
+        }).when(mHandlerMock).postDelayed(any(Runnable.class), anyLong());
+        doAnswer(new Answer<Void>() {
+
+            @Override
+            public Void answer(InvocationOnMock invocation) {
+                ((Runnable) invocation.getArguments()[0]).run();
```
guperrot: Why executing the runnable since here the goal is to mock canceling it?

sdk/appcenter/src/test/java/com/microsoft/appcenter/utils/ApplicationLifecycleListenerTest.java
```diff
@@ -0,0 +1,263 @@
...
+    public void onActivityDoubleResumedTest() {
+
+        /* Prepare data. */
+        doAnswer(new Answer<Void>() {
+
+            @Override
+            public Void answer(final InvocationOnMock invocation) {
+
+                /* Post with a small delay to ensure that application resumes first. */
+                new Timer().schedule(new TimerTask() {
```
guperrot: Can't we use semaphores or alike to make the test reliable? We have done that in other places.
This could lead to flakiness to use timers.

sdk/appcenter/src/test/java/com/microsoft/appcenter/utils/ApplicationLifecycleListenerTest.java
guperrot: Looks like we don't need to mock anything for removeCallbacks, we just verify invocations.

sdk/appcenter/src/test/java/com/microsoft/appcenter/utils/ApplicationLifecycleListenerTest.java
```diff
@@ -0,0 +1,263 @@
...
+
+                    @Override
+                    public void run() {
+                        ((Runnable) invocation.getArguments()[0]).run();
+                    }
```
guperrot: Same comment here.

sdk/appcenter/src/test/java/com/microsoft/appcenter/utils/ApplicationLifecycleListenerTest.java
```diff
@@ -0,0 +1,263 @@
...
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+@RunWith(PowerMockRunner.class)
+public class ApplicationLifecycleListenerTest {
```
guperrot: Which one is the test where we transition activities faster than 700ms?

sdk/appcenter/src/main/java/com/microsoft/appcenter/AppCenter.java
MatkovIvan: @guperrot that line subscribes services to application (not activity) lifecycle events.

sdk/appcenter/src/test/java/com/microsoft/appcenter/utils/ApplicationLifecycleListenerTest.java
russelarms: **activityTransitionTest** is responsible for that scenario. We'll add more comments.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/ApplicationLifecycleListener.java
```diff
@@ -0,0 +1,181 @@
+/*
+ * Copyright (C) 2017 The Android Open Source Project
+ * Portions Copyright (c) Microsoft Corporation
```
guperrot: Discussed outside github.

sdk/appcenter/src/test/java/com/microsoft/appcenter/utils/ApplicationLifecycleListenerTest.java
```diff
@@ -0,0 +1,238 @@
...
+
+        /* Call onActivityResumed. */
+        mApplicationLifecycleListener.onActivityResumed(mActivityMock);
+        verify(mHandlerMock, never()).removeCallbacks(any(Runnable.class));
+
+        /* Call onActivityResumed again. */
+        mApplicationLifecycleListener.onActivityResumed(mActivityMock);
+        verify(mHandlerMock, never()).removeCallbacks(any(Runnable.class));
+
+        /* Pause.*/
```
guperrot:
```suggestion
        /* Pause. */
```

PR #1392, opened 2020-03-31
Automatically check for update when application switches from background to foreground

MatkovIvan:
* [x] Did you add unit tests?

## Description

This feature will improve in-app update experience on Android. Currently one needs to back from main activity (depending on the app design, this might mean killing the app) to get a new update prompt. It will enable testers to get the prompt either when the app is running or when the user moves the app to background and then brings it back to foreground.

## Related PRs or issues

[AB#73043](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/73043)
https://github.com/microsoft/appcenter/issues/1650


ThirdPartyNotices.txt
```diff
@@ -5,6 +5,7 @@ Visual Studio App Center SDK for Android incorporates third party material from
 
 1.	Gradle (http://gradle.org)
 2.	Android Robot image (https://developer.android.com/distribute/tools/promote/brand.html)
+3.  Android Lifecycle Process (https://mvnrepository.com/artifact/androidx.lifecycle/lifecycle-process)
```
thyeggman:
```suggestion
3.    Android Lifecycle Process (https://mvnrepository.com/artifact/androidx.lifecycle/lifecycle-process)
```

MatkovIvan: Good catch, but let's keep tabs instead of spaces here
```suggestion
3.	Android Lifecycle Process (https://mvnrepository.com/artifact/androidx.lifecycle/lifecycle-process)
```

PR #1398, opened 2020-04-06
Fix filter of minidump files for NDK crash dumps

vvechkanov:
* [x] Did you add unit tests?

## Description

Change filter of minidump files fro NDK crash dumps, in the way that only "*.dmp" files can pass. Also clear directory with minidumps files after processing those files. 

Modify time while testing crash to be in seconds instead of milliseconds, because setLastModified do not work correctly with milliseconds, and also when the equal check fails between two dates differs only in milliseconds, it is hard to find a reason, because Date not show milliseconds in string

## Related PRs or issues

[AB#78588](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/78588)

## Misc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### App Center Crashes
 
 * **[Fix]** Remove the multiple attachments warning as that is now supported by the portal.
+* **[Fix]** Fix remove minidump files when the sending crash report was discarded.
 
 ### App Center Distribute
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### App Center Crashes
 
 * **[Fix]** Remove the multiple attachments warning as that is now supported by the portal.
+* **[Fix]** Change minidump filter to use file extension instead of name.
 * **[Fix]** Fix remove minidump files when the sending crash report was discarded.
 
 ### App Center Distribute
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 * **[Fix]** Remove the multiple attachments warning as that is now supported by the portal.
 * **[Fix]** Change minidump filter to use file extension instead of name.
-* **[Fix]** Fix remove minidump files when the sending crash report was discarded.
+* **[Fix]** Fix removing minidump files when the sending crash report was discarded.
 
 ### App Center Distribute
 
--- a/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
+++ b/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
@@ -982,7 +982,7 @@ public class Crashes extends AbstractAppCenterService {
                         iterator.remove();
                         removeAllStoredErrorLogFiles(id);
                     }
-                    ErrorLogHelper.cleanPendingMinidumpDirectory();
+                    ErrorLogHelper.cleanDirectory(ErrorLogHelper.getPendingMinidumpDirectory());
                 }
 
                 /* We send the crash. */
--- a/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
+++ b/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
@@ -700,7 +700,7 @@ public class Crashes extends AbstractAppCenterService {
 
                 @Override
                 public boolean accept(File dir, String filename) {
-                    return !filename.equals(DEVICE_INFO_FILE);
+                    return filename.endsWith(".dmp");
                 }
             });
             if (minidumpSubfolderFiles == null || minidumpSubfolderFiles.length == 0) {
@@ -982,6 +982,7 @@ public class Crashes extends AbstractAppCenterService {
                         iterator.remove();
                         removeAllStoredErrorLogFiles(id);
                     }
+                    ErrorLogHelper.cleanPendingMinidumpDirectory();
                 }
 
                 /* We send the crash. */
--- a/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
+++ b/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
@@ -982,7 +982,7 @@ public class Crashes extends AbstractAppCenterService {
                         iterator.remove();
                         removeAllStoredErrorLogFiles(id);
                     }
-                    ErrorLogHelper.cleanDirectory(ErrorLogHelper.getPendingMinidumpDirectory());
+                    ErrorLogHelper.cleanPendingMinidumps();
                 }
 
                 /* We send the crash. */
--- a/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
+++ b/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
@@ -66,7 +66,7 @@ import static android.content.ComponentCallbacks2.TRIM_MEMORY_RUNNING_LOW;
 import static android.content.ComponentCallbacks2.TRIM_MEMORY_RUNNING_MODERATE;
 import static android.util.Log.getStackTraceString;
 import static com.microsoft.appcenter.Constants.WRAPPER_SDK_NAME_NDK;
-import static com.microsoft.appcenter.crashes.utils.ErrorLogHelper.DEVICE_INFO_FILE;
+import static com.microsoft.appcenter.crashes.utils.ErrorLogHelper.MINIDUMP_FILE_EXTENSION;
 
 /**
  * Crashes service.
@@ -700,7 +700,7 @@ public class Crashes extends AbstractAppCenterService {
 
                 @Override
                 public boolean accept(File dir, String filename) {
-                    return filename.endsWith(".dmp");
+                    return filename.endsWith(MINIDUMP_FILE_EXTENSION);
                 }
             });
             if (minidumpSubfolderFiles == null || minidumpSubfolderFiles.length == 0) {
--- a/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelper.java
+++ b/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelper.java
@@ -574,8 +574,7 @@ public class ErrorLogHelper {
      * Clear directory.
      * @param dir a folder.
      */
-    @NonNull
-    public static void cleanDirectory(File dir) {
+    public static void cleanDirectory(@NonNull File dir) {
         if (dir.isDirectory()) {
             File[] files = dir.listFiles();
             if (files != null && files.length > 0) {
--- a/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelper.java
+++ b/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelper.java
@@ -574,8 +574,9 @@ public class ErrorLogHelper {
      * Clear directory.
      * @param dir a folder.
      */
+    @NonNull
     public static void cleanDirectory(File dir) {
-        if (dir != null && dir.isDirectory()) {
+        if (dir.isDirectory()) {
             File[] files = dir.listFiles();
             if (files != null && files.length > 0) {
                 for (File file : files) {
--- a/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelper.java
+++ b/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelper.java
@@ -571,14 +571,17 @@ public class ErrorLogHelper {
     }
 
     /**
-     * Clear pending minidump directory.
+     * Clear directory.
+     * @param dir a folder.
      */
-    public static void cleanPendingMinidumpDirectory() {
-        File[] files = getPendingMinidumpDirectory().listFiles();
-        if (files != null && files.length > 0) {
-            for (File file : files) {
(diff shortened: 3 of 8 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### App Center Crashes
 
 * **[Fix]** Remove the multiple attachments warning as that is now supported by the portal.
+* **[Fix]** Fix remove minidump files when the sending crash report was discarded.
```
MatkovIvan, line 8: I see that processing rules are changed, so the main fix should be about processing and sending the crashes but not files clean up, right?

russelarms: I think it can be written as two separate items. Something about fixing the minidump filter should also be added.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -700,7 +700,7 @@ private void processMinidumpFiles() {
 
                 @Override
                 public boolean accept(File dir, String filename) {
-                    return !filename.equals(DEVICE_INFO_FILE);
+                    return filename.endsWith(".dmp");
```
russelarms, line 703: Let's move the extension to ErrorLogHelper constants with a comment like we did for json:
https://github.com/microsoft/appcenter-sdk-android/blob/develop/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelper.java#L61

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -982,6 +982,7 @@ public void run() {
                         iterator.remove();
                         removeAllStoredErrorLogFiles(id);
                     }
+                    ErrorLogHelper.cleanDirectory(ErrorLogHelper.getPendingMinidumpDirectory());
```
russelarms, line 985: I think it's better to use a method without a parameter here since it's not used anywhere else and more clear for a reader. Also, we have a method `FileManager.deleteDir` which can be confused with this one.
e.g. `ErrorLogHelper.cleanPendingMinidumps()`

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelper.java
```diff
@@ -570,6 +570,23 @@ private static StackFrame getModelStackFrame(StackTraceElement stackTraceElement
...
+     * @param dir a folder.
+     */
+    @NonNull
+    public static void cleanDirectory(File dir) {
+        if (dir.isDirectory()) {
+            File[] files = dir.listFiles();
+            if (files != null && files.length > 0) {
+                for (File file : files) {
+                    boolean isDelete = file.delete();
+                    AppCenterLog.debug(Crashes.LOG_TAG, String.format("Minidump file '%s' has delete status: '%s'", file.getPath(), isDelete));
```
russelarms, line 584: It's excessive logging IMO.

sdk/appcenter-crashes/src/test/java/com/microsoft/appcenter/crashes/CrashesTest.java
```diff
@@ -1120,19 +1125,23 @@ public ManagedErrorLog answer(InvocationOnMock invocation) {
...
+    @SuppressWarnings("ResultOfMethodCallIgnored")
     private ManagedErrorLog testNativeCrashLog(long appStartTime, long crashTime, boolean correlateSession, boolean hasDeviceInfo) throws Exception {
 
-        /* Setup mock for a crash in disk. */
-        File minidumpFile = mock(File.class);
-        when(minidumpFile.getName()).thenReturn("mockFile");
-        when(minidumpFile.isDirectory()).thenReturn(false);
-        when(minidumpFile.lastModified()).thenReturn(crashTime);
+        /* Create minidump sub-folder. */
+        File minidumpSubfolder = new File("./mockFolder");
```
russelarms, line 1132: Let's use TemporaryFolder rule for the creation of test files and folders in this PR. It has some benefits, like removing all its content after the test automatically.
See: https://github.com/microsoft/appcenter-sdk-android/blob/develop/sdk/appcenter-crashes/src/test/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelperTest.java#L76

sdk/appcenter-crashes/src/test/java/com/microsoft/appcenter/crashes/CrashesTest.java
```diff
@@ -1185,14 +1194,19 @@ public ManagedErrorLog answer(InvocationOnMock invocation) {
...
+        FileManager.deleteDir(minidumpFile);
         return (ManagedErrorLog) log.getValue();
     }
 
     @Test
     @PrepareForTest({SessionContext.class, DeviceInfoHelper.class})
     public void minidumpDeviceInfoNull() throws Exception {
-        long appStartTime = 95L;
-        long crashTime = 126L;
+        long appStartTime = 95000L;
```
russelarms, line 1208: Why change?

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
vvechkanov: My points to use the directory as a parameter:
- Easier to test
- Reusable (we could (and maybe even should) move this function into FileManager class).

Any other opinions?

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
vvechkanov: Done

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
vvechkanov: Create a function `ErrorLogHelper.cleanPendingMinidumps() `and use newly created `FileManager.cleanDir(File dir)` inside. In this case, we have pros from both ideas:
- Easy to test (we test mostly` FileManager.cleanDir`);
- Easy to read (no parameters in `ErrorLogHelper.cleanPendingMinidumps)`;
- Possible reuse of function in FileManager.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelper.java
vvechkanov: Removed

sdk/appcenter-crashes/src/test/java/com/microsoft/appcenter/crashes/CrashesTest.java
vvechkanov: Fixed

CHANGELOG.md
vvechkanov: I add one more line to address changes in this PR more correctly.

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/FileManager.java
```diff
@@ -242,6 +242,20 @@ public static boolean deleteDir(File file) {
...
 
+    /**
+     * Delete all files inside this directory, without deleting directory itself.
+     *
+     * @param file The directory to delete.
+     */
+    public static void cleanDir(@NonNull File file) {
+        File[] contents = file.listFiles();
+        if (contents != null) {
+            for (File f : contents) {
```
MatkovIvan, line 253: Avoid one letter names

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/FileManager.java
```diff
@@ -242,6 +242,20 @@ public static boolean deleteDir(File file) {
         return file.delete();
     }
 
+    /**
+     * Delete all files inside this directory, without deleting directory itself.
+     *
+     * @param file The directory to delete.
+     */
+    public static void cleanDir(@NonNull File file) {
```
MatkovIvan, line 250:
```suggestion
    public static void cleanDirectory(@NonNull File directory) {
```

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### App Center Crashes
 
 * **[Fix]** Remove the multiple attachments warning as that is now supported by the portal.
+* **[Fix]** Change minidump filter to use file extension instead of name.
+* **[Fix]** Fix remove minidump files when the sending crash report was discarded.
```
guperrot, line 9:
```suggestion
* **[Fix]** Fix removing minidump files when the sending crash report was discarded.
```

sdk/appcenter-crashes/src/test/java/com/microsoft/appcenter/crashes/CrashesTest.java
```diff
@@ -1120,19 +1129,23 @@ public ManagedErrorLog answer(InvocationOnMock invocation) {
         verify(mockChannel, never()).enqueue(any(ManagedErrorLog.class), eq(crashes.getGroupName()), anyInt());
     }
 
+    @SuppressWarnings("ResultOfMethodCallIgnored")
```
guperrot, line 1132: Can assert true in tests instead of suppressing.

sdk/appcenter-crashes/src/test/java/com/microsoft/appcenter/crashes/CrashesTest.java
```diff
@@ -1120,19 +1129,23 @@ public ManagedErrorLog answer(InvocationOnMock invocation) {
...
+        File minidumpFile = new File(minidumpSubfolder, "mockFile.dmp");
+        minidumpFile.createNewFile();
+        minidumpFile.setLastModified(crashTime);
 
-        /* Mock sub-folder. */
-        File minidumpSubfolder = mock(File.class);
-        when(minidumpSubfolder.getName()).thenReturn("mockFolder");
-        when(minidumpSubfolder.isDirectory()).thenReturn(true);
-        when(minidumpSubfolder.listFiles(any(FilenameFilter.class))).thenReturn(new File[]{minidumpFile});
+        /* Create an additional file in a folder to be filtered later */
```
guperrot, line 1144:
```suggestion
        /* Create an additional file in a folder to be filtered later. */
```

sdk/appcenter-crashes/src/test/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelperTest.java
```diff
@@ -365,6 +367,24 @@ public void validateProperties() {
         assertEquals(truncatedMapItem, actualProperties.get(truncatedMapItem));
     }
 
+    @Test
+    public void cleanDirectory() throws java.lang.Exception {
+        /* Prepare data. */
```
guperrot, line 372:
```suggestion

        /* Prepare data. */
```

sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/FileManager.java
```diff
@@ -232,16 +232,30 @@ public static boolean delete(@NonNull File file) {
...
         return file.delete();
     }
 
+    /**
+     * Delete all files inside this directory, without deleting directory itself.
+     *
+     * @param directory The directory to delete.
+     */
+    public static void cleanDirectory(@NonNull File directory) {
+        File[] contents = directory.listFiles();
```
guperrot, line 251: listFiles will just return null on a regular file?

vvechkanov: Yes, exactly. 
We also have a test, to check that is we put a regular file here, then no crashes will happen (`cleanDirectoryWhenNotDirectory`).

sdk/appcenter-crashes/src/test/java/com/microsoft/appcenter/crashes/CrashesTest.java
vvechkanov: Thank you, fixed.

sdk/appcenter-crashes/src/test/java/com/microsoft/appcenter/crashes/CrashesTest.java
```diff
@@ -627,6 +634,8 @@ public void handleUserConfirmationDoNotSend() throws JSONException {
 
         verify(mockListener, never()).getErrorAttachments(any(ErrorReport.class));
```
MatkovIvan, line 636: Add comment or remove empty line

sdk/appcenter-crashes/src/test/java/com/microsoft/appcenter/crashes/CrashesTest.java
```diff
@@ -603,30 +608,34 @@ public void getChannelListenerErrors() {
 
     @Test
     public void handleUserConfirmationDoNotSend() throws JSONException {
+
+        /* Prepare data. Mock classes */
```
guperrot, line 612:
```suggestion
        /* Prepare data. Mock classes. */
```

sdk/appcenter-crashes/src/test/java/com/microsoft/appcenter/crashes/CrashesTest.java
```diff
@@ -603,30 +608,34 @@ public void getChannelListenerErrors() {
...
         Crashes crashes = Crashes.getInstance();
         LogSerializer logSerializer = mock(LogSerializer.class);
         when(logSerializer.deserializeLog(anyString(), anyString())).thenReturn(mErrorLog);
-
         crashes.setLogSerializer(logSerializer);
         crashes.setInstanceListener(mockListener);
         crashes.onStarting(mAppCenterHandler);
         crashes.onStarted(mock(Context.class), mock(Channel.class), "", null, true);
 
+        /* Verify */
```
guperrot, line 634:
```suggestion
        /* Verify. */
```

PR #1417, opened 2020-04-29
Add CodeQL Analysis workflow

AnastasiaKubova:
mirror #1416


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,44 @@
+name: "Code Scanning - Action"
+
+on:
+  push:
+  schedule:
+    - cron: '0 0 * * 0'
```
Jamminroot, line 6: any specific reason to use cron scheduling?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,44 @@
...
+
+jobs:
+  CodeQL-Build:
+
+    strategy:
+      fail-fast: false
```
Jamminroot, line 12: I think this task should immediately stop and report fail if something goes wrong, so using `fail-fast: true` might be more appropriate.

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
```diff
@@ -319,7 +319,7 @@ protected static void trackPage(String name, Map<String, String> properties) {
      * Additional validation rules apply depending on the configured secret.
      * <p>
      * For App Center, the name cannot be longer than 256 and is truncated otherwise.
-     * For One Collector, the name needs to match the <tt>[a-zA-Z0-9]((\.(?!(\.|$)))|[_a-zA-Z0-9]){3,99}</tt> regular expression.
+     * For One Collector, the name needs to match the <code>[a-zA-Z0-9]((\.(?!(\.|$)))|[_a-zA-Z0-9]){3,99}</code> regular expression.
```
Jamminroot, line 322: Im not sure - what are those changes?

.github/workflows/codeql.yml
MatkovIvan: @Jamminroot we need to get full error list here

sdk/appcenter-analytics/src/main/java/com/microsoft/appcenter/analytics/Analytics.java
MatkovIvan: Javadoc errors
`error: tag not supported in the generated HTML version: tt`

.github/workflows/codeql.yml
Jamminroot: This one resolved.

.github/workflows/codeql.yml
AnastasiaKubova: We have the same implementation in another our PR
https://github.com/microsoft/appcenter-sdk-react-native/pull/826
I think we need to keep uniformity anywhere.

PR #1437, opened 2020-07-10
Modify the test app to test the new Distribute library

AnastasiaKubova:

* [ ] ~Has `CHANGELOG.md` been updated?~
* [ ] ~Are tests passing locally?~
* [ ] ~Did you add unit tests?~

## Description

Add a build variant on our test app that integrates this new package.

## Related PRs or issues

[AB#81913](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/81913)


apps/sasquatch/build.gradle
```diff
@@ -31,6 +31,14 @@ android {
             dimension "dependency"
             applicationIdSuffix ".jcenter"
         }
+        projectGooglePlay {
+            dimension "dependency"
```
MatkovIvan, line 35: It doesn't seem correct. Please look at `flavorDimensions` and add another one

apps/sasquatch/build.gradle
```diff
@@ -31,6 +31,14 @@ android {
...
+            applicationIdSuffix ".project"
+        }
+        jcenterGooglePlay {
+            dimension "dependency"
+            applicationIdSuffix ".jcenter"
+        }
         vanilla {
             dimension "pushLibrary"
```
MatkovIvan, line 43: BTW @guperrot do we still need non firebase variant?

guperrot: We don't really use it nowadays.

apps/sasquatch/build.gradle
```diff
@@ -68,15 +74,22 @@ dependencies {
...
     projectDependencyImplementation project(':sdk:appcenter-push')
 
     def appCenterSdkVersion = "3.2.2"
     jcenterDependencyImplementation "com.microsoft.appcenter:appcenter-analytics:${appCenterSdkVersion}"
     jcenterDependencyImplementation "com.microsoft.appcenter:appcenter-crashes:${appCenterSdkVersion}"
-    jcenterDependencyImplementation "com.microsoft.appcenter:appcenter-distribute:${appCenterSdkVersion}"
     jcenterDependencyImplementation "com.microsoft.appcenter:appcenter-push:${appCenterSdkVersion}"
 
+    def taskRequests = getGradle().getStartParameter().getTaskRequests().toString()
+    if (taskRequests.contains("GooglePlay")) {
```
MatkovIvan, line 85: You don't need conditions here, just use prefix like in other places

apps/sasquatch/build.gradle
MatkovIvan: ok, we've filed a separate task to clean this up

apps/sasquatch/build.gradle
AnastasiaKubova: We cannot do this, because we have in `productFlavors` there are other products that will have to be prescribed for each dependency.

guperrot: We should still be able to combine projectDependency and appCenter for example.
We should test via dimension specs it as we will probably not document this with a if.

AnastasiaKubova: I checked how it works with use `projectDependencyAppCenterImplementation` and `projectDependencyGooglePlayImplementation`, but in these cases it is failing with error "Could not find method ..". So using this condition is one way to resolve it.

PR #1452, opened 2020-08-06
Add set ndk script to post clone

annakocheshkova:
~* [ ] Has `CHANGELOG.md` been updated?~
~* [ ] Are tests passing locally?~
~* [ ] Did you add unit tests?~
~* [ ] Did you check UI tests on the sample app? They are not executed on CI.~

## Description

Fix App Center build for Android 11.

### DIFF
--- a/apps/appcenter-post-clone.sh
+++ b/apps/appcenter-post-clone.sh
@@ -1,3 +1,4 @@
 #!/usr/bin/env bash
 echo "Executing post clone script in `pwd`"
 echo "dimension.dependency=$DIMENSION_DEPENDENCY" >> $APPCENTER_SOURCE_DIRECTORY/local.properties
+../scripts/set-ndk-version.sh


apps/appcenter-post-clone.sh
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env bash
 echo "Executing post clone script in `pwd`"
 echo "dimension.dependency=$DIMENSION_DEPENDENCY" >> $APPCENTER_SOURCE_DIRECTORY/local.properties
+../scripts/set-ndk-version.sh
```
MatkovIvan, line 4: BTW why that script is in `scripts` folder but changes the app only?

annakocheshkova: I think it's more convenient to store them all in one folder

PR #1458, opened 2020-09-10
Adding Custom Error File managing system

Zornikachu:
## Description

Adding a Custom Error log file option to ErrorLogHelper class to differentiate Error logs folders between instances of appcenter


AnastasiaKubova (changes requested):
Hi and thanks for creating this PR. I checked it. 
Since we don’t support multiple instances running simultaneously, so this implementation will be work unexpected. In this case the `sCustomErrorDirectory` value will be rewritten each time when will be called method `setCustomErrorDirectory` and will be used the last set value.

DmitriyKirakosyan:
I'm closing this PR since there was no activity for a while.

PR #1480, opened 2020-12-02
Remove throwable files

AnastasiaKubova:

* [x] Did you add unit tests?

## Description

Repro steps:

- Build and run the Sasquatch app in debug mode from Android Studio.
- Set the storage size to 32768 (32kb).
- Turn off Wi-Fi and LTE, so there's no network on the device/emulator.
- Crash the app several times (7-10 times).
- Turn on Wi-Fi/LTE.
- Run the Sasquatch app in debug mode from Android Studio, open Device File Explorer in the IDE and see extra throwable files.

**Actual:**
There are several .throwable files that are not cleaned up even after restart.

**Expected:**
After network turned on .throwable files should be cleaned up.

## Related PRs or issues

[AB#83943](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/83943)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Version 4.0.1 (Under development)
 
+### App Center Crashes
+
+* **[Fix]** Fix removing throwable files after rewritten error logs due to small database size.
+
 ### App Center Distribute
 
 * **[Fix]** Fix browser opening when using a private distribution group on Android 11.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### App Center Crashes
 
-* **[Fix]** Fix removing throwable files after rewritten error logs due to small database size.
+* **[Fix]** Fix removing throwable files after rewriting error logs due to small database size.
 
 ### App Center Distribute
 
--- a/sdk/appcenter-crashes/src/androidTest/java/com/microsoft/appcenter/crashes/WrapperSdkExceptionManagerAndroidTest.java
+++ b/sdk/appcenter-crashes/src/androidTest/java/com/microsoft/appcenter/crashes/WrapperSdkExceptionManagerAndroidTest.java
@@ -109,7 +109,10 @@ public class WrapperSdkExceptionManagerAndroidTest {
             startFresh();
 
             /* Save crash. */
-            error.id = WrapperSdkExceptionManager.saveWrapperException(Thread.currentThread(), null, new Exception(), error.data);
+            Exception exception = new Exception();
+            exception.setType("type");
+            exception.setMessage("message");
+            error.id = WrapperSdkExceptionManager.saveWrapperException(Thread.currentThread(), null, exception, error.data);
             String loadedData = WrapperSdkExceptionManager.loadWrapperExceptionData(error.id);
 
             if (error.data == null) {
--- a/sdk/appcenter-crashes/src/androidTest/java/com/microsoft/appcenter/crashes/WrapperSdkExceptionManagerAndroidTest.java
+++ b/sdk/appcenter-crashes/src/androidTest/java/com/microsoft/appcenter/crashes/WrapperSdkExceptionManagerAndroidTest.java
@@ -109,10 +109,7 @@ public class WrapperSdkExceptionManagerAndroidTest {
             startFresh();
 
             /* Save crash. */
-            Exception exception = new Exception();
-            exception.setType("type");
-            exception.setMessage("message");
-            error.id = WrapperSdkExceptionManager.saveWrapperException(Thread.currentThread(), null, exception, error.data);
+            error.id = WrapperSdkExceptionManager.saveWrapperException(Thread.currentThread(), null, new Exception(), error.data);
             String loadedData = WrapperSdkExceptionManager.loadWrapperExceptionData(error.id);
 
             if (error.data == null) {
--- a/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
+++ b/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
@@ -108,6 +108,12 @@ public class Crashes extends AbstractAppCenterService {
     static final String ERROR_GROUP = "groupErrors";
 
     /**
+     * Minidump file.
+     */
+    @VisibleForTesting
+    static final String MINIDUMP_FILE = "minidump";
+
+    /**
      * Name of the service.
      */
     private static final String SERVICE_NAME = "Crashes";
@@ -474,8 +480,11 @@ public class Crashes extends AbstractAppCenterService {
         if (isInstanceEnabled()) {
             processPendingErrors();
 
-            /* Remove throwable files. */
-            ErrorLogHelper.removeLostThrowableFiles();
+            if (mErrorReportCache.isEmpty()) {
+
+                /* Remove lost throwable files. */
+                ErrorLogHelper.removeLostThrowableFiles();
+            }
         }
     }
 
@@ -915,6 +924,9 @@ public class Crashes extends AbstractAppCenterService {
 
     @VisibleForTesting
     String buildStackTrace(Exception exception) {
+        if (exception.getType() == null || exception.getMessage() == null) {
+            return null;
+        }
         String stacktrace = String.format("%s: %s", exception.getType(), exception.getMessage());
         if (exception.getFrames() == null) {
             return stacktrace;
@@ -933,12 +945,7 @@ public class Crashes extends AbstractAppCenterService {
             report.setDevice(log.getDevice());
             return report;
         } else {
-            String stackTrace;
-            if (log.getException().getType().equals("minidump")) {
-                stackTrace = getStackTraceString(new NativeException());
-            } else {
-                stackTrace = buildStackTrace(log.getException());
-            }
+            String stackTrace = null;
 
             /* If exception in the log doesn't have stack trace try get it from the .throwable file. */
             if (stackTrace == null) {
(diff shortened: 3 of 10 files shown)


sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -1148,34 +1142,16 @@ private UUID saveErrorLogFiles(Throwable throwable, ManagedErrorLog errorLog) th
         String filename = errorLogId.toString();
         AppCenterLog.debug(Crashes.LOG_TAG, "Saving uncaught exception.");
         File errorLogFile = new File(errorStorageDirectory, filename + ErrorLogHelper.ERROR_LOG_FILE_EXTENSION);
+
+        /* Set stacktrace to error log. */
+        if (errorLog.getException() != null) {
```
russelarms, line 1147: The check is unnecessary since we don't expect the exception to be null here.
It comes either from @NonNull final Exception exception
```
@NonNull
    public static ManagedErrorLog createErrorLog(@NonNull Context context, @NonNull final java.lang.Thread thread, @NonNull final Exception exception, @NonNull final Map<java.lang.Thread, StackTraceElement[]> allStackTraces, final long initializeTimestamp, boolean fatal) {
```
or from
```
Exception modelException = new Exception();
modelException.setType("minidump");
modelException.setWrapperSdkName(WRAPPER_SDK_NAME_NDK);
modelException.setMinidumpFilePath(dest.getPath());
ManagedErrorLog errorLog = new ManagedErrorLog();
```

russelarms (changes requested):
Good work. Also, we need to remove `CallbackProcessor.shouldDeleteThrowable` method and its usages.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Version 4.0.1 (Under development)
 
+### App Center Crashes
+
+* **[Fix]** Fix removing throwable files after rewritten error logs due to small database size.
```
guperrot, line 7: Did you mean rewriting?

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelper.java
```diff
@@ -438,6 +439,23 @@ public static void removeStoredErrorLogFile(@NonNull UUID id) {
         }
     }
 
+    /**
+     * Remove throwable files.
+     */
+    public static void removeLostThrowableFiles() {
+        File[] throwableFiles = getErrorStorageDirectory().listFiles(new FilenameFilter() {
+            @Override
```
guperrot, line 447:
```suggestion

            @Override
```

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -1148,34 +1122,14 @@ private UUID saveErrorLogFiles(Throwable throwable, ManagedErrorLog errorLog) th
         String filename = errorLogId.toString();
         AppCenterLog.debug(Crashes.LOG_TAG, "Saving uncaught exception.");
         File errorLogFile = new File(errorStorageDirectory, filename + ErrorLogHelper.ERROR_LOG_FILE_EXTENSION);
+
+        /* Set stacktrace to error log. */
+        errorLog.getException().setStackTrace(getStackTraceString(throwable));
```
guperrot, line 1127: Will this be sent in the JSON? We should clean it up before sending.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -931,26 +913,18 @@ void setUncaughtExceptionHandler(UncaughtExceptionHandler handler) {
...
     @VisibleForTesting
-    @Nullable
     ErrorReport buildErrorReport(ManagedErrorLog log) {
         UUID id = log.getId();
         if (mErrorReportCache.containsKey(id)) {
             ErrorReport report = mErrorReportCache.get(id).report;
             report.setDevice(log.getDevice());
             return report;
         } else {
-            File file = ErrorLogHelper.getStoredThrowableFile(id);
```
guperrot, line 942: getStoredThrowableFile is not used anymore, but I didn't see the removal of this function.

guperrot (changes requested):

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
AnastasiaKubova: Do you mean that this report will be sent on the portal? If so, then I suggest that the report for sending is generated [here](https://github.com/microsoft/appcenter-sdk-android/blob/1a04d59f0f4382a67e6ee2f02bbf0a1bb9589733/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java#L916) and stack trace info adding to error report only once.

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/utils/ErrorLogHelper.java
```diff
@@ -438,6 +434,24 @@ public static void removeStoredErrorLogFile(@NonNull UUID id) {
...
+        File[] throwableFiles = getErrorStorageDirectory().listFiles(new FilenameFilter() {
+
+            @Override
+            public boolean accept(File dir, String filename) {
+                return filename.endsWith(THROWABLE_FILE_EXTENSION);
+            }
+        });
+        if (throwableFiles != null && throwableFiles.length > 0) {
+            for (File file : throwableFiles) {
+                removeStoredThrowableFile(UUID.fromString(file.getName().replaceFirst("[.][^.]+$", "")));
```
MatkovIvan, line 450: Do we really need regex here? Also, Using `[]` for this case is kind of wierd

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -931,26 +913,18 @@ void setUncaughtExceptionHandler(UncaughtExceptionHandler handler) {
...
-            if (file != null) {
-                String stackTrace = null;
-                if (file.length() > 0) {
-                    stackTrace = FileManager.read(file);
-                }
-                ErrorReport report = ErrorLogHelper.getErrorReportFromErrorLog(log, stackTrace);
-                mErrorReportCache.put(id, new ErrorLogReport(log, report));
-                return report;
-            }
+            String stackTrace = log.getException() == null ? null : log.getException().getStackTrace();
```
MatkovIvan, line 923: It looks like you didn't support data from old versions of SDK. You should keep reading as backup for now

MatkovIvan (changes requested):

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
guperrot: Yes this field would be sent in json but ignored by backend and can be big, it duplicates the structured stack trace.
So there should be a place in the code that unset the field when sending (at latest before enqueuing in channel).

sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -931,26 +913,18 @@ void setUncaughtExceptionHandler(UncaughtExceptionHandler handler) {
...
+            ErrorReport report = ErrorLogHelper.getErrorReportFromErrorLog(log, stackTrace);
```
guperrot, line 924: @AnastasiaKubova you can probably remove stacktrace from log.getException() (regarding https://github.com/microsoft/appcenter-sdk-android/pull/1480/files#r535530792).

AnastasiaKubova: I checked remove the stacktrace here but looks like it will not work in this case, because if after restarting the error reports are not sent, then the error data will be lost because error reports are overwritten.

guperrot: You should at least be able to remove the stack trace before the `Channel.enqueue` call.

guperrot: I see latest changes solve this issue. I was wondering what kind of tests cover the new code with the stack trace building.

AnastasiaKubova: Mainly this case is covered these two [tests]( https://github.com/microsoft/appcenter-sdk-android/pull/1480/files#diff-d1cf30587c3b0216759ecfd672e71cab0560bdc19d6a10894cca332ee0f2e193R1526).
But some other tests were also modified for checking it.

russelarms (approved):
The change looks good to me. 
I've tested the PR with the repro steps (migration dev > this branch, and on this branch solely) and with NDK crashes, works fine.

DmitriyKirakosyan (dismissed):

sdk/appcenter-crashes/src/androidTest/java/com/microsoft/appcenter/crashes/WrapperSdkExceptionManagerAndroidTest.java
```diff
@@ -109,7 +109,10 @@ public void saveWrapperException() {
             startFresh();
 
             /* Save crash. */
-            error.id = WrapperSdkExceptionManager.saveWrapperException(Thread.currentThread(), null, new Exception(), error.data);
+            Exception exception = new Exception();
+            exception.setType("type");
```
guperrot, line 113: What is the reason for those placeholder values for type and message?

AnastasiaKubova: reverted

DmitriyKirakosyan (approved):

guperrot (approved):

PR #1565, opened 2021-10-25
Update implementation of installing a new release

AnastasiaKubova:

* [x] Did you add unit tests?

## Description

We need to replace the current implementation of installing a new release to **PackageInstaller.Session**.

## Related PRs or issues

[AB#88756](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/88756)

### DIFF
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -1906,6 +1906,7 @@ public class Distribute extends AbstractAppCenterService {
     }
 
     @UiThread
+    @VisibleForTesting
     synchronized void notifyInstallProgress(boolean isInProgress) {
         mInstallInProgress = isInProgress;
         if (isInProgress) {
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -609,6 +609,10 @@ public class Distribute extends AbstractAppCenterService {
      * Register package installer receiver.
      */
     private synchronized void registerReceiver() {
+        if (!isInstanceEnabled()) {
+            AppCenterLog.warn(LOG_TAG, "Couldn't register receiver due to Distribute module is disabled.");
+            return;
+        }
         if (mForegroundActivity != null) {
             try {
                 mForegroundActivity.registerReceiver(mAppCenterPackageInstallerReceiver, mPackageInstallerReceiverFilter);
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -836,9 +836,6 @@ public class Distribute extends AbstractAppCenterService {
 
             /* Do nothing during installing a new release. */
             if (mInstallInProgress) {
-                if (mReleaseInstallerListener != null) {
-                    mReleaseInstallerListener.showInstallProgressDialog(mForegroundActivity);
-                }
                 AppCenterLog.info(LOG_TAG, "Installing in progress...");
                 return;
             }
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -13,7 +13,6 @@ import android.app.DownloadManager;
 import android.app.Notification;
 import android.app.NotificationChannel;
 import android.app.NotificationManager;
-import android.app.PendingIntent;
 import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.DialogInterface;
@@ -33,7 +32,6 @@ import android.text.TextUtils;
 import android.widget.Toast;
 
 import com.microsoft.appcenter.AbstractAppCenterService;
-import com.microsoft.appcenter.DependencyConfiguration;
 import com.microsoft.appcenter.Flags;
 import com.microsoft.appcenter.channel.Channel;
 import com.microsoft.appcenter.distribute.channel.DistributeInfoTracker;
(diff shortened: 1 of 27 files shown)


sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -571,6 +599,45 @@ protected synchronized void applyEnabledState(boolean enabled) {
...
+    private synchronized void unregisterReceiver() {
+        if (mForegroundActivity != null) {
+            try {
+                mForegroundActivity.unregisterReceiver(mAppCenterPackageInstallerReceiver);
+                AppCenterLog.debug(LOG_TAG, "The receiver for installing a new release was unregistered.");
+            } catch (IllegalArgumentException e) {
+                AppCenterLog.error(LOG_TAG, "The receiver wasn't unregistered.", e);
+            }
+        } else {
+            AppCenterLog.warn(LOG_TAG, "Couldn't register unregister due to activity is null.");
```
DmitriyKirakosyan, line 640:
```suggestion
            AppCenterLog.warn(LOG_TAG, "Couldn't unregister due to activity is null.");
```

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/InstallerUtils.java
```diff
@@ -121,4 +122,79 @@ public static boolean isUnknownSourcesEnabled(@NonNull Context context) {
...
+    public synchronized static boolean isSystemAlertWindowsEnabled(@NonNull Context context) {
+
+        /*
+        * From Android 10 (29 API level) or higher we have to
+        * use this permission for restarting activity after update.
+        * See more about restrictions on starting activities from the background:
+        * - https://developer.android.com/guide/components/activities/background-starts
+        * - https://developer.android.com/about/versions/10/behavior-changes-all#sysalert-go
+        */
+        return Build.VERSION.SDK_INT < Build.VERSION_CODES.Q ||
```
fabiendem, line 141: Sorry for the unsolicited comment, but isn't it relevant only for Android Go (Low ram) devices?
Instead of all `>= Build.VERSION_CODES.Q`

fabiendem: See https://stackoverflow.com/a/65176398

sdk/appcenter-distribute/src/main/res/values/appcenter_distribute.xml
```diff
@@ -17,9 +17,13 @@
     <!-- TODO: Deprecated. Remove this key and move the translated strings to appcenter_distribute_downloading_update. -->
     <string name="appcenter_distribute_downloading_mandatory_update">Downloading update</string>
     <!-- TODO: Add translations. -->
+    <string name="appcenter_distribute_alert_system_dialog_message" tools:ignore="MissingTranslation">For security, your device is set to block start application after update.</string>
     <string name="appcenter_distribute_downloading_version" tools:ignore="MissingTranslation">Downloading version %1$s (%2$d)</string>
     <string name="appcenter_distribute_downloading_error" tools:ignore="MissingTranslation">Failed to download app update</string>
     <string name="appcenter_distribute_download_progress_number_format">%1$d MB of %2$d MB</string>
+    <string name="appcenter_distribute_install_progress_number_format" tools:ignore="MissingTranslation">%1$d of %2$d</string>
```
DmitriyKirakosyan, line 24: What about localizaion?

DmitriyKirakosyan (changes requested):

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/InstallerUtils.java
AnastasiaKubova: Hi! Thanks for your suggestion!
Actually, I'm not sure. The public docs [describe](https://developer.android.com/about/versions/10/behavior-changes-all#sysalert-go):
> Apps running on Android 10 (Go edition) devices cannot receive the SYSTEM_ALERT_WINDOW permission.

Your link suggests using `isLowRamDevice` API to check this permission, but in the [public docs](https://developer.android.com/reference/android/app/ActivityManager#isLowRamDevice())`isLowRamDevice` API has a different purpose.
Unfortunately, I do not have the required device to make sure that this solution will work on these devices and that result of this API doesn't depend on some other permissions.

fabiendem: I think they use `isLowRamDevice` to detect the Android Go devices, which are low end devices.
https://www.android.com/versions/go-edition/
"A lightning speed OS, built for smartphones with 2 GB of RAM or less."

They use this because "SYSTEM_ALERT_WINDOW is disabled on on low ram devices starting from Q"

I was worried the PR would make us ask an unnecessary permission on non-Android Go device and got confused.
Looks like worst case you will have an unexpected behaviour on the Android Go devices.

Anyway, you have much more context than me 😅 
Thanks for looking at my comment. 🙌

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -757,6 +826,20 @@ private synchronized void resumeDistributeWorkflow() {
                 return;
             }
 
+            /* Continue to install a new release if before resume was shown dialog. */
```
DmitriyKirakosyan, line 829:
```suggestion
            /* Continue installing a new release if dialog was shown before resumeDistributeWorkflow. */
```

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/InstallerUtils.java
AnastasiaKubova: We use this permission only on Androids 10 and higher for the ability to restart the application after updating. This permission is optional for the users and can be ignored. In the worst case, it just won't open the app after the update.
Thanks for your attention and thoughts on this PR :).

DmitriyKirakosyan (approved):

PR #1600, opened 2022-02-22
Add null check for AppCenter.trackError

AnastasiaKubova:

* [ ] ~Did you add unit tests?~

## Description

[#2382](https://github.com/microsoft/appcenter/issues/2382/)

## Related PRs or issues

[AB#91298](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/91298)

### DIFF
--- a/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
+++ b/sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
@@ -266,7 +266,7 @@ public class Crashes extends AbstractAppCenterService {
      * @param throwable The throwable describing the handled error.
      */
     @SuppressWarnings({"SameParameterValue", "WeakerAccess"})
-    public static void trackError(Throwable throwable) {
+    public static void trackError(@NonNull Throwable throwable) {
         trackError(throwable, null, null);
     }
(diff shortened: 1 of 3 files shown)


sdk/appcenter-crashes/src/main/java/com/microsoft/appcenter/crashes/Crashes.java
```diff
@@ -266,7 +266,7 @@ static synchronized void unsetInstance() {
      * @param throwable The throwable describing the handled error.
      */
     @SuppressWarnings({"SameParameterValue", "WeakerAccess"})
-    public static void trackError(Throwable throwable) {
+    public static void trackError(@NonNull Throwable throwable) {
```
DmitriyKirakosyan, line 269: If I understand it correctly, it doesn't check for `null`, but generates a warning if I pass a `null` here.
I think we should check with `if` as well, shouldn't we?

DmitriyKirakosyan: Decided to avoid adding additional conditions. Annotation should be enough.

DmitriyKirakosyan (approved):

Issue #1603, opened 2022-03-01
Memory leak due to didn't unregister broadcastReceiver on Distribute.


gediseer:
### **Description**
Memroy leak has been found on All Activity, heap dump is below:
![image](https://user-images.githubusercontent.com/7627120/156122208-33fe1d9c-4fc1-4312-8e82-17031315c89a.png)
Looks lile it's caused by not unregistering the broadcast receiver after it's been registered on every "onResume" of each activity.
suggested fixed is unregister broadcast receiver on "onPause"
![image](https://user-images.githubusercontent.com/7627120/156122383-c9bb0b9a-2af1-4aba-9142-7b889aecc146.png)

Please describe the issue you are facing using the SDK.
The memory usage keep increasing while the "onResume" callback of each Activity being call many times, and it's won't decrease until 
OOM.

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. start an Activity whose "onResume" callback will be observed by Appcenter. 
2. click back button on navigation bar, and then go back to the previous activity, to make sure it's "onResume" callback being invoded again.
3. check the memory usage on Android Studio Profiler.

### **Details**

1. Which SDK version are you using?
    - 4.4.2
2. Which OS version did you experience the issue on?
    - Android 12
4. What device version did you see this error on?  Were you using an emulator or a physical device?
    - Google Pixel 4xl physical device
5. What third party libraries are you using?
    - com.microsoft.appcenter:appcenter-distribute:4.4.2
6. Please enable verbose logging for your app using `AppCenter.setLogLevel(Log.VERBOSE)` before your call to `AppCenter.start(...)` and include the logs here:

////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
2022-03-01 15:07:16.316 21265-21265/com.microsoft.example I/AppCenter: App Center SDK configured successfully.
2022-03-01 15:07:16.316 21265-21265/com.microsoft.example D/AppCenter: Cannot read instrumentation variables in a non-test environment.
2022-03-01 15:07:16.316 21265-21265/com.microsoft.example D/AppCenter: Cannot read instrumentation variables in a non-test environment.
2022-03-01 15:07:16.327 21265-21318/com.microsoft.example D/AppCenter: Loaded stored sessions: {}
2022-03-01 15:07:16.331 21265-21318/com.microsoft.example W/AppCenter: Unable to get installID from Shared Preferences
[...]

PR #1605, opened 2022-03-04
Fix superfluous register/unregister receiver for installing new release

AnastasiaKubova:

* [x] Did you add unit tests?

## Related PRs or issues

[AB#91443](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/91443)
#1603

### DIFF
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -540,6 +540,12 @@ public class Distribute extends AbstractAppCenterService {
     }
 
     @Override
+    public void onActivityStarted(Activity activity) {
+        super.onActivityStarted(activity);
+        registerReceiver(activity);
+    }
+
+    @Override
     public synchronized void onActivityResumed(Activity activity) {
         mForegroundActivity = activity;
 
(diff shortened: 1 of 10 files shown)


DmitriyKirakosyan (approved):

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -539,6 +539,12 @@ private boolean tryResetWorkflow() {
         return false;
     }
 
+    @Override
+    public void onActivityStarted(Activity activity) {
+        super.onActivityStarted(activity);
+        registerReceiver(activity);
```
gediseer, line 545: Now that we register receiver in "onStart" callback, Do we need unregister it in "onStop" so that it comes in pairs?

AnastasiaKubova: No, it doesn’t depend on lifecycle of application. Receiver is registered only when distribute is enabled and is unregistered when distribute is disabled. This call need in this method to avoid issue when distribute was enabled, but activity has not started yet. If receiver was already registered so this call will be ignored.

DmitriyKirakosyan (approved):

Issue #1599, opened 2022-02-14
Crash when direct boot


vvb2060:
### **Description**

```
[ 2022-02-13T10:30:39.144    11471:  3145:  3347 E/AndroidRuntime  ] FATAL EXCEPTION: AppCenter.Looper
Process: [edited], PID: 3145
java.lang.IllegalStateException: SharedPreferences in credential encrypted storage are not available until after user is unlocked
 at android.app.ContextImpl.getSharedPreferences(ContextImpl.java:494)
 at android.app.ContextImpl.getSharedPreferences(ContextImpl.java:479)
 at android.content.ContextWrapper.getSharedPreferences(ContextWrapper.java:188)
 at com.microsoft.appcenter.utils.storage.SharedPreferencesManager.initialize(SharedPreferencesManager.java:45)
 at com.microsoft.appcenter.AppCenter.finishConfiguration(AppCenter.java:832)
 at com.microsoft.appcenter.AppCenter.access$500(AppCenter.java:61)
 at com.microsoft.appcenter.AppCenter$5.run(AppCenter.java:737)
 at android.os.Handler.handleCallback(Handler.java:938)
 at android.os.Handler.dispatchMessage(Handler.java:99)
 at android.os.Looper.loop(Looper.java:223)
 at android.os.HandlerThread.run(HandlerThread.java:67)
```

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. Add ` AppCenter.start()` when `Application.onCreate()`
2. add receiver
```
        <receiver
            android:name=".BootReceiver"
            android:exported="true"
            android:directBootAware="true">
            <intent-filter>
                <action android:name="android.intent.action.LOCKED_BOOT_COMPLETED" />
            </intent-filter>
        </receiver>
```
3. reboot and `logcat -b crash`

### **Details**

1. Which SDK version are you using?
   4.4.2
2. Which OS version did you experience the issue on?
   Android 7.0+
3. What device version did you see this error on?  Were you using an emulator or a physical device?
   All Android 7.0+ FBE devices

https://developer.android.com/training/articles/direct-boot

AnatolyPristensky:
Hello @vvb2060 
And thank you for getting in touch with us!
I'm working on your issue. Will update this conversation once get results.

vvb2060:
No need to reproduce, just follow this guide https://developer.android.com/training/articles/direct-boot to fix it.

vvb2060:
[...]

PR #1623, opened 2022-05-11
Fix crash when direct boot

AnatolyPristensky:
* [X] Has `CHANGELOG.md` been updated?
* [x] Did you add unit tests?

## Description

Fix crash when direct boot is used. 

## Related PRs or issues

#1599
[AB#91315](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/91315)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Version 4.4.4 (Under active development)
 
+### App Center
+
+* **[Fix]** Crash when direct boot is used.
+
 ### App Center Distribute
 
 * **[Fix]** Fix possible crash on resume event before initialization.
--- a/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/SharedPreferencesManager.java
+++ b/sdk/appcenter/src/main/java/com/microsoft/appcenter/utils/storage/SharedPreferencesManager.java
@@ -5,9 +5,10 @@
 
 package com.microsoft.appcenter.utils.storage;
 
-import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.os.Build;
+
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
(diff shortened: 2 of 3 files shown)


sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/utils/storage/SharedPreferencesManagerAndroidTest.java
```diff
@@ -102,7 +106,16 @@ private static SharedPreferencesTestData[] generateSharedPreferenceData() throws
 
         return testData;
     }
-
+    @Test
```
DmitriyKirakosyan, line 109: Empty space before and after this method

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/utils/storage/SharedPreferencesManagerAndroidTest.java
```diff
@@ -102,7 +106,16 @@ private static SharedPreferencesTestData[] generateSharedPreferenceData() throws
...
+    public void sharedPreferencesProtectedStorage() {
+        Boolean isProtectedStorage = context.isDeviceProtectedStorage();
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
```
DmitriyKirakosyan, line 112: Condition in test function? Wouldn't it better to separate the testing on two test methods?

sdk/appcenter/src/androidTest/java/com/microsoft/appcenter/utils/storage/SharedPreferencesManagerAndroidTest.java
```diff
@@ -102,7 +106,16 @@ private static SharedPreferencesTestData[] generateSharedPreferenceData() throws
...
+            assertEquals(isProtectedStorage, false);
+        }
+        else {
```
DmitriyKirakosyan, line 115: `else` on the same line with `}`

MatkovIvan (commented):
Need to be tested manually before merge

DmitriyKirakosyan:
@AnatolyPristensky I've tested your changes and found a problem. In this case, the updated SDK will create new preferences and thus the previously saved data will be lost.
Instead of writing lots of code implementing the migration from old preferences to the new ones, It would make sense to create the protected preferences only in the case of the exception. Let's create normal shared preferences as we did before and handle the error, and only create protected shared preferences in case we catch the error. This way we can avoid supporting the migration between preferences.

PR #1633, opened 2022-06-19
No need to keep class names

vvb2060:
* [ ] Did you add unit tests?

## Description

`keepclasseswithmembers` keep class names, which is not need because we include class in arguments.

### DIFF
--- a/sdk/appcenter/proguard-rules.pro
+++ b/sdk/appcenter/proguard-rules.pro
@@ -3,7 +3,7 @@
 -dontskipnonpubliclibraryclasses
 -verbose
 
--keepclasseswithmembers class * implements com.microsoft.appcenter.AppCenterService {
+-keepclassmembers class * implements com.microsoft.appcenter.AppCenterService {
     public static ** getInstance();
 }
 


MatkovIvan (changes requested):
Class names are required for wrapper SDKs (like Xamarin, React-Native, etc)

vvb2060:
Do these libraries (not even Java libraries) require reflection to use appcenter-crashes/appcenter-analytics?  Why not add keep rules to their library?  Most users will not use these libraries.

MatkovIvan:
Unity one [use it for sure](https://github.com/microsoft/appcenter-sdk-unity/blob/a501cf9ce84c93cbad6a3e63c11852d83cfa8762/Assets/AppCenter/Plugins/AppCenterSDK/Analytics/Android/AnalyticsInternal.cs#L14).

Could you please elaborate why exactly is it important for you?
For now this PR cannot be merged as-is because it requires changes in/testing of other parts of the product. But the benefit is not clear.

vvb2060:
I would like to completely obfuscate appcenter, reducing the size of the app is one reason. Another reason is that some users are very wary of telemetry, and they will suspect our app will compromise privacy because of the dex scanner warning without reading our privacy policy. I hope to mitigate this problem.

MatkovIvan:
> reducing the size of the app is one reason

Names of 2 classes won't help with this.

Regarding hiding "annoying" names - ok, it seems valid improvement. But as I mentioned before it requires some additional work from our team, so it cannot be merged now.
Anyway, thanks for your contribution and highlighting this topic!

DmitriyKirakosyan:
I'm closing this PR as stale.

PR #1662, opened 2022-10-21
Decouple versioning of sasquatch app from SDK

shadyabarada:

* [ ] Did you add unit tests?

## Description

Currently the Sasquatch App version is coupled with the AppCenter Android SDK version due to the unity of the versioning in the versions.gradle file.

This coupling will raise an issue when trying to test the distribute feature with the latest SDK module, here's an example scenario:
I want to verify that the distribute feature works as expected for the new app center sdk version (5.0.0)
If I downgrade the version from the versions.gradle file (let's say to 4.0.0) for the sake of downgrading the Sasquatch application to test the distribute behaviour, the sdk's version is thus downgraded as well (which is not what we want because we want to actually test the behaviour of the sdk against version 5.0.0 😊) 

Therefore, it makes sense for the Sasquatch app to have its own versioning criteria that does not get affected by the SDK. One of the solutions can be overriding the global (versions.gradle) versioning in the Sasquatch's build.gradle. 

## Related PRs or issues

## Misc

### DIFF
--- a/apps/sasquatch/build.gradle
+++ b/apps/sasquatch/build.gradle
@@ -14,6 +14,8 @@ android {
     flavorDimensions "dependency", "distribute"
 
     defaultConfig {
+        versionCode 300
+        versionName "5.0.0"
         externalNativeBuild {
             ndkBuild {
                 arguments "NDK_APPLICATION_MK=Application.mk", "V=1"


thewulf7:
@shadyabarada planning to merge it?

aleksandr-dorofeev (approved):

PR #1663, opened 2022-10-25
Fix NPE in Distribute module on resume workflow

shadyabarada:

* [x] Did you add unit tests?

## Description

My initial thoughts around this issue:
The issue is happening under the scenario of which the distribute SDK needs to show the "Update Setup Failed" dialog. 
When the "showUpdateSetupFailedDialog" method is called, the alert dialog uses the global variable activity instance mForegroundActivity.
The mForegroundActivity in the above scenario is null. Thus the crash.
How could this variable be null? Let's observe the workflow:
In our Distribute.java class, mForegroundActivity is a global private variable used in the scope of this class.
The value of the mForegroundActivity is set under the following two cases:
onActivityResumed => The value of mForegroundActivity is set to the value of the activity being resumed. 
onActivityPaused => The value of mForegroundActivity is explicitly set to null.
The call for "showUpdateSetupFailedDialog" occurs inside the "resumeDistributeWorkflow" function which is called by the onActivityResumed method. 
The "resumeDistributeWorkflow" has a condition on its very first line which checks if our mForegroundActivity is null, and apparently, this check passes before reaching the line of code that should show the "update failed" alert dialog where our code is crashing.
Therefore, the value of mForegroundActivity has been tampered with along the way before reaching the "show update failed" dialog. Weird right? We are within the same block of code, under the same method, and between few lines of code that do not explicitly change the value of our mForegroundActivity, the value has actually changed. 
This issue is most probably a "thread shared memory issue". I am assuming that one of the running threads have executed the onActivityPause function which updates the value of the mForegroundActivity to null, while on onActivityResume code is being executed. Thus failing to show the alert dialog and producing the above null pointer exception.
I will be investigating further and trying to create a small test with multithreads trying to access and update global variables, and look into possible solutions. 

Note that the user which this issue is occurring with, might be stuck in the loop of crashes every time he/she opens their app because of having the values with the running threads continually stored in memory - I will reply on the GitHub conversation thread for this issue asking if they can try to clear app cache and data for their app from the device's settings and try again to see if the issue continuous. 

## Related PRs or issues

https://dev.azure.com/msmobilecenter/Mobile-Center/_workitems/edit/95531

## Misc

### DIFF
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -1580,35 +1580,38 @@ public class Distribute extends AbstractAppCenterService {
         if (!shouldRefreshDialog(mUpdateSetupFailedDialog)) {
             return;
         }
-        if(mForegroundActivity!=null) {
-            AppCenterLog.debug(LOG_TAG, "Show update setup failed dialog.");
-            AlertDialog.Builder dialogBuilder = new AlertDialog.Builder(mForegroundActivity);
-            dialogBuilder.setCancelable(false);
-            dialogBuilder.setTitle(R.string.appcenter_distribute_update_failed_dialog_title);
-            dialogBuilder.setMessage(R.string.appcenter_distribute_update_failed_dialog_message);
-            dialogBuilder.setPositiveButton(R.string.appcenter_distribute_update_failed_dialog_ignore, new DialogInterface.OnClickListener() {
 
-                @Override
-                public void onClick(DialogInterface dialog, int which) {
-                    storeUpdateSetupFailedPackageHash(dialog);
-                }
-            });
-            dialogBuilder.setNegativeButton(R.string.appcenter_distribute_update_failed_dialog_reinstall, new DialogInterface.OnClickListener() {
-
-                @Override
-                public void onClick(DialogInterface dialog, int which) {
-                    handleUpdateFailedDialogReinstallAction(dialog);
-                }
-            });
-            mUpdateSetupFailedDialog = dialogBuilder.create();
-            showAndRememberDialogActivity(mUpdateSetupFailedDialog);
-
-            /* Don't show this dialog again. */
-            SharedPreferencesManager.remove(PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY);
-        }else
-        {
+        /* Check if foreground activity is null, if null we return, if not we proceed to show the setup failed dialog */
+        if (mForegroundActivity == null) {
             AppCenterLog.debug(LOG_TAG, "Failed to show the update setup failed dialog. The foreground activity is null");
+            return;
         }
+
+        AppCenterLog.debug(LOG_TAG, "Show update setup failed dialog.");
+        AlertDialog.Builder dialogBuilder = new AlertDialog.Builder(mForegroundActivity);
+        dialogBuilder.setCancelable(false);
+        dialogBuilder.setTitle(R.string.appcenter_distribute_update_failed_dialog_title);
+        dialogBuilder.setMessage(R.string.appcenter_distribute_update_failed_dialog_message);
+        dialogBuilder.setPositiveButton(R.string.appcenter_distribute_update_failed_dialog_ignore, new DialogInterface.OnClickListener() {
+
+            @Override
+            public void onClick(DialogInterface dialog, int which) {
+                    storeUpdateSetupFailedPackageHash(dialog);
+            }
+        });
+        dialogBuilder.setNegativeButton(R.string.appcenter_distribute_update_failed_dialog_reinstall, new DialogInterface.OnClickListener() {
+
+            @Override
+            public void onClick(DialogInterface dialog, int which) {
+                handleUpdateFailedDialogReinstallAction(dialog);
+            }
+        });
+        mUpdateSetupFailedDialog = dialogBuilder.create();
+        showAndRememberDialogActivity(mUpdateSetupFailedDialog);
+
+        /* Don't show this dialog again. */
+        SharedPreferencesManager.remove(PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY);
+
     }
 
     /**
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -1581,7 +1581,6 @@ public class Distribute extends AbstractAppCenterService {
             return;
         }
 
-        /* Check if foreground activity is null, if null we return, if not we proceed to show the setup failed dialog */
         if (mForegroundActivity == null) {
             AppCenterLog.debug(LOG_TAG, "Failed to show the update setup failed dialog. The foreground activity is null");
             return;
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -1574,7 +1574,8 @@ public class Distribute extends AbstractAppCenterService {
      * Show update setup failed dialog.
      */
     @UiThread
-    private synchronized void showUpdateSetupFailedDialog() {
+    @VisibleForTesting
+    protected synchronized void showUpdateSetupFailedDialog() {
 
         /* Check if we need to replace the dialog. */
         if (!shouldRefreshDialog(mUpdateSetupFailedDialog)) {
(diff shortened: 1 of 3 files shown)


sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1580,30 +1580,35 @@ private synchronized void showUpdateSetupFailedDialog() {
         if (!shouldRefreshDialog(mUpdateSetupFailedDialog)) {
             return;
         }
-        AppCenterLog.debug(LOG_TAG, "Show update setup failed dialog.");
-        AlertDialog.Builder dialogBuilder = new AlertDialog.Builder(mForegroundActivity);
-        dialogBuilder.setCancelable(false);
-        dialogBuilder.setTitle(R.string.appcenter_distribute_update_failed_dialog_title);
-        dialogBuilder.setMessage(R.string.appcenter_distribute_update_failed_dialog_message);
-        dialogBuilder.setPositiveButton(R.string.appcenter_distribute_update_failed_dialog_ignore, new DialogInterface.OnClickListener() {
+        if(mForegroundActivity!=null) {
```
DmitriyKirakosyan, line 1583: Let's make it simpler:

```
if (mForegroundActivity == null) {
    AppCenterLog.debug(LOG_TAG, "Failed to show the update setup failed dialog. The foreground activity is null");
    return;
}
```

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1580,6 +1580,13 @@ private synchronized void showUpdateSetupFailedDialog() {
         if (!shouldRefreshDialog(mUpdateSetupFailedDialog)) {
             return;
         }
+
+        /* Check if foreground activity is null, if null we return, if not we proceed to show the setup failed dialog */
```
DmitriyKirakosyan, line 1584: I would avoid such comments, if the written code is clean and easily understandable :)

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1589,7 +1596,7 @@ private synchronized void showUpdateSetupFailedDialog() {
 
             @Override
             public void onClick(DialogInterface dialog, int which) {
-                storeUpdateSetupFailedPackageHash(dialog);
+                    storeUpdateSetupFailedPackageHash(dialog);
```
DmitriyKirakosyan, line 1599: let's revert this

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1604,6 +1611,7 @@ public void onClick(DialogInterface dialog, int which) {
 
         /* Don't show this dialog again. */
         SharedPreferencesManager.remove(PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY);
+
```
DmitriyKirakosyan, line 1614: Revert pls. Normally we only add empty lines before a comment.

shadyabarada: Done

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
shadyabarada: done

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
shadyabarada: done

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
shadyabarada: done

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1580,6 +1580,12 @@ private synchronized void showUpdateSetupFailedDialog() {
         if (!shouldRefreshDialog(mUpdateSetupFailedDialog)) {
             return;
         }
+
```
DmitriyKirakosyan, line 1583:
```suggestion
```

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1580,6 +1580,12 @@ private synchronized void showUpdateSetupFailedDialog() {
...
+        if (mForegroundActivity == null) {
+            AppCenterLog.debug(LOG_TAG, "Failed to show the update setup failed dialog. The foreground activity is null");
+            return;
+        }
+
```
DmitriyKirakosyan, line 1588:
```suggestion
```

DmitriyKirakosyan (changes requested):

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeTest.java
```diff
@@ -448,6 +448,21 @@ public void doNotShowUpdateSetupFailedDialogInBackground() {
...
     }
 
+    @Test
+    public void doNotShowUpdateSetupFailedDialogInNullForegroundActivity() {
+        when(SharedPreferencesManager.getString(PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY)).thenReturn("failed_message");
+
+        /* Trigger call. */
+        Distribute.getInstance().showUpdateSetupFailedDialog();
+        mActivity = null;
+        start();
```
DmitriyKirakosyan, line 458: Do we need to call `start`?

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeTest.java
```diff
@@ -448,6 +448,21 @@ public void doNotShowUpdateSetupFailedDialogInBackground() {
         SharedPreferencesManager.remove(PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY);
     }
 
+    @Test
+    public void doNotShowUpdateSetupFailedDialogInNullForegroundActivity() {
+        when(SharedPreferencesManager.getString(PREFERENCE_KEY_UPDATE_SETUP_FAILED_MESSAGE_KEY)).thenReturn("failed_message");
```
DmitriyKirakosyan, line 453: I think we don't need to set this preference, as we only call `showUpdateSetupFailedDialog`

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeTest.java
```diff
@@ -448,6 +448,21 @@ public void doNotShowUpdateSetupFailedDialogInBackground() {
...
+
+        /* Trigger call. */
+        Distribute.getInstance().showUpdateSetupFailedDialog();
+        mActivity = null;
```
DmitriyKirakosyan, line 457: I think we don't need this. But if we needed, I think it should be like this:
```
Distribute.getInstance().onActivityResumed(null);
```

and before we call `showUpdateSetupFailedDialog`.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeTest.java
```diff
@@ -448,6 +448,21 @@ public void doNotShowUpdateSetupFailedDialogInBackground() {
...
+        start();
+
+        /* Verify dialog. */
+        verify(mDialogBuilder, never()).create();
+        verifyStatic(SharedPreferencesManager.class, never());
```
DmitriyKirakosyan, line 462: We don't need to check `SharedPreferencesManager`. All we need to verify is that the call doesn't throw NPE and it doesn't reach the dialog creation. So I think checking that dialog builder was not created is enough here.

DmitriyKirakosyan (changes requested):

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeTest.java
shadyabarada: We will need to set the activity to null, I will do it through calling the on resume with null value

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeTest.java
shadyabarada: Not needed, removed

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/DistributeTest.java
shadyabarada: Update: on resume call not needed, activity is already null if not set by on resume

Priskorn (approved):

DmitriyKirakosyan (approved):

PR #1690, opened 2023-06-05
Ignore file size check if original update file has AAB extension format

MikhailSuendukov:
* [x] Did you add unit tests?

## Description

Added the `is_install_page=true` parameter to the URL for the release details request in order to receive a link to download the APK file even if an ABB file has been uploaded to the portal. Also, due to the fact that when uploading an AAB file to the portal, the release details indicate the size of the AAB file, even if we download the APK file, we ignore the check of the size of the uploaded file in this case.

### DIFF
--- a/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
+++ b/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
@@ -19,7 +19,6 @@ import static com.microsoft.appcenter.distribute.DistributeConstants.HEADER_API_
 import static com.microsoft.appcenter.distribute.DistributeConstants.LOG_TAG;
 import static com.microsoft.appcenter.distribute.DistributeConstants.PARAMETER_DISTRIBUTION_GROUP_ID;
 import static com.microsoft.appcenter.distribute.DistributeConstants.PARAMETER_INSTALL_ID;
-import static com.microsoft.appcenter.distribute.DistributeConstants.PARAMETER_IS_INSTALL_PAGE;
 import static com.microsoft.appcenter.distribute.DistributeConstants.PARAMETER_RELEASE_ID;
 import static com.microsoft.appcenter.distribute.DistributeConstants.PARAMETER_UPDATE_SETUP_FAILED;
 import static com.microsoft.appcenter.distribute.DistributeConstants.POSTPONE_TIME_THRESHOLD;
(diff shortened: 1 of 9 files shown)


sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/ReleaseDetails.java
```diff
@@ -195,6 +203,15 @@ int getMinApiLevel() {
         return minApiLevel;
     }
 
+    /**
+     * Get The original update file extension value.
+     *
+     * @return the fileExtension value
```
DmitriyKirakosyan, line 209:
```suggestion
     * @return the FileExtension value.
```

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java
```diff
@@ -1064,6 +1065,7 @@ synchronized void getLatestReleaseDetails(final String distributionGroupId, Stri
         } else {
             url += String.format(GET_LATEST_PRIVATE_RELEASE_PATH_FORMAT, mAppSecret, releaseHash, getReportingParametersForUpdatedRelease(false, distributionGroupId));
         }
+        url += "&" + PARAMETER_IS_INSTALL_PAGE + "=" + "true";
```
DmitriyKirakosyan, line 1068:
```suggestion
        url += "&" + PARAMETER_IS_INSTALL_PAGE + "=true";
```

sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/download/manager/DownloadManagerReleaseDownloader.java
```diff
@@ -249,6 +250,9 @@ synchronized void onDownloadError(RuntimeException e) {
     }
 
     private boolean isDownloadedFileValid() {
+        if (mReleaseDetails.getFileExtension() != FileExtension.apk) {
```
DmitriyKirakosyan, line 253: Let's leave a comment here, because looking at code it is not clear what happens. "If the file is not apk, then it is valid" doesn't make sense, so we need to clarify why this condition exists here.

sdk/appcenter-distribute/src/test/java/com/microsoft/appcenter/distribute/download/manager/DownloadManagerReleaseDownloaderTest.java
```diff
@@ -449,12 +454,30 @@ public void errorOnOpenDownloadedFile() throws IOException {
         verify(mListener).onError(anyString());
     }
 
+    @Test
+    public void errorOnCloseFileDescriptor() throws IOException {
```
DmitriyKirakosyan, line 458: what is this test for?

MikhailSuendukov: In order to cover all error cases just in case

DmitriyKirakosyan (approved):

PR #1734, opened 2024-08-12
Enable ESRP task

DordeDimitrijev:
Enable ESRP task in order to publish sdk release


MikhailSuendukov (approved):

IlyaBausovAkvelon (approved):
@DordeDimitrijev could you please check if there is a EsrpRelease@5 task instead of SFP.release-tasks.custom-build-release-task.EsrpRelease@4 so it could be possibly used in the future changes?

